PR #6, opened 2025-10-24
fix: do not use localhost on dev server

RomainJeff:
### Why this change ?
I've been facing some issues to get the frontend to load properly in widgets inside ChatGPT because trying to reach `localhost` was not possible on ChatGPT UI somehow.

My fix have been to directly use the host that is making the call, since anyway we are exposing our local through tools like `ngrok` and `cloudflared`.

I don't know if i'm missing something, otherwise, would make sense to either always use the host (like i did) or to at least make it more configurable.

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -57,12 +57,10 @@ export class McpServer extends McpServerBase {
       },
       async (_uri, extra) => {
         const serverUrl =
-          process.env.NODE_ENV === "production"
-            ? `https://${
-                extra?.requestInfo?.headers?.["x-forwarded-host"] ??
-                extra?.requestInfo?.headers?.host
-              }`
-            : `http://localhost:3000`;
+          `https://${
+            extra?.requestInfo?.headers?.["x-forwarded-host"] ??
+            extra?.requestInfo?.headers?.host
+          }`;
 
         const templateData = {
           serverUrl,


fredericbarthelet:
Hey @RomainJeff, thanks for opening this issue :)

> trying to reach localhost was not possible on ChatGPT UI somehow
Could you share a bit more on that issue (any error in the console)?

The main reason assets are served from localhost in dev is to be able to use the hot module replacement feature from Vite. This one leverage a secure websocket connection that does not play well with local reverse proxy utils like ngrok and cloudlared tunnel. Did you manage to make this HMR connection work through your config? I'd be interested in understanding how you made that work.

Only the MCP server tool and resource need to be accessible publicly on internet in order for OpenAI backends to reach it. Skybridge (OpenAI runtime in charge of starting the iFrame for your widget) runs on your browser. Once it loads the HTML from the resource of your server, any additional JS and CSS assets that needs to be downloaded will be downloaded by your browser and can therefor be directly served by your dev server on localhost.

RomainJeff:
>Hey @RomainJeff, so does adding the header Access-Control-Allow-Private-Network: true in the server response had any effect on your side ?

Sorry for delayed answer : It fixed it on Brave but not on Chrome Canary (upcoming version of Chrome).

>In the case of requests made from an iframe, to make it work the document that declares the iframe must add a parameter to it so that it's authorized to make local requests allow="local-network-access *"

So that would be OpenAI who would need to do it right ? Tbh I think (hope) they have a LOT of things to update on their platform to make the developer experience much better for the MCP & Apps 😅

RomainJeff:
hey @harijoe @fredericbarthelet 

Good news, I just tested right now on Chrome Canary and it now asks my permission to give access to my local network and when I accept it works !

Maybe some guidelines to add to the documentation to clarify this extra step in the upcoming Chrome (and derived browsers) 🎉 

Here is a little screenshot : 
image

PR #8, opened 2025-11-17
Add useCallTool hook

fredericbarthelet:


harijoe (commented):
Are we sure we want to provide all the boolean helpers (like isPending, isError etc.)?
And if we do, maybe refacto by getting these boolean helpers from the status?
The current implementation feels quite verbose.

PR #28, opened 2025-11-28
add useSendFollowUpMessage hook

### LINKED ISSUE #27: Enable a widget to send a prompt to the LLM
## Description
Use the `window.openai.sendFollowUpMessage` method to send a prompt back to the client.

## Current Behavior
Currently it's possible to do this by calling `useOpenAiGlobal("sendFollowUpMessage")` but this seems to much of a "bare bone" solution for a framework.

## New Behavior
Creating a hook called `useSendFollowUpMessage` that returns the `sendFollowUpMessage` function. This function can be called inside button handlers to send a new prompt to the LLM.

```ts
export function useSendFollowUpMessage() {
  const sendFollowUpMessage = useCallback(
    async (prompt: string): Promise<void> => {
      if (!window.openai?.sendFollowUpMessage) {
        throw new Error("window.openai.sendFollowUpMessage is not available");
      }
      return window.openai.sendFollowUpMessage({ prompt });
    },
    []
  );
  return { sendFollowUpMessage };
}
```

Usage
```ts
const { sendFollowUpMessage } = useSendFollowUpMessage()

const handleRecipeRequest = () => {
    sendFollowUpMessage(
      'Based on the current fridge contents, propose me some recipes I can make with these ingredients.',
    );
  };

<button className='recipe-button' type='button' onClick={handleRecipeRequest}>
      Propose me recipes
</button>
```

## Resources
[window.openai docs](https://developers.openai.com/apps-sdk/build/chatgpt-ui)

Rodriguespn:
New `useSendFollowUpMessage` hook to follow up send prompt to the LLM

Closes #27

### DIFF
--- a/src/web/use-send-follow-up-message.ts
+++ b/src/web/use-send-follow-up-message.ts
@@ -11,5 +11,5 @@ export function useSendFollowUpMessage() {
     []
   );
 
-  return { sendFollowUpMessage };
+  return sendFollowUpMessage;
 }


fredericbarthelet (approved):
Thanks for your contribution @Rodriguespn ! LGTM
Only suggestion I may have is on the API of the hook.
There is only a single item returned from the hook, so I'd return it directly to allow easy renaming in implementation.

```ts
const send = useSendFollowUpMessage()
```

WDYT?

PR #34, opened 2025-11-28
Improve useToolInfo

fredericbarthelet:
This PR aims at providing improvements to the `useToolInfo` hook.

This hook always returns the input to the tool that triggered the widget rendering, but output and response metadata are initially null.

Indeed, the widget rendering happens before the tool is even fired - this can be witnessed on tools that require human approval: the widget is rendering behind the approval modal.
In order to provide an easier to accommodate for this, `useToolInfo` hook now returns the following additional props:
- `status`, that can be `pending` (the initial state) and becomes `success` once output and/or metadata have been hydrated on window.openai
- `isPending` a convenient boolean indicating the status value is pending
- `isSuccess` a convenient boolean indicating the status value is success

In addition to those properties, `input`, `output` and `responseMetadata` types have been updated. The types of each of these property can now be narrow down using a generic. The generic accepts only a single type argument and is flexible to allow providing only a subset of those types.

_Full typing_
```ts
const { input, output, responseMetadata } = useToolInfo<{
  input: { name: string };
  output: { value: number };
  responseMetadata: { word: string; illustrationUrl: string };
}>();

typeof input; // { name: string }
typeof output; // { value: number }
typeof responseMetadata; // { word: string; illustrationUrl: string }
```

_Partial typing_
```ts
const { input, output, responseMetadata } = useToolInfo<{
  input: { name: string };
}>();

typeof input; // { name: string }
typeof output; // UnknownObject
typeof responseMetadata; // UnknownObject
```

The types also adapts depending on the status value

```ts
const { status, output } = useToolInfo<{
  output: { value: number };
}>();

if (status === "pending") {
  typeof output; // undefined
} else {
  typeof output; // { value: number }
}
```

### DIFF
--- a/src/web/hooks/use-tool-info.ts
+++ b/src/web/hooks/use-tool-info.ts
@@ -44,6 +44,7 @@ export function useToolInfo<
   }>
 >() {
   const [status, setStatus] = useState<"pending" | "success">("pending");
+  const input = useOpenAiGlobal("toolInput")!;
   const output = useOpenAiGlobal("toolOutput") ?? undefined;
   const responseMetadata = useOpenAiGlobal("toolResponseMetadata") ?? undefined;
 
@@ -56,7 +57,7 @@ export function useToolInfo<
   }, [output, responseMetadata]);
 
   return {
-    input: useOpenAiGlobal("toolInput")!,
+    input,
     status,
     isPending: status === "pending",
     isSuccess: status === "success",


src/web/hooks/use-tool-info.ts
```diff
@@ -1,9 +1,73 @@
...
+    );
+  }, [output, responseMetadata]);
 
-export function useToolInfo() {
   return {
-    output: useOpenAiGlobal("toolOutput"),
-    input: useOpenAiGlobal("toolInput"),
-    responseMetadata: useOpenAiGlobal("toolResponseMetadata"),
-  };
+    input: useOpenAiGlobal("toolInput")!,
```
harijoe: It may work well, but declaring a hook in the returned object feel very suspicious to me.

src/web/hooks/use-tool-info.ts
```diff
@@ -1,9 +1,73 @@
...
+>() {
+  const [status, setStatus] = useState<"pending" | "success">("pending");
+  const output = useOpenAiGlobal("toolOutput") ?? undefined;
+  const responseMetadata = useOpenAiGlobal("toolResponseMetadata") ?? undefined;
+
+  useEffect(() => {
+    setStatus(
+      output === undefined && responseMetadata === undefined
+        ? "pending"
+        : "success"
```
harijoe, line 55: I'm actually not a big fan of this.

The type of `useOpenAiGlobal("toolOutput")` already specifies that the result can but `ToolOutput | null`. I don't find necessary to add a layer making the null check and providing the response in a boolean. It's quite common in react to handle values that are initialized as null or undefined and then takes their values.

Maybe making a suspense version of useToolInfo instead would bring more value?

src/web/types.ts
```diff
@@ -50,7 +50,7 @@ export type OpenAiGlobals<
 
   // state
   toolInput: ToolInput;
-  toolOutput: ToolOutput | null;
+  toolOutput: ToolOutput | { text: string } | null;
```
harijoe, line 53: During my tests, toolOutput had the following signature :
```
{
    "result": {
        "meta": null,
        "content": [],
        "structuredContent": null,
        "isError": false
    },
    "error": null,
    "error_data": null,
    "error_code": null,
    "error_http_status_code": null,
    "error_http_headers": null,
    "internal": {
        "external_call_time_ms": 1657
    },
    "connector_id": "connector_69221a1deda48191951c194e55da6855",
    "connector_name": "local",
    "action_name": "upload",
    "is_sync": false
}
```

Is it really possible to get a `{ text: string }` ?

src/web/hooks/use-tool-info.test.ts
```diff
@@ -28,21 +27,51 @@ describe("useToolInfo", () => {
...
     });
-    expect(result.current.output).toEqual({
-      name: "pikachu",
-      color: "yellow",
-      description:
-        "When several of these POKéMON gather, their\felectricity could build and cause lightning storms.",
-    });
-    expect(result.current.responseMetadata).toEqual({
-      id: 12,
+    expect(result.current.status).toEqual("pending");
```
harijoe: Maybe use object comparison to make it more readable ?

src/web/types.ts
fredericbarthelet: In order of priority, `toolOutput` contains:
- the content of `structuredContent` property in the tool response body
- if there is no `structuredContent` property, the content of the first text item in `content` property in the tool response body (in the form of `{text: ...}`
- if `content` is an empty array, the internal response object used by OpenAI (the one you share in your comment) 

I assumed the last case is more a bug from OpenAI implementation that something that's going to stay, but I might me mistaken

src/web/hooks/use-tool-info.ts
fredericbarthelet: It was exactly this implementation before. Or are you refering to something else?

harijoe: Well, you're right! I'm challenging my own implementation 😅

src/web/hooks/use-tool-info.ts
fredericbarthelet: I'm using the `useQuery` from `react-query` as reference implementation here. But I'm happy to source inspiration from other external fetching or state library. Can you share one that would be better to get inspired from?

In addition, this implementation still allows you to only rely on toolOuput value and check if it is null before using it if you prefere to use it this way. Is there a downside adding this boolean here?

src/web/hooks/use-tool-info.test.ts
fredericbarthelet: :+1 will update

src/web/hooks/use-tool-info.ts
harijoe: No downside I think, we can merge it ✅

I just find the value quite thin compared to react query; When using react-query, the call's result can be undefined so we can't rely on checking it and thus a boolean `isPending` is useful. 
Our case is a bit different as ToolOuput can't be at the same time successful and undefined/null.

harijoe: Food for thought : I noticed that some calls (like sendFollowUpMessage) will not work properly if the tool call has not resolved yet. Sometimes it's misleading because we can render a widget and continue the conversation before the tool even responds. There may be an opportunity to improve devX there, I will keep experimenting about it with my advent calendar app.

src/web/hooks/use-tool-info.ts
fredericbarthelet: I moved it outside of the return statement anyway :)

src/web/hooks/use-tool-info.ts
fredericbarthelet: > Our case is a bit different as ToolOuput can't be at the same time successful and undefined/null.

True for toolOuput, but not for toolResponseMetadata. It can be undefined, and can therefore benefit from the boolean (it was the case in my calendar app ^^)

Issue #37, opened 2025-12-01
Type API for useCallTool does not support tool without input args and with partial type definition


fredericbarthelet:
The following exemple shows that required implementation for a `play` tool with no input args and with only partial typing of the output (structuredContent and meta).

```ts
const { callTool: play, isPending } = useCallTool<
    // eslint-disable-next-line @typescript-eslint/no-empty-object-type
    {},
    // @ts-expect-error - TODO: fix this
    { structuredContent: { id: string }; meta: { word: string; illustrationUrl: string } }
>("play");
```
[Source](https://github.com/fredericbarthelet/advent-chatgpt-app/blob/4bfc5a01241e575ede1a4e845dcd66f28ee75874/web/src/widgets/play.tsx#L38-L43)

and in use:

```ts
play({});
```

2 updates are necessary on the hook generic type definition:
- allow for no input
- allow partial definition of tool ouput

## Target

```ts
const { callTool: play, isPending } = useCallTool<null, { structuredContent: { id: string }; meta: { word: string; illustrationUrl: string } }>("play");
```

and

```ts
play();
```

Issue #10, opened 2025-11-21
Improvements on hooks : missing lvl 2 and new lvl 3


fredericbarthelet:
## Level 2 hooks (UI only hooks not one-on-one with openai skybridge runtime)

- `useToolCall` -> already done. Need to add callbacks `onSettled` and `onError`.
- `useState` -> combination of `useState` from react, `useWidgetState` from skybridge - the idea is to provide
- `useLayout` -> proxy to set and get current with helpful boolean

### Level 3 hooks (UI + Conversation hooks)

- `useNavigation`

On click, use navigate + setWidgetState to persist the location of the user + gives context to the model of the actual shown detail page
navigate takes only the technical id
setWidgetState takes detailed information that might be relevant for the model to provide an answer for

has the added advantage of being a way to do progressive disclosure for the model

- `useFollowUp`

On click, use navigate + setWidgetState to persist the location of the user + gives context to the model of the actual shown detail page
navigate takes only the technical id
setWidgetState takes detailed information that might be relevant for the model to provide an answer for

PR #39, opened 2025-12-01
Use call tool on settled

harijoe:
Fixes https://github.com/alpic-ai/skybridge/issues/37
Contributes to https://github.com/alpic-ai/skybridge/issues/10

I could not avoid having to pass `null` as the first argument if only output needs to be typed, e.g. :
```
  const {
    callTool: play,
    isPending,
    data: playData,
  } = useCallTool<null, { structuredContent: { id: string }; meta: { word: string; illustrationUrl: string } }>("play");
  ```
  
  It types correctly the playData object though which is quite cool;

### CODE REVIEW

File: src/web/hooks/use-call-tool.ts [hunk 1]
```diff
@@ -60,18 +67,29 @@ export const useCallTool = <
...
     }
   ) => {
     callToolAsync(toolArgs)
       .then((data) => {
         if (sideEffects?.onSuccess) {
           sideEffects.onSuccess(data, toolArgs);
         }
+        if (sideEffects?.onSettled) {
+          sideEffects.onSettled(data, undefined, toolArgs);
+        }
```
fredericbarthelet, line 141: Inside a finally would avoid duplication ?

harijoe: We would have to store data/error somewhere (like in the sideEffects object) to be able to pass it to the `onSettled` callback so not so clean I think.
Duplication is admittedly verbose but the readibility remains good I think.

src/web/types.ts
```diff
@@ -70,11 +70,15 @@ export type CallToolResponse = {
   meta: Record<string, unknown>;
 };
 
+export type CallToolResponseConstraint = Partial<
+  Pick<CallToolResponse, "structuredContent" | "meta">
+>;
```
fredericbarthelet: this has nothing to do here - it's not exactly part of the openai API and is solely used by useCallTool. I'd put this definition over there

src/web/types.ts
```diff
@@ -70,11 +70,15 @@ export type CallToolResponse = {
...
+
 type API<WidgetState extends UnknownObject> = {
   /** Calls a tool on your MCP. Returns the full response. */
   callTool: <
     ToolArgs extends CallToolArgs = null,
-    ToolResponse extends CallToolResponse = CallToolResponse
+    ToolResponse extends CallToolResponseConstraint = CallToolResponse
```
fredericbarthelet: and this is not exactly true. The following becomes valid in a TS world:

```ts
const response = await callTool<null, {meta: string}>(...);

// { meta: string } only. Missing all other props from base CallToolResponse
typeof response
```

src/web/hooks/use-call-tool.ts
```diff
@@ -83,7 +98,7 @@ export const useCallTool = <
     isPending: status === "pending",
     isSuccess: status === "success",
     isError: status === "error",
-  } as CallToolState<ToolResponse>;
+  } as CallToolState<CallToolResponse & ToolResponse>;
```
fredericbarthelet: you can take a look at src/web/hooks/use-tool-info.ts to get an insight on how to do partial typing without overwritting all other types:

```ts
as ToolState<
    ToolSignature["input"] extends UnknownObject ? ToolSignature["input"] : {},
    ToolSignature["output"] extends UnknownObject
      ? ToolSignature["output"]
      : UnknownObject,
    ToolSignature["responseMetadata"] extends UnknownObject
      ? ToolSignature["responseMetadata"]
      : UnknownObject
  >;
```

src/web/types.ts
harijoe: But I need it in `type API<WidgetState extends UnknownObject> = {` definition below ?

src/web/hooks/use-call-tool.ts
harijoe: But don't we want to override types in this case ? Like don't we want to the CallToolState type to match both the generic CallToolResponse and the more-specific-user-provided ToolResponse at the same time ?

src/web/types.ts
```diff
@@ -50,7 +50,7 @@ export type OpenAiGlobals<
 
   // state
   toolInput: ToolInput;
-  toolOutput: ToolOutput | { text: string } | null;
+  toolOutput: ToolOutput | null;
```
fredericbarthelet: I think this change is inadverted

src/web/types.ts
```diff
@@ -78,7 +78,7 @@ type API<WidgetState extends UnknownObject> = {
   >(
     name: string,
     args: ToolArgs
-  ) => Promise<ToolResponse>;
+  ) => Promise<CallToolResponse & ToolResponse>;
```
fredericbarthelet: `ToolResponse` extends `CallToolResponse`, so the intersection of both those types is the same as `ToolResponse`. I don't get your change here, maybe as well an inadverted change?

src/web/hooks/use-tool-info.ts
```diff
@@ -2,47 +2,35 @@ import { useOpenAiGlobal } from "./use-openai-global.js";
...
+      responseMetadata: undefined;
+    }
+  | {
+      status: "success";
+      isPending: false;
+      isSuccess: true;
+      input: ToolInput;
+      output: ToolOutput;
+      responseMetadata: ToolResponseMetadata;
+    };
```
fredericbarthelet: same as my previous comment on useCallTool types -> fine with defining literal types for each state (instead of relying on `BaseToolState`) but keep the final union of types as a union of named types to improve readability on TS popup on hover

src/web/hooks/use-tool-info.test.ts
```diff
@@ -75,3 +83,126 @@ describe("useToolInfo", () => {
     });
   });
 });
+
+describe("useToolInfo - TypeScript typing", () => {
```
fredericbarthelet: According to vitest documentation

> By default all tests inside *.test-d.ts files are considered type tests, but you can change it with [typecheck.include](https://vitest.dev/config/#typecheck-include) config option.

source: https://vitest.dev/guide/testing-types

Might be worth splitting between runtime tests and type tests in 2 seperate files. WDYT?

src/web/hooks/use-tool-info.test.ts
```diff
@@ -75,3 +83,126 @@ describe("useToolInfo", () => {
...
+  it("should have correct types when no generic parameter is provided", () => {
+    const { result } = renderHook(() => useToolInfo());
```
fredericbarthelet: If you do test in `.test-d.ts` files, I think you don't even need renderHook anymore

src/web/hooks/use-tool-info.test.ts
```diff
@@ -75,3 +83,126 @@ describe("useToolInfo", () => {
...
+    );
+
+    expectTypeOf<TestInput>(result.current.input);
+
+    // When pending, output and responseMetadata should be undefined
+    if (result.current.status === "pending") {
+      expectTypeOf<undefined>(result.current.output);
+      expectTypeOf<undefined>(result.current.responseMetadata);
+      expectTypeOf<true>(result.current.isPending);
+      expectTypeOf<false>(result.current.isSuccess);
```
fredericbarthelet: i feel like this should be part of the first test - it's independent from generic types for input, output and metadata

src/web/hooks/use-tool-info.test.ts
```diff
@@ -75,3 +83,126 @@ describe("useToolInfo", () => {
...
+    }
+
+    // When success, output and responseMetadata should be defined
+    if (result.current.status === "success") {
+      expectTypeOf<TestOutput>(result.current.output);
+      expectTypeOf<TestMetadata>(result.current.responseMetadata);
+      expectTypeOf<false>(result.current.isPending);
+      expectTypeOf<true>(result.current.isSuccess);
```
fredericbarthelet: same here

src/web/hooks/use-call-tool.ts
```diff
@@ -1,51 +1,86 @@
...
 };
 
-type IdleCallToolState = BaseCallToolState<"idle">;
-type PendingCallToolState = BaseCallToolState<"pending">;
-type SuccessCallToolState<TData extends CallToolResponse = CallToolResponse> =
-  BaseCallToolState<"success", TData>;
-type ErrorCallToolState = BaseCallToolState<"error">;
+type CallToolAsyncFn<TA, TR> = TA extends null
+  ? () => Promise<TR>
+  : (toolArgs: TA) => Promise<TR>;
```
fredericbarthelet: Can you use full name for generics in order to improve readability ?

```suggestion
type CallToolAsyncFn<ToolArgs, ToolResponse> = ToolArgs extends null
  ? () => Promise<ToolResponse>
  : (toolArgs: ToolArgs) => Promise<ToolResponse>;
```

src/web/hooks/use-call-tool.ts
```diff
@@ -1,51 +1,86 @@
...
-  | ErrorCallToolState;
+type CallToolResponseConstraint = Partial<
+  Pick<CallToolResponse, "structuredContent" | "meta">
+>;
 
 export const useCallTool = <
   ToolArgs extends CallToolArgs = null,
-  ToolResponse extends CallToolResponse = CallToolResponse
+  ToolResponse extends CallToolResponseConstraint = CallToolResponseConstraint
 >(
```
fredericbarthelet: For coherence with useToolInfo type :)?


```suggestion
type ToolResponseSignature = Pick<CallToolResponse, "structuredContent" | "meta">;

export const useCallTool = <
  ToolArgs extends CallToolArgs = null,
  ToolResponse extends Partial<ToolResponseSignature> = {}
>(
```

src/web/hooks/use-call-tool.ts
```diff
@@ -1,51 +1,86 @@
...
   name: string
 ) => {
+  type CombinedCallToolResponse = CallToolResponse & ToolResponse;
```
fredericbarthelet, line 75: Again, for the sake of coherence, I'd go for 

```ts
  type StructuredContent = TS["structuredContent"] extends CallToolResponse["structuredContent"] ? TS["structuredContent"] : CallToolResponse["structuredContent"];
// ...
```

or edit the useToolInfo implementation to go for intersection

src/web/hooks/use-call-tool.ts
```diff
@@ -55,25 +90,54 @@ export const useCallTool = <
...
+  ): void;
+  function callTool(
     toolArgs: ToolArgs,
-    sideEffects?: {
-      onSuccess?: (data: ToolResponse, toolArgs: ToolArgs) => void;
-      onError?: (error: unknown, toolArgs: ToolArgs) => void;
+    sideEffects?: SideEffects<ToolArgs, CombinedCallToolResponse>
+  ): void;
+  function callTool(
+    arg1?: ToolArgs | SideEffects<ToolArgs, CombinedCallToolResponse>,
```
fredericbarthelet:
```suggestion
    firstArg?: ToolArgs | SideEffects<ToolArgs, CombinedCallToolResponse>,
```

src/web/types.ts
harijoe: Indeed!

src/web/hooks/use-call-tool.ts
harijoe: I chose to edit useToolInfo

PR #41, opened 2025-12-02
Enable components to provide llm context

harijoe:
# Enable components to provide llm context
### Summary

This PR adds a system that lets React components provide structured context to the LLM. Developers can annotate JSX elements with a new `llm-describe` attribute, which is compiled into an `LLMDescribe` wrapper component. All descriptions form a hierarchical tree that is serialized into an indented bullet-list and stored in `window.openai.widgetState.__widget_context`.

### Key changes

- **LLMDescribe component** (`src/web/llm-describe.tsx`)  
  Tracks nodes by `id`/`parentId`, builds a tree, and updates `__widget_context` whenever annotated elements mount, update, or unmount.

- **Babel/Vite plugin** (`src/server/describe-plugin.ts`)  
  Detects `llm-describe`, removes the attribute, wraps the element in `<LLMDescribe content={...}>`, and auto-adds imports.

- **Widget state handling** (`src/web/hooks/use-widget-state.ts`)  
  Keeps `__widget_context` inside `window.openai.widgetState` but filters it out of the React-visible widget state.

- **Type support** (`src/web/types.ts`)  
  Extends React HTML attributes to include `"llm-describe"`.

### Usage example

```tsx
<div llm-describe="User card">...</div>

### DIFF
--- a/src/web/hooks/use-widget-state.ts
+++ b/src/web/hooks/use-widget-state.ts
@@ -18,7 +18,7 @@ function filterWidgetContext<T extends UnknownObject>(
   return filteredState as T;
 }
 
-function persistWidgetContext<T extends UnknownObject>(
+function preserveWidgetContext<T extends UnknownObject>(
   newState: T | null
 ): T | null {
   if (newState === null) {
@@ -75,7 +75,7 @@ export function useWidgetState<T extends UnknownObject>(
   const setWidgetState = useCallback((state: SetStateAction<T | null>) => {
     _setWidgetState((prevState) => {
       const newState = typeof state === "function" ? state(prevState) : state;
-      const stateToSet = persistWidgetContext(newState);
+      const stateToSet = preserveWidgetContext(newState);
 
       if (stateToSet !== null) {
         window.openai.setWidgetState(stateToSet);

### CODE REVIEW

**Comment by fredericbarthelet** [comment]
I like your first scaffolding of this feature to feedback context on what's in the UI back to the LLM for better synchronization.

The component self-responsability to expose what should be part of the context is brilliant. The nesting of all those exposed context to go in a single interaction back to the LLM while preserving structure imposed by UI components is great.

My initial believe when seeing your first implementation was that the **children** of the `LLMDescribe` was going to be serialized somehow and sent back to the model, having the actual rendered DOM as the `content` prop without requiring the developper to do this manually.

Example of this would be :

```ts
<Card>
  <CardHeader>
    <LLMDescribe>
      <h1>Trip to marrakech for 2</h1>
    </LLMDescribe>
  </CardHeader>
  <CardBody>
    //....
  </CardBody>
</Card>
```
that would result in exposing something like `# Trip to marrakech for 2` (since LLM are good at markdown reading, serializing DOM to markdown could be a good option).

I understand from your proposal now that the intent was not to use the `children` prop of `LLMDescribe` but an explicit property `content` to be hydrated by the developper, leaving room for what and how he wants to serialize the info back to the LLM.

```ts
<Card>
  <CardHeader>
    <LLMDescribe content="marrakech">
      <h1>Trip to marrakech for 2</h1>
    </LLMDescribe>
  </CardHeader>
  <CardBody>
    //....
  </CardBody>
</Card>
```

If we go down that route, I'd argue we don't even need a custom component. This property could be set on any React node, in a similar fashion to `aria` properties to help a11y.

```ts
<Card>
  <CardHeader>
    <h1 llm-context="marrakech">Trip to marrakech for 2</h1>
  </CardHeader>
  <CardBody>
    //....
  </CardBody>
</Card>
```

This keeps the API simple, it does not add another indentation level to React component structure, it mimicks a concept developper are already familiar with.

WDYT?

src/web/hooks/use-widget-state.ts
fredericbarthelet: if you just use

```suggestion
  const { [WIDGET_CONTEXT_KEY]: widgetContext, ...widgetStateFromWindow }  = useOpenAiGlobal("widgetState") as T;
```

and persist both or return only one of them later on, don't you simplify your implementation?

src/web/hooks/use-widget-state.ts
```diff
@@ -26,24 +68,22 @@ export function useWidgetState<T extends UnknownObject>(
...
+  const setWidgetState = useCallback((state: SetStateAction<T | null>) => {
+    _setWidgetState((prevState) => {
+      const newState = typeof state === "function" ? state(prevState) : state;
+      const stateToSet = preserveWidgetContext(newState);
 
-        if (newState !== null) {
-          window.openai.setWidgetState(newState);
-        }
+      if (stateToSet !== null) {
+        window.openai.setWidgetState(stateToSet);
```
fredericbarthelet, line 81: Here would be the new implementation:

```suggestion
        window.openai.setWidgetState({ newState, [WIDGET_CONTEXT_KEY]: widgetContext });
```

src/web/llm-describe.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+const ParentIdContext = createContext<string | null>(null);
+
+interface LLMDescribeProps {
+  content: LLMDescribeContent | null | undefined;
+  children?: ReactNode;
+}
+
+export function LLMDescribe({ content, children }: LLMDescribeProps) {
```
fredericbarthelet: Just a question - shouldn't we align `LLMDescribe` and `llm-data` in terms of naming?

src/web/hooks/use-widget-state.ts
harijoe: But we would not handle the case when `useOpenAiGlobal("widgetState")` returns null (when initialized for example) ?

src/web/llm-describe.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const parentId = useContext(ParentIdContext);
+  const id = useId();
+
+  useEffect(() => {
+    setNode({
+      id,
+      parentId,
+      content: content ?? null,
```
fredericbarthelet: can you really set null on the content prop then? I feel like there is no benefit from tiggering `onChange` after registering a new node who's content is null.

src/web/llm-describe.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+export type LLMDescribeContent = string;
+
+export interface LLMDescribeNode {
+  id: string;
+  parentId: string | null;
+  content: string | null;
+}
+
+const nodes = new Map<string, LLMDescribeNode>();
```
fredericbarthelet: not sure the Map is the most efficient way to store your structure data. Tree instead maybe?

src/web/plugin/transform-data-llm.ts
fredericbarthelet, line 1: why use babel for this? I was under the impression vite is using esbuild for dev and rollup for prod, but I've never implemented such thing before

src/web/llm-describe.tsx
harijoe: I could not find a good lib to do this. I think Map is good enough for our usage and should not be a bottleneck.

src/web/plugin/transform-data-llm.ts
harijoe: Not sure what to do : Would a regex-based implementation be better to avoid the babel/core dependency ?

src/web/hooks/use-widget-state.ts
fredericbarthelet: Then there is an issue on type casting here because it's casted as `T`.

src/web/plugin/data-llm.ts
```diff
@@ -0,0 +1,141 @@
+import type { PluginObj, TransformOptions, types } from "@babel/core";
+
+const LLM_IMPORT_SOURCE = "skybridge/web";
+
+interface State {
+  hasLLMDescribeImport?: boolean;
+  needsLLMDescribeImport?: boolean;
```
fredericbarthelet:
```suggestion
  hasLLMDataImport?: boolean;
  needsLLMDataImport?: boolean;
```

src/web/plugin/data-llm.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        const llmAttrIndex = attrs.findIndex(
+          (attr) =>
+            t.isJSXAttribute(attr) &&
+            t.isJSXIdentifier(attr.name, { name: "data-llm" })
+        );
+
+        if (llmAttrIndex === -1) return;
+
+        const llmAttr = attrs[llmAttrIndex] as types.JSXAttribute;
```
fredericbarthelet: Please avoid abbreviation except for known well-documented single letter API (like zod's `z` or 
```suggestion
        const dataLLMAttribute = attributes.find(
          (attribute) =>
            t.isJSXAttribute(attribute) &&
            t.isJSXIdentifier(attribute.name, { name: "data-llm" })
        );

        if (dataLLMAttribute === undefined) return;
```

src/web/plugin/data-llm.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        const newAttrs = [...attrs];
+        newAttrs.splice(llmAttrIndex, 1);
+        opening.attributes = newAttrs;
```
fredericbarthelet: why do you need to remove the attribute from the generated DOM? Isn't better to keep it?

src/web/plugin/data-llm.ts
```diff
@@ -0,0 +1,141 @@
...
+
+        let contentExpression: types.Expression;
+
+        if (!llmAttr.value) {
+          contentExpression = t.stringLiteral("");
```
fredericbarthelet: should we really build an LLMData component with an empty string in that case?

src/web/hooks/use-widget-state.ts
harijoe: I prefer to have the filtering function as a dedicated function, I think it's easier to maintain if we happen to need another special key in the future ?

src/web/plugin/data-llm.ts
harijoe: With the current implementation, Babel loops if we don't remove the data-llm attribute

src/web/hooks/use-widget-state.ts
```diff
@@ -2,6 +2,48 @@ import { useCallback, useEffect, useState, type SetStateAction } from "react";
...
+  }
+
+  const { [WIDGET_CONTEXT_KEY]: _, ...filteredState } = state as T & {
+    [WIDGET_CONTEXT_KEY]?: unknown;
+  };
+
+  return filteredState as T;
+}
+
+function persistWidgetContext<T extends UnknownObject>(
```
fredericbarthelet: sorry, I removed my comment beforehand, you were right: "preserve" was better than "persist" here :)

src/web/hooks/use-widget-state.ts
fredericbarthelet: Ok

src/web/plugin/transform-data-llm.ts
fredericbarthelet: let's stick with babel here :)

fredericbarthelet (approved):
LGTM ! Thanks @harijoe - this is huge!

Issue #43, opened 2025-12-02
Current test command masks vitest warning


fredericbarthelet:
The test command in package.json is `pnpm vitest --silent`

This prevents developper from seeing important warning on React state updates issue.

```
 skybridge git:(upload-api) ✗ pnpm exec vitest

 DEV  v2.1.9 /Users/frederic/Sites/skybridge

stderr | src/web/hooks/use-call-tool.test.ts > useCallTool - onSuccess callback > should call window.openai.callTool with correct arguments
An update to TestComponent inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */

This ensures that you're testing the behavior the user would see in the browser. Learn more at https://react.dev/link/wrap-tests-with-act

stderr | src/web/hooks/use-call-tool.test.ts > useCallTool - TypeScript typing > should have correct return types when ToolArgs is null and ToolResponse is specified
An update to TestComponent inside a test was not wrapped in act(...).

When testing, code that causes React state updates should be wrapped into act(...):

act(() => {
  /* fire events that update state */
});
/* assert on the output */
// ...
```

I'd recommend fixing the issues - currently 4 of them, all in `src/web/hooks/use-call-tool.test.ts`
And then updating the command to its simplest implementation: `pnpm vitest`

PR #48, opened 2025-12-03
Remove silent flag from test run and fix warnings

harijoe:
Fixes https://github.com/alpic-ai/skybridge/issues/43

### DIFF
--- a/src/web/hooks/use-call-tool.test.ts
+++ b/src/web/hooks/use-call-tool.test.ts
@@ -41,7 +41,7 @@ describe("useCallTool - onSuccess callback", () => {
     const { result } = renderHook(() =>
       useCallTool<typeof args, typeof data>(toolName)
     );
-    act(() => {
+    await act(async () => {
       result.current.callTool(args);
     });
     expect(OpenaiMock.callTool).toHaveBeenCalledWith(toolName, args);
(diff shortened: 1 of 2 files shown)


src/web/hooks/use-call-tool.test.ts
```diff
@@ -41,7 +41,7 @@ describe("useCallTool - onSuccess callback", () => {
     const { result } = renderHook(() =>
       useCallTool<typeof args, typeof data>(toolName)
     );
-    act(() => {
+    await act(async () => {
```
fredericbarthelet, line 44: can you explain why this is working @harijoe 😅?
`callTool` is not async so I don't really understand this

harijoe: Making the act callback async makes it return a Promise, so React enters async-act mode.
In this mode, act waits for all async updates triggered by the hook.
Because React waits for those updates to settle, there’s no warning, and we await act so that expectations like expect(OpenaiMock.callTool) run after those updates have occurred.

fredericbarthelet: so basically any function that are "fire and forget" but whose assertions depends on the accomplishment of the "forget" part requires this switch to async mode, right?

PR #52, opened 2025-12-04
Improve docs

fredericbarthelet:


docs/docs/api-reference/useCallTool.md
```diff
@@ -232,6 +217,28 @@ function SubmitForm() {
 }
 ```
 
+### Tools without arguments
+
+You can pass side effect options as the first argument to the `callTool` function if it doesn't require any arguments.
```
harijoe, line 222: This is not what's shown in the code example below. At the moment our api allow `onSuccess` to be set on `callTool` function, not as the hook second argument like react-query api.

harijoe (changes requested):

docs/docs/api-reference/useCallTool.md
fredericbarthelet: Good catch! Just changed it

PR #59, opened 2025-12-04
Add documentations on all hooks

fredericbarthelet:


docs/docs/api-reference/use-display-mode.md
```diff
@@ -0,0 +1,128 @@
+---
+sidebar_position: 2
+---
+
+# useDisplayMode
+
+The `useDisplayMode` hook allows you to read and control the display mode of your widget. Widgets can be displayed in three modes: `inline`, `fullscreen`, or `pip` (picture-in-picture).
```
harijoe, line 7: Should we also mention the `modal` mode here ?

docs/docs/api-reference/use-display-mode.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+## Returns
+
+The hook returns a tuple with two elements:
+
+### `displayMode`
+
+```tsx
+displayMode: "inline" | "fullscreen" | "pip"
```
harijoe, line 38: We should definitely add the `modal` mode here

docs/docs/api-reference/use-display-mode.md
fredericbarthelet: There is a modal mode ?!?
Only saw mention of inline, fullscreen and pip in documentation (https://developers.openai.com/apps-sdk/build/chatgpt-ui#understand-the-windowopenai-api)

image

harijoe: Yes, when requestModal is triggered, displayMode changes to value "modal"

harijoe: image

Issue #51, opened 2025-12-04
Emulation local for development environment


fredericbarthelet:
## Introduction

Feedback loop developing on ChatGPT itself feels slow. Having a faster environment to iterate on would greatly improve DX.
This issue aims at investigating multiple solutions involving more or less emulation of ChatGPT in a dedicated local dev environment (by opposition to a remote dev environment).

## Proposed implementation

Any level of emulation require serving a dedicated local dev env on a separate endpoint (could be a different host, different port, different path).

## Options

- No emulation: developing directly on ChatGPT (current implementation)
- Partial emulation (no LLM involved):
  - rendering widget based on tool response (à la MCPJam), but without additional emulation. This provides everything necessary for initial rendering of the component and allows fast iteration (with all methods stubbed).
  - rendering widget based on tool response + provide a few additional APIs (to be decided)
- Full emulation - bringing a complete copy of ChatGPT in local dev including a model

## Criterion to consider

- Cost of development / cost of maintenance
- Added value compared to existing solutions (if we do MCPJam like emulation, what's our value?)

harijoe:
@rodincave Is this what you had in mind?

PR #88, opened 2025-12-12
Mcp emulator

harijoe:
image

### Summary

This PR provides a new emulator to make iterating quickly on mcp apps easier.

### Shortcomings

- Some borders still need to be resizable
- Case when mcp server is not reachable is not handled properly (the "connected" indicator is static)
- The properties panel misses many of what the openai object provides
- Widget state is not editable, neither is tool output : we'd like both of them to be editable and the widget to reflect their updated values
- setWidgetState is trigered multiple times in different testing scenarios such as when `data-llm` is used. It may be a bug with `data-llm` implementation : this needs more investigation
- The case when the mcp server does not provide any tool is not handled
- Emulator dark mode is disabled : it requires some design fixes around reactJsonViewer
- The openai locale list must be properly implemented, at the moment it's just a text-input
- The generated layout of the input form when the selected tool requires multiple inputs must be checked/fixed
- Visual hierarchy of the whole emulator is currently not ideal and could be significantly improved
- A widget refresh button could be handy to mock a page reload on chatgpt conversation page
- Tool's inputs are not persisted if page is reloaded : if the persisted input is still compatible with the tool input schema, it would be handy to fill the tool input form automatically
- Visual animation of the landing is starting only after a few seconds, it would look better by animating right after render
- The message shown when skybridge is started should be updated to showcase the emulator and skybridge instead of showing nodemon default messages
- Consider implementing openai maxHeight to provide an easy way to check how the widget renders with different heights

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

Implements a standalone MCP emulator application with React frontend for testing and development. The emulator provides a web-based interface to connect to MCP servers, list available tools, and execute them with visual feedback.

**Major changes:**
- New React-based emulator application in `src/emulator/` with full UI for MCP tool testing
- Added `emulatorStaticServer` to serve the built emulator application
- Modified `mountWidget` to wait for OpenAI object initialization before mounting
- Added "modal" display mode to `DisplayMode` type
- Configured workspace to include emulator as a separate package

**Issues found:**
- `.env` file committed to repository without being in `.gitignore` (should be removed from git history)
- Debug `console.log` statement left in production code

- This PR is mostly safe to merge with a few issues that should be addressed
- The implementation is well-structured and adds valuable development tooling. However, the `.env` file being committed to the repository is a security concern that needs to be fixed before merging. The code quality is good overall with proper error handling and type safety, but includes a leftover debug statement.
- Pay attention to `src/emulator/.env` and `src/emulator/.gitignore` to ensure environment files are properly excluded from version control

### DIFF
--- a/packages/devtools/src/App.tsx
+++ b/packages/devtools/src/App.tsx
@@ -1,7 +1,7 @@
 import { QueryClientProvider } from "@tanstack/react-query";
+import { NuqsAdapter } from "nuqs/adapters/react";
 import AppLayout from "./components/layout/app-layout.js";
 import { queryClient } from "./lib/query-client.js";
-import { NuqsAdapter } from "nuqs/adapters/react";
 
 function App() {
   return (
--- a/packages/devtools/src/lib/mcp/index.ts
+++ b/packages/devtools/src/lib/mcp/index.ts
@@ -1,9 +1,9 @@
-import { useStore } from "@/lib/store.js";
 import type { Tool } from "@modelcontextprotocol/sdk/types.js";
 import { useMutation, useQuery, useSuspenseQuery } from "@tanstack/react-query";
 import type { CallToolArgs, OpenAiProperties } from "skybridge/web";
-import { McpClient } from "./client.js";
+import { useStore } from "@/lib/store.js";
 import { useSelectedToolName } from "../nuqs.js";
+import { McpClient } from "./client.js";
 
 const client = new McpClient();
 
--- a/packages/devtools/src/lib/store.ts
+++ b/packages/devtools/src/lib/store.ts
@@ -1,4 +1,4 @@
-import { set, cloneDeep } from "lodash-es";
+import { cloneDeep, set } from "lodash-es";
 import type {
   CallToolResponse,
   OpenAiProperties,
--- a/packages/devtools/src/components/layout/header.tsx
+++ b/packages/devtools/src/components/layout/header.tsx
@@ -1,6 +1,6 @@
+import { ExternalLink } from "lucide-react";
 import { useServerInfo } from "@/lib/mcp";
 import { useSelectedToolName } from "@/lib/nuqs";
-import { ExternalLink } from "lucide-react";
 import { Button } from "../ui/button";
 
 export const Header = () => {
--- a/packages/devtools/src/components/layout/tool-panel/openai-inspector.tsx
+++ b/packages/devtools/src/components/layout/tool-panel/openai-inspector.tsx
@@ -1,20 +1,20 @@
+import ReactJsonView from "@microlink/react-json-view";
+import { MoonIcon, SunIcon } from "lucide-react";
 import { Button } from "@/components/ui/button";
 import { ButtonGroup } from "@/components/ui/button-group";
 import { Field, FieldLabel, FieldSet } from "@/components/ui/field";
 import { Input } from "@/components/ui/input";
 import {
   Select,
-  SelectItem,
+  SelectContent,
   SelectGroup,
+  SelectItem,
   SelectTrigger,
   SelectValue,
-  SelectContent,
 } from "@/components/ui/select";
 import { Tabs, TabsContent, TabsList, TabsTrigger } from "@/components/ui/tabs";
 import { useSelectedTool } from "@/lib/mcp";
 import { useCallToolResult, useStore } from "@/lib/store";
-import ReactJsonView from "@microlink/react-json-view";
-import { MoonIcon, SunIcon } from "lucide-react";
 
 export const OpenAiInspector = () => {
   const tool = useSelectedTool();
--- a/packages/devtools/src/components/layout/tool-panel/tool-panel.tsx
+++ b/packages/devtools/src/components/layout/tool-panel/tool-panel.tsx
@@ -1,3 +1,4 @@
+import { Suspense } from "react";
 import {
   ResizableHandle,
   ResizablePanel,
@@ -6,7 +7,6 @@ import {
 import { useResizablePanelSize } from "@/hooks/use-resizable-panel-size";
 import { useSelectedTool } from "@/lib/mcp/index.js";
 import { useCallToolResult } from "@/lib/store.js";
-import { Suspense } from "react";
 import { InputForm } from "./input-form.js";
 import { OpenAiInspector } from "./openai-inspector.js";
 import { OpenAiLogs } from "./openai-logs.js";
--- a/packages/devtools/src/components/layout/tool-panel/widget/widget.tsx
+++ b/packages/devtools/src/components/layout/tool-panel/widget/widget.tsx
@@ -1,6 +1,6 @@
+import { useCallback, useRef } from "react";
 import { useSelectedTool, useSuspenseResource } from "@/lib/mcp";
 import { useCallToolResult, useStore } from "@/lib/store";
-import { useCallback, useRef } from "react";
 import { createAndInjectOpenAi } from "./create-openai-mock";
 import { injectWaitForOpenai } from "./utils";
 
--- a/packages/devtools/src/components/ui/button-group.tsx
+++ b/packages/devtools/src/components/ui/button-group.tsx
@@ -1,9 +1,9 @@
-import { mergeProps } from "@base-ui/react/merge-props"
-import { useRender } from "@base-ui/react/use-render"
-import { cva, type VariantProps } from "class-variance-authority"
+import { mergeProps } from "@base-ui/react/merge-props";
+import { useRender } from "@base-ui/react/use-render";
+import { cva, type VariantProps } from "class-variance-authority";
 
-import { cn } from "@/lib/utils"
-import { Separator } from "@/components/ui/separator"
+import { Separator } from "@/components/ui/separator";
+import { cn } from "@/lib/utils";
 
 const buttonGroupVariants = cva(
   "rounded-none has-[>[data-slot=button-group]]:gap-2 has-[select[aria-hidden=true]:last-child]:[&>[data-slot=select-trigger]:last-of-type]:rounded-none flex w-fit items-stretch [&>*]:focus-visible:z-10 [&>*]:focus-visible:relative [&>[data-slot=select-trigger]:not([class*='w-'])]:w-fit [&>input]:flex-1",
@@ -19,8 +19,8 @@ const buttonGroupVariants = cva(
     defaultVariants: {
       orientation: "horizontal",
     },
-  }
-)
+  },
+);
 
 function ButtonGroup({
   className,
@@ -28,6 +28,7 @@ function ButtonGroup({
   ...props
 }: React.ComponentProps<"div"> & VariantProps<typeof buttonGroupVariants>) {
   return (
+    // biome-ignore lint/a11y/useSemanticElements: generated file from shadcn/ui
     <div
       role="group"
       data-slot="button-group"
@@ -35,7 +36,7 @@ function ButtonGroup({
       className={cn(buttonGroupVariants({ orientation }), className)}
       {...props}
     />
-  )
+  );
 }
 
 function ButtonGroupText({
@@ -49,16 +50,16 @@ function ButtonGroupText({
       {
         className: cn(
           "bg-muted gap-2 rounded-none border px-2.5 text-xs font-medium [&_svg:not([class*='size-'])]:size-4 flex items-center [&_svg]:pointer-events-none",
-          className
+          className,
         ),
       },
-      props
+      props,
     ),
     render,
     state: {
       slot: "button-group-text",
     },
-  })
+  });
 }
 
 function ButtonGroupSeparator({
@@ -72,11 +73,11 @@ function ButtonGroupSeparator({
       orientation={orientation}
       className={cn(
         "bg-input relative self-stretch data-[orientation=horizontal]:mx-px data-[orientation=horizontal]:w-auto data-[orientation=vertical]:my-px data-[orientation=vertical]:h-auto",
-        className
+        className,
       )}
       {...props}
     />
-  )
+  );
 }
 
 export {
@@ -84,4 +85,4 @@ export {
   ButtonGroupSeparator,
   ButtonGroupText,
   buttonGroupVariants,
-}
+};
--- a/packages/devtools/src/components/ui/card.tsx
+++ b/packages/devtools/src/components/ui/card.tsx
@@ -1,4 +1,4 @@
-import * as React from "react";
+import type * as React from "react";
 
 import { cn } from "@/lib/utils";
 
--- a/packages/devtools/src/components/ui/field.tsx
+++ b/packages/devtools/src/components/ui/field.tsx
@@ -1,9 +1,9 @@
-import { useMemo } from "react";
 import { cva, type VariantProps } from "class-variance-authority";
+import { useMemo } from "react";
 
-import { cn } from "@/lib/utils";
 import { Label } from "@/components/ui/label";
 import { Separator } from "@/components/ui/separator";
+import { cn } from "@/lib/utils";
 
 function FieldSet({ className, ...props }: React.ComponentProps<"fieldset">) {
   return (
@@ -227,13 +227,13 @@ function FieldError({
 
 export {
   Field,
-  FieldLabel,
+  FieldContent,
   FieldDescription,
   FieldError,
   FieldGroup,
+  FieldLabel,
   FieldLegend,
   FieldSeparator,
   FieldSet,
-  FieldContent,
   FieldTitle,
 };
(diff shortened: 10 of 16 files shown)


pnpm-workspace.yaml
```diff
@@ -1,3 +1,3 @@
 packages:
   - "docs"
-
```
fredericbarthelet, line 3: I'd like to target a root `packages` folder with every individual packages. I understand that this emulator is a package of his own, could you move it outside of `src` ?
In fine, the whole `src` folder will disappear but I don't think this should be part of this already mega long PR 😅

package.json
```diff
@@ -17,8 +17,9 @@
     }
   },
   "scripts": {
-    "build": "tsc && pnpm run build:templates",
+    "build": "tsc && pnpm run build:templates && pnpm run build:emulator",
     "build:templates": "cp -r src/server/templates dist/src/server/",
+    "build:emulator": "pnpm --filter emulator build && cp -r src/emulator/dist dist/src/emulator/",
```
fredericbarthelet: I think this should be packaged and used as a separate dependency by end user. Packaging emulator in the current package of skybridge means people only using hooks will still go in prod with a dependency code base containing assets required for a development emulator.

I mainly see the issue coming where the emulator packages his own version of the @modelcontextprotocol/sdk package while skybridge ask for the user to BYO mcp SDK to avoid conflicting versions

biome.json
```diff
@@ -18,15 +18,28 @@
     "rules": {
       "recommended": true,
       "correctness": {
-        "noUnusedVariables": "error"
-      }
+        "noUnusedVariables": "error",
+        "useExhaustiveDependencies": "warn"
+      },
+      "style": {
+        "noNonNullAssertion": "warn"
```
fredericbarthelet: This is already handled by #90
I'd remove all changes made to `biome.json` and keep this PR focused on the emulator to avoid side effect changes and keep everything related only to the added feature :)

src/web/types.ts
```diff
@@ -133,7 +133,7 @@ export type CallTool = (
   args: Record<string, unknown>,
 ) => Promise<CallToolResponse>;
 
-export type DisplayMode = "pip" | "inline" | "fullscreen";
+export type DisplayMode = "pip" | "inline" | "fullscreen" | "modal";
```
fredericbarthelet: This is already handled by #92

src/web/mount-widget.ts
```diff
@@ -6,7 +6,40 @@ import { installOpenAILoggingProxy } from "./proxy.js";
...
+          writable: true,
+          value,
+        });
+        resolve(undefined);
+      },
+    });
+  });
+};
+
+export const mountWidget = async (component: React.ReactNode) => {
```
fredericbarthelet: Ideally, `mountWidget` shouldn't have to change to accommodate for the emulator. This feels like changing prod for a dev feature. Can't you ensure `window.openai` is always defined on the emulator side?

I suggested in #90 to implement throwing in case this property in missing (the component is not rendered in OpenAI environment), this ensure it never throws which is counter-productive for the other implementation.

In addition, we're going to relax the interface here to support MCP Apps as well, so `window.openai` will not always be set, and should not be mocked in case it is absent (when rendered in an MCP App host for example).

src/server/emulatorStaticServer.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Serve the built emulator React app
+ * This router serves static files from the emulator's dist directory.
+ * It should be installed at the application root, like so:
+ *
+ *  const app = express();
+ *
+ * if (env.NODE_ENV !== "production") {
+ *   app.use(await emulatorStaticServer(server));
```
fredericbarthelet: the doc says you should have the server as param but not the actual implementation

docs/docs/devtools/devtools.md
fredericbarthelet, line 1: With docs localized in the docs folder, I'm not sure it's a good practice to keep having READMEs at each package root.

src/emulator/package.json
```diff
@@ -0,0 +1,63 @@
...
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "lint": "eslint .",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "@base-ui-components/react": "1.0.0-rc.0",
```
fredericbarthelet: release candidate in prod env ?

packages/devtools/src/App.tsx
fredericbarthelet, line 1: No routing is intentional? I feel this would sometime be more relevant that store and survive refresh :) Or you wanted to keep the express router for the emulator contained to `/` only?

pnpm-workspace.yaml
harijoe: I agree, it felt weird to put it in src, I have the same issue with the create-skybridge command

package.json
harijoe: Something like `@skybridge/emulator` ?

biome.json
harijoe: This will require some time-consuming refactoring to get rid of every non-null assertions though.

src/web/types.ts
harijoe: Waiting for it to be merged to rebase on

src/web/mount-widget.ts
harijoe: This is challenging. This check is an arbitrage I did to avoid having to define two iframes, let's discuss it.

src/server/emulatorStaticServer.ts
harijoe: Indeed, the docs are outdated. I used to require the server as a param but it's not the case anymore. I'll fix it.

src/emulator/package.json
harijoe: I'll look into that

packages/devtools/src/App.tsx
harijoe: Yes I wanted to disturb as little as possible the express server that's why I chose not to use routes. Open to discussion.

pnpm-workspace.yaml
fredericbarthelet: Target is everything in `packages` without any remaining root `src` just like next.js repo

package.json
fredericbarthelet: -> `@skybridge/devtools`

src/server/emulatorStaticServer.ts
fredericbarthelet: You still have the issue where the doc is misaligned with the interface implementation. Might be worth removing the doc altogether to avoid this issue. WDYT?

packages/devtools/.env.example
```diff
@@ -0,0 +1 @@
+VITE_MCP_SERVER_URL=http://localhost:3000/mcp
```
fredericbarthelet: Feels strange to have a `.env.example` - nobody should be installing the project locally expect to contribute, no?

I feel like this emulator should only be accessible when at some point the dev env will be started with a `skybridge start`. This will have the effect of starting a local server packaging whatever our entry point will be (might still require a server entry). Bridging the emulator on the started server should not require additional config to know the server to interact with - it's the one it's mounted on.

packages/devtools/src/lib/store.ts
```diff
@@ -0,0 +1,89 @@
...
+  setToolData: (tool: string, data: Partial<ToolData>) => void;
+  pushOpenAiLog: (tool: string, log: Omit<OpenAiLog, "id">) => void;
+  updateOpenaiObject: (
+    tool: string,
+    key: keyof OpenAiProperties,
+    value: unknown,
+  ) => void;
+};
+
+export const useStore = create<Store>()((set) => ({
```
fredericbarthelet: There is a typo here no? only createStore does not require any args
```suggestion
export const useStore = create<Store>((set) => ({
```

packages/devtools/src/lib/store.ts
```diff
@@ -0,0 +1,89 @@
...
+        tools: {
+          ...state.tools,
+          [tool]: {
+            ...state.tools[tool],
+            openaiObject: updatedOpenaiObject,
+          },
+        },
+      };
+    }),
+  getToolData: (tool: string) => (state: Store) => state.tools[tool],
```
fredericbarthelet: isn't this a duplicate with `useCallToolResult` ?

packages/devtools/src/lib/mcp/index.ts
```diff
@@ -0,0 +1,114 @@
...
+      right: 0,
+    },
+  },
+  toolInput: {},
+  toolOutput: null,
+  toolResponseMetadata: null,
+  widgetState: null,
+};
+
+export const useTools = () => {
```
fredericbarthelet:
```suggestion
export const useSuspenseTools = () => {
```

or replace use of suspense with useQuery directly. This ensures people know the hook should be wrapped in Suspense tag to handle loading state

packages/devtools/src/lib/mcp/index.ts
```diff
@@ -0,0 +1,114 @@
...
+  const { data } = useSuspenseQuery<Tool[]>({
+    queryKey: ["list-tools"],
+    queryFn: async () => await client.listTools(),
```
fredericbarthelet:
```suggestion
    queryFn: () => client.listTools(),
```

packages/devtools/src/lib/mcp/index.ts
```diff
@@ -0,0 +1,114 @@
...
+  },
+  toolInput: {},
+  toolOutput: null,
+  toolResponseMetadata: null,
+  widgetState: null,
+};
+
+export const useTools = () => {
+  const { data } = useSuspenseQuery<Tool[]>({
+    queryKey: ["list-tools"],
```
fredericbarthelet, line 39: Is it really a good idea to use cache from react-query on list tools? What happens if I add a tool on my MCP server at that moment?

packages/devtools/src/lib/mcp/index.ts
```diff
@@ -0,0 +1,114 @@
...
+import { useMutation, useQuery, useSuspenseQuery } from "@tanstack/react-query";
+import type { CallToolArgs, OpenAiProperties } from "skybridge/web";
+import { useStore } from "@/lib/store.js";
+import { McpClient } from "./client.js";
+
+const client = new McpClient();
+
+client.connect(import.meta.env.VITE_MCP_SERVER_URL).then(() => {
+  console.info("Connected to MCP server");
+});
```
fredericbarthelet, line 12: I'm not sure this guarantee only a single MCPClient will be instantiated and used at a single time. If two separate files import a hook from this file, will it import and run the new Client code twice ? `@modelcontextprotocol/use-mcp` uses React ref to guarantee the instance stays the same across render -> https://github.com/modelcontextprotocol/use-mcp/blob/main/src/react/useMcp.ts

packages/devtools/src/lib/mcp/index.ts
```diff
@@ -0,0 +1,114 @@
...
+  });
+  return data;
+};
+
+export const useServerInfo = () => {
+  return client.getServerInfo();
+};
+
+export const useCallTool = () => {
+  const { setToolData } = useStore();
```
fredericbarthelet, line 50: why not simply rely on mutation state to keep track of tool call results since you're already using a mutation to do the actual tool call? -> https://tanstack.com/query/latest/docs/framework/react/reference/useMutationState

This would have the added benefit of not having to pass the data around to the setToolData op + you'll get infinite history (maybe no use for it at the moment though)

packages/devtools/src/lib/mcp/index.ts
fredericbarthelet: why not use the key factory pattern rather that reexporting custom hooks ?
https://tanstack.com/query/v4/docs/framework/react/community/lukemorales-query-key-factory

packages/devtools/src/lib/store.ts
```diff
@@ -0,0 +1,89 @@
...
+  input: Record<string, unknown>;
+  response: CallToolResponse;
+  openaiRef: React.RefObject<HTMLIFrameElement> | null;
+  openaiLogs: OpenAiLog[];
+  openaiObject: OpenAiProperties | null;
+};
+
+export type Store = {
+  selectedTool: string | null;
+  setSelectedTool: (tool: string | null) => void;
```
fredericbarthelet: wouldn't it be easier for selectedTool param to leverage an URL param instead to be consistent across re-render and easily accessible everywhere while still only exposing the emulator on `/` to keep the retour simple?

packages/devtools/src/components/layout/tool-panel/widget/create-openai-mock.ts
```diff
@@ -0,0 +1,117 @@
...
+    },
+  });
+}
+
+export function createAndInjectOpenAi(
+  iframeWindow: Window & { openai?: unknown },
+  initialValues: OpenAiProperties | null,
+  log: (command: string, args: UnknownObject) => void,
+  setValue: (key: keyof OpenAiProperties, value: unknown) => void,
+): void {
```
fredericbarthelet, line 107: use a single object arg, it's hard to read what is expected where otherwise

packages/devtools/src/lib/store.ts
```diff
@@ -0,0 +1,89 @@
...
+  tools: {
+    [name: string]: ToolData;
+  };
+  setToolData: (tool: string, data: Partial<ToolData>) => void;
+  pushOpenAiLog: (tool: string, log: Omit<OpenAiLog, "id">) => void;
+  updateOpenaiObject: (
```
fredericbarthelet, line 30: I would have pictured openai object as a completly different store, whose lifecycle is tied to a single widget only. It's a bit strange to have it in the global application store

packages/devtools/.env.example
harijoe: The idea was to allow someone to start the emulator as a standalone but we can remove it

packages/devtools/src/lib/store.ts
harijoe: No type here : This is how it's meant to be used. I think this is for inference purposes when middlewares are involved. (https://zustand.docs.pmnd.rs/apis/create#updating-state-based-on-previous-state)

packages/devtools/src/lib/store.ts
harijoe: Indeed, removed!

packages/devtools/src/lib/mcp/index.ts
harijoe: I think it it is :
- It prevents making a network request on every render if we were to use `client.listTools` directly
- By default react-query refreshes the queries when tab focus is changed so a new tool will quickly become visible

packages/devtools/src/lib/mcp/index.ts
harijoe: ES modules are singletons: the module code runs once, so all imports will share the same client. I think useMCP uses a ref to let the user clear the client object and instantiate another one (potentially for another server).

packages/devtools/src/lib/mcp/index.ts
harijoe: The reason is tool data is not only about the response of the tool call. For example it also stores the openai object ref, the logs etc. The idea here was to keep all data related to a tool in a single place for easier management (even though it meant not relying on using the mutation state). Relying on mutation state felt more complicated with not much added benefits compared to keeping everything organised under a single tool name key in the global store.

packages/devtools/src/lib/mcp/index.ts
harijoe: Using query keys factory feels completely overkill for what I try to achieve : basically I need only two queries (with keys ['list-tools'] and ['resource', uri]. Those are just simple queries that don't require any specific cache management strategies. Also I don't use these queries elsewhere in the codebase so the type-safety offered by the query factory isn't useful either.

packages/devtools/src/lib/store.ts
harijoe: Easier I am not sure (as having a global zustand store as the source of truth is a pretty comfortable pattern I think) but it would allow to be consistent across page refresh : I'll implement it 👍

packages/devtools/src/lib/store.ts
harijoe: Not sure what you mean : each openai object *is* tied to a single widget only. This *updateOpenaiObject* method is just a convenient way to update a key of the openai object specific to a given tool. Maybe I don't get it correctly, let's discuss it!

packages/devtools/.env.example
fredericbarthelet: standalone use will only be for package maintainer. Otherwise, you'd better use MCPJam for a battery-included emulator.

fredericbarthelet:
Closes #51

Issue #84, opened 2025-12-11
Replace occurences of `any` types in


fredericbarthelet:
This should remove remaining biome warnings regarding `lint/suspicious/noExplicitAny`

Issue #83, opened 2025-12-11
Remove occurences of non-null assertion


fredericbarthelet:
This should remove all warnings in biome of `lint/style/noNonNullAssertion`

PR #90, opened 2025-12-12
chore: fix biome and lint warnings

Rodriguespn:
## Changes Made

This PR addresses multiple linting and code quality improvements:

### 1. Non-null Assertion Fixes
- **Files**: Multiple hook files in `src/web/hooks/`
- Replaced all non-null assertions (`!`) with explicit undefined checks and descriptive error messages

### 2. TypeScript Type Safety
- **Files**: `src/test/widget.test.ts`, `src/web/data-llm.test.tsx`
- Replaced `any` types with proper TypeScript types

### 3. CI/CD Improvements
- **Files**: `.github/workflows/ci.yml`, `package.json`, `.nvmrc`
- Created `.nvmrc` file using `lts/krypton` to avoid locking to a specific minor/patch version
- Separated code quality checks from test job for better CI feedback:
  - **`code-quality` job**: Runs `biome ci .` (formatting and linting) and TypeScript type checking independently
  - **`test` job**: Builds and runs unit tests independently
- Updated CI to use `biome ci` instead of `biome check` following [Biome's CI documentation](https://biomejs.dev/recipes/continuous-integration/) for better GitHub Actions integration (annotations)
- Simplified `test` script to only run unit tests; code quality checks are handled by the separate CI job

### 4. Biome Extension Configuration
- **File**: `.vscode/settings.json` (if applicable)
- Configured Biome extension to work properly in Cursor IDE

## Rationale

The separation of code quality and test jobs allows:
- **Non-blocking tests**: Tests can run even if TypeScript has errors, providing faster feedback during development when you might have incomplete type definitions but still want to verify runtime behavior
- **Better CI integration**: Using `biome ci` provides GitHub Actions annotations for better visibility of code quality issues
- **Clear separation of concerns**: Code quality checks (formatting, linting, type checking) are TypeScript-specific and run independently from functional tests

Closes: https://github.com/alpic-ai/skybridge/issues/83
Closes: https://github.com/alpic-ai/skybridge/issues/84

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "test": "pnpm run test:unit && pnpm run test:type && pnpm run test:format",
     "test:unit": "vitest run",
     "test:type": "tsc --noEmit",
-    "test:format": "biome ci --error-on-warnings",
+    "test:format": "biome ci",
     "docs:dev": "pnpm --filter @skybridge/docs start",
     "docs:build": "pnpm --filter @skybridge/docs build",
     "docs:serve": "pnpm --filter @skybridge/docs serve"


.nvmrc
```diff
@@ -0,0 +1,2 @@
+24.12.0
```
fredericbarthelet: Use `lts/krypton` here instead to avoid locking to a minor/patch

src/web/hooks/use-display-mode.ts
```diff
@@ -2,7 +2,12 @@ import type { DisplayMode } from "../types.js";
 import { useOpenAiGlobal } from "./use-openai-global.js";
 
 export function useDisplayMode() {
-  const displayMode = useOpenAiGlobal("displayMode")!;
+  const displayMode = useOpenAiGlobal("displayMode");
```
fredericbarthelet, line 5: You'd avoid a lot of duplicated code by making this check directly in `useOpenAiGlobal` hook directly
https://github.com/alpic-ai/skybridge/blob/ee49380aad402306c7c42925f9d93650f25b76af/src/web/hooks/use-openai-global.ts#L36

But this was taken from OpenAI implementation repo. Are we sure there is no lifecycle phase where `window.openai` is not yet defined/available? Or this was just in place to make the execution fails silently in case the widget was rendered in a different environment ?

.github/workflows/ci.yml
```diff
@@ -6,6 +6,31 @@ on:
     branches: [main]
 
 jobs:
+  code-quality:
```
fredericbarthelet: test is already testing type, format and unit. Why did you add this job ?

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-  "editor.defaultFormatter": "biomejs.biome"
+  "editor.defaultFormatter": "biomejs.biome",
+  "biome.lsp.bin": "${workspaceFolder}/node_modules/.bin/biome"
```
fredericbarthelet, line 3:
```suggestion
  "biome.lspBin": "${workspaceFolder}/node_modules/.bin/biome"
```
Should be lspBin param according to biome doc: [biome.lspBin](https://biomejs.dev/fr/reference/vscode/#biomelspbin)

src/web/hooks/use-call-tool.test.ts
```diff
@@ -202,14 +202,17 @@ describe("useCallTool - TypeScript typing", () => {
...
       promise = result.current.callToolAsync(testArgs);
       expectTypeOf<Promise<typeof mockResponse>>(promise);
       resolvedValue = await promise;
     });
 
-    expect(resolvedValue!).toEqual(mockResponse);
+    if (resolvedValue === undefined) {
+      throw new Error("Expected resolvedValue to be defined");
+    }
+    expect(resolvedValue).toEqual(mockResponse);
```
fredericbarthelet: You can do simpler, act accepts a return value that will be forwarded back outside of the callback scope for this purpose :)

```suggestion
    const returnedValue = await act(async () => {
      return result.current.callToolAsync();
    });

    expectTypeOf<typeof mockResponse>(returnedValue);
    expect(returnedValue).toEqual(mockResponse);
```

src/web/hooks/use-call-tool.test.ts
```diff
@@ -235,13 +238,16 @@ describe("useCallTool - TypeScript typing", () => {
     OpenaiMock.callTool.mockResolvedValueOnce(mockResponse);
 
     let promise: Promise<typeof mockResponse>;
-    let resolvedValue: typeof mockResponse;
+    let resolvedValue: typeof mockResponse | undefined;
```
fredericbarthelet: same comment here

package.json
```diff
@@ -20,6 +20,7 @@
     "build": "tsc && pnpm run build:templates",
     "build:templates": "cp -r src/server/templates dist/src/server/",
     "format": "biome check --write",
+    "lint": "biome lint --error-on-warnings",
```
fredericbarthelet: biome lint is a subset of [biome check](https://biomejs.dev/reference/cli/#biome-check), why do you need this additional command in particular?

Rodriguespn: Answered [here](https://github.com/alpic-ai/skybridge/pull/90#issuecomment-3650739010)

.github/workflows/ci.yml
Rodriguespn: Let's continue this discussion [here](https://github.com/alpic-ai/skybridge/pull/90#issuecomment-3650739010)

.nvmrc
Rodriguespn: Done ✅

.vscode/settings.json
Rodriguespn: It's weird because in the [english version](https://biomejs.dev/reference/vscode/#biomelspbin) of Biome docs, the correct param is `biome.lsp.bin`, but in the French one is `biome.lspBin` 🤣

fredericbarthelet: Hahaha 🤣 That's a first. `biome.lsp.bin` is the correct one. I guess Biome's french documentation needs updating. Thanks for finding that out @Rodriguespn !

package.json
fredericbarthelet: It was removed, this is now resolved :)

package.json
```diff
@@ -20,10 +20,9 @@
     "build": "tsc && pnpm run build:templates",
     "build:templates": "cp -r src/server/templates dist/src/server/",
     "format": "biome check --write",
-    "test": "pnpm run test:unit && pnpm run test:type && pnpm run test:format",
```
fredericbarthelet, line 23: please keep this in order to be able to run locally and iterate on a specific dev

Rodriguespn:
> > Regarding biome lint script: Removed it frompackage.json since biome check already includes linting.
> 
> Perfect 👌
> 
> > Regarding separate code-quality job: I separated it for two reasons: non-blocking test and biome ci
> 
> My believes on that matter is that CI is an automated shared quality check to ensure no merging occurs on a broken PR. Feedback loop on types correctness and unit tests pass should be done locally on your IDE while developing the feature, not once you push to Github :) However I'm open to give better visibility with separated jobs that will all always run in parallel. But I'd like to keep it DRY (so no double installs) and as simple as possible. Hereafter is a suggested implementation where install is still made once, and 3 jobs are separated using a matrix. WDYT ?
> 
> switching from `biome check` to `biome ci` is a good improvement, but unrelated to jobs running in series or in parallel. You can however update the command in the package.json
> 
> ```yaml
> name: CI
> 
> on:
>   pull_request:
>   push:
>     branches: [main]
> 
> jobs:
>   install:
>     name: Install
>     runs-on: ubuntu-latest
>     steps:
>       - uses: actions/checkout@v4
> 
>       - uses: pnpm/action-setup@v4
> 
>       - uses: actions/setup-node@v4
>         with:
>           node-version: .nvmrc
>           cache: 'pnpm'
> 
>       - run: pnpm install --frozen-lockfile
> 
>       - uses: actions/cache/save@v4
>         with:
>           path: |
>             node_modules
>             dist
>           key: build-${{ github.sha }}
> 
>   test:
>     name: ${{ matrix.name }}
>     needs: install
>     runs-on: ubuntu-latest
>     strategy:
>       fail-fast: false
>       matrix:
>         include:
>           - name: Unit Tests
>             command: test:unit
>           - name: Type Check
>             command: test:type
>           - name: Lint & Format
>             command: test:format
>     steps:
>       - uses: actions/checkout@v4
> 
>       - uses: pnpm/action-setup@v4
> 
>       - uses: actions/setup-node@v4
>         with:
>           node-version: .nvmrc
>           cache: 'pnpm'
> 
>       - uses: actions/cache/restore@v4
>         with:
>           path: |
>             node_modules
>             dist
>           key: build-${{ github.sha }}
> 
>       - run: pnpm run ${{ matrix.command }}
>       ```
> ```

Agreed on CI being for shared quality checks, not developer feedback loops. The separate jobs are mainly for better visibility during PR reviews - when a check fails, it's immediately clear whether it's tests, types, or formatting without having depencies between them. I understand that the test suit is not big at the moment to run locally but as it grows, it will come handy to run it in the CI without being blocked by lint and formatting. Happy to go with my suggestion or your proposal nevertheless.

I've implemented your matrix approach with one change: the build runs in the install job so the `dist` folder is cached and available to all test jobs. The sequential `test` script is kept in package.json for local development.

Here's the updated `.github/workflows/ci.yml`:

```yaml
name: CI

on:
  pull_request:
  push:
    branches: [main]

jobs:
  install:
    name: Install
    runs-on: ubuntu-latest
    permissions:
      contents: read
    steps:
      - uses: actions/checkout@v4

      - uses: pnpm/action-setup@v4

      - uses: actions/setup-node@v4
        with:
          node-version-file: ".nvmrc"
          cache: "pnpm"

      - name: Install dependencies
        run: pnpm install --frozen-lockfile

      - name: Build project
        run: pnpm run build

      - uses: actions/cache/save@v4
        with:
          path: |
            node_modules
            dist
          key: build-${{ github.sha }}

  test:
    name: ${{ matrix.name }}
    needs: install
    runs-on: ubuntu-latest
    permissions:
      contents: read
    strategy:
      fail-fast: false
      matrix:
        include:
          - name: Unit Tests
            command: test:unit
          - name: Type Check
            command: test:type
          - name: Lint & Format
            command: test:format # Runs "biome ci ."
    steps:
      - uses: actions/checkout@v4

      - uses: pnpm/action-setup@v4

      - uses: actions/setup-node@v4
        with:
          node-version-file: ".nvmrc"
          cache: "pnpm"

      - uses: actions/cache/restore@v4
        with:
          path: |
            node_modules
            dist
          key: build-${{ github.sha }}

      - name: Build project
        if: matrix.name == 'Unit Tests'
        run: pnpm run build

      - name: Run ${{ matrix.name }}
        run: pnpm run ${{ matrix.command }}
```

WDYT about this implementation?

src/web/hooks/use-display-mode.ts
Rodriguespn: You're right about avoiding duplicated code. The check is already at the top of `useOpenAiProperty` (line 38), so the optional chaining in the `getSnapshot` function (line 63) is redundant. We should remove the optional chaining and rely on the explicit check at the top. 
If we implement the proposal in #94 , we can apply this change in that PR. If we end up not going through with #94, happy to change it in this PR

.github/workflows/ci.yml
```diff
@@ -6,25 +6,65 @@ on:
...
+
+      - uses: actions/cache/restore@v4
+        with:
+          path: |
+            node_modules
+          key: build-${{ github.sha }}
+
       - name: Build project
+        if: matrix.name == 'Unit Tests'
         run: pnpm run build
```
fredericbarthelet: why do you need build to run unit tests? Vitest handles running tests against the source files, not the built ones. I tested locally on my maching (deleting `dist` to ensure all tests still pass).

src/web/hooks/use-call-tool.test.ts
```diff
@@ -201,15 +201,12 @@ describe("useCallTool - TypeScript typing", () => {
...
-      expectTypeOf<Promise<typeof mockResponse>>(promise);
-      resolvedValue = await promise;
+    const promise = act(async () => {
+      return result.current.callToolAsync(testArgs);
     });
-
-    expect(resolvedValue!).toEqual(mockResponse);
+    expectTypeOf<Promise<typeof mockResponse>>(promise);
+    const resolvedValue = await promise;
+    expect(resolvedValue).toEqual(mockResponse);
```
fredericbarthelet: Why did you did a separate implementation that the one in `should correctly type callToolAsync when ToolArgs is null` ?
```suggestion
   const returnedValue = await act(async () => {
      return result.current.callToolAsync();
    });
    expectTypeOf<typeof mockResponse>(returnedValue);
    expect(returnedValue).toEqual(mockResponse);
```

package.json
```diff
@@ -19,11 +19,11 @@
   "scripts": {
     "build": "tsc && pnpm run build:templates",
     "build:templates": "cp -r src/server/templates dist/src/server/",
-    "format": "biome check --write",
+    "format": "biome check --write --error-on-warnings",
     "test": "pnpm run test:unit && pnpm run test:type && pnpm run test:format",
     "test:unit": "vitest run",
     "test:type": "tsc --noEmit",
-    "test:format": "biome check",
+    "test:format": "biome ci .",
```
fredericbarthelet: I think file path is not needed (`.` by default)

```suggestion
    "test:format": "biome ci",
```

src/web/hooks/use-display-mode.ts
fredericbarthelet: I don't understand what you are referring to - I don't have any occurence of a method `useOpenAiProperty` or `getSnapshot` in the current code base. Do you mean `src/web/hooks/use-openai-global.ts` ?
The check done in this file for the presence of `window.openai` only displays a warning. It does not throw. My recommendation was to change from warning to error, and to implement in external store getter (rather than as a hook early check to allow for external store creation). If you want to give it more thoughts, happy to treat this in a separate PR in order for this one to move forward :)

fredericbarthelet (changes requested):
Thanks for udpating your PR @Rodriguespn  :)

A few final comments and we should be good to go.

src/web/hooks/use-display-mode.ts
Rodriguespn: Apologies for the confusion. I misunderstood your suggestion at first.
You're right, the current check only logs a warning, and the optional chaining in the getSnapshot callback allows silent failures. I'll apply your recommendation and move the error to the external store getter.

We can continue discussing the unified useOpenAiGlobal hook for properties in #94.

.github/workflows/ci.yml
Rodriguespn: True, let me change that. Thank you for noticing this.

On another note, shouldn't we also check if the project builds as a CI job or part of one?

package.json
Rodriguespn: Changed ✅ 
One thing that I've noticed, is that `biome ci --error-on-warnings` ignores the flag and doesn't throw an error on warnings, unlike `biome check --error-on-warnings`. Do we want to keep this difference between local dev scripts and the CI job?

src/web/hooks/use-display-mode.ts
Rodriguespn: Sorry I misunderstood your question and was working over old code on my branch (hence the mention of the `useOpenAiProperty` or `getSnapshot` ghost methods).

It's true that the example from the OpenAI repo only throws a warning, but that raises a non-nullable warning when the hook is called (the `style/noNonNullAssertion` lint rule we're enforcing in this PR).

After thinking about this, I came up with a solution using function overloads: `useOpenAiGlobal` now defaults to `required: true` (throws if `undefined`), but accepts `{ required: false }` for hooks that need to handle missing values gracefully (like `useWidgetState` or `toolOutput`/`toolResponseMetadata` in pending state).

This centralizes the error handling, removes the duplicated `undefined` checks from consumers, and satisfies the lint rule without non-null assertions.

fredericbarthelet: Thanks for updating with this strategy @Rodriguespn :)
I don't quite get the necessity for this new required options and overload.
`toolOuput` and `toolResponseMetadata` already have a typing representing a loading state and a loaded state.
I took the liberty to push a new commit to show what I had in mind. Let me know what you think and feel free to merge if you like my proposal, or push another commit if you have a better idea :)

Rodriguespn: Got it now, thanks for your patience and for committing the change 👍. I'm happy to merge this with your implementation.

Before you do it, could you please address these two questions I've raised please? 

1. https://github.com/alpic-ai/skybridge/pull/90#discussion_r2621145983
2. https://github.com/alpic-ai/skybridge/pull/90#discussion_r2621168649
 
They are not blockers so feel free to merge this PR and if you find them relevant, we can continue on another PR. Thank you!

package.json
fredericbarthelet: For sure let's delete the flag where it doesn't have any effect.

.github/workflows/ci.yml
fredericbarthelet: Since tsc is the builder and test:type tests `tsc`, we should be good to go for the time being

PR #92, opened 2025-12-13
Make useRequestModal accurate to window.openai

khandrew1:
* Adds `params` and `anchor` to the useRequestModal hook.
* `useRequestModal` now returns `[open, isOpen]`, to open the modal and to check if the modal is open.
* After testing with ChatGPT, all options are considered optional.
* Originally I had said that you had to use `view` to determine if the widget is a modal. This isn't true as ChatGPT also updates the `displayMode` to modal as well, so its fine to just track that I feel.
* Also updated test case to better cover the new Modal.

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR updates `useRequestModal` to match `window.openai.requestModal` behavior by making all parameters optional and adding `params` and `anchor` options. The hook now returns `[open, isOpen]` instead of just a function, where `isOpen` tracks modal state via `displayMode === "modal"`.

**Major changes:**
- Breaking API change: return type changed from function to `[open, isOpen]` tuple
- All `requestModal` parameters now optional (`title`, `params`, `anchor`)
- Added `"modal"` to `DisplayMode` type definition
- Tests updated to cover new functionality

**Issues:**
- Documentation in `docs/docs/api-reference/use-request-modal.md` not updated for the breaking API change - all examples still use the old API

- This PR introduces a breaking API change with outdated documentation that will break existing usage patterns
- The code implementation and tests are solid, but the documentation was not updated to reflect the breaking API changes. All documentation examples will fail with the new API, which would confuse users and break existing code that follows the docs
- The documentation file `docs/docs/api-reference/use-request-modal.md` requires updates to match the new API

### DIFF
--- a/packages/core/src/web/hooks/use-request-modal.test.ts
+++ b/packages/core/src/web/hooks/use-request-modal.test.ts
@@ -7,6 +7,7 @@ describe("useRequestModal", () => {
 
   beforeEach(() => {
     requestModalMock = vi.fn();
+    vi.stubGlobal("skybridge", { hostType: "apps-sdk" });
     vi.stubGlobal("openai", {
       requestModal: requestModalMock,
       view: { mode: "inline" },
--- a/packages/core/src/web/hooks/use-request-modal.ts
+++ b/packages/core/src/web/hooks/use-request-modal.ts
@@ -1,4 +1,4 @@
-import { useAppsSdkBridge } from "../bridges/use-apps-sdk-bridge.js";
+import { useAppsSdkBridge } from "../bridges/hooks/use-apps-sdk-bridge.js";
 import type { RequestModalOptions } from "../types.js";
 
 /**


src/web/hooks/use-request-modal.ts
```diff
@@ -1,8 +1,19 @@
+import { useOpenAiGlobal } from "./use-openai-global.js";
+
 /**
  * Triggers a modal containing the widget rendered in display mode "modal"
  */
 export function useRequestModal() {
-  return (options: { title: string }) => {
+  const displayMode = useOpenAiGlobal("displayMode");
+  const isOpen = displayMode === "modal";
```
fredericbarthelet: Exemple from OpenAI uses the `view` param on `window.openai`. Why do you use the `displayMode` param here?
https://github.com/openai/openai-apps-sdk-examples/blob/31f00df908a1a98659239a0ed44e87a44327e140/src/pizzaz-shop/index.tsx#L599-L601

src/web/hooks/use-request-modal.ts
```diff
@@ -1,8 +1,19 @@
...
+
+  const open = (options: {
+    title?: string;
+    params?: Record<string, unknown>;
+    anchor?: { top?: number; left?: number; width?: number; height?: number };
```
fredericbarthelet: To avoid duplication, export type from `types.ts` like it was done for `CallToolArgs` or `CallToolResponse` and use it here :)

src/web/hooks/use-request-modal.ts
```diff
@@ -1,8 +1,19 @@
...
+  }) => {
     window.openai.requestModal(options);
   };
+
+  return [open, isOpen] as const;
```
fredericbarthelet: I'm not a big fan of tupple return interfaces similar to `useState`. It was initially implemented this way in React to avoid having to rename return params all the time (as tupple destructuring is only using position to match scope variable to returned variables)

```ts
const [myCustomPropName, myCustomSetterName] = useState();
```

requires less characters than

```ts
const { value: myCustomPropName, setValue: myCustomSetterName } = useState();
```

But in our case, renaming is highly improbable, and I'd stick with more modern approaches of object being return (with the additional benefit of being able to add new props in the future on the return object without it being a breaking change - like `close` for example).


```suggestion
  return { open, isOpen }
```

Similar implementations:
- [useHooks](https://usehooks-ts.com/react-hook/use-boolean)
- [chakra](https://v2.chakra-ui.com/docs/hooks/use-disclosure)

fredericbarthelet:
Thanks for the detailed explanation @khandrew1 

I'm happy to track displayMode value in that case, view isn't documented at all and while it misses the "modal" mode, displayMode is at least mentioned in the apps SDK docs.

Are params available in `view.params` the ones passed to the modal via the requestModal `params` argument?

If a complete new widget is rendered within an iframe to display the modal content, I'd love to keep updating this hook API and provide a Component definition as one of the hook args to be rendered as widget content. This could have the added benefit of controlling component props with a generic type on the hook. I'm fine with your current implementation. But I'd welcome a v2 if you want to keep digging in that direction :) 

Just before merging this, could you also update the hook documentation? Thanks 🙏

khandrew1:
@fredericbarthelet Yes, the params available in `view.params` is the one that is passed via the `params` argument. In the image above you can see that the pizza app passes in `checkout` and `selectedCartItemId` to correctly render the modal with the chicken pizza modal.

The Component definition sounds like a great idea! I'd love to look into that in a separate PR.

Will update the docs soon!

khandrew1:
@fredericbarthelet Sorry for the delay on this! Updated this PR with the new Apps SDK bridge as well as documentation updates.

fredericbarthelet (approved):
LGTM! Thanks @khandrew1 
We've just moved src to a dedicated packages/core subfolder. I'll update the changes and merge your PR :)

Next step on this hook could also include the use of a generic type for `params` in order to be able to defined at hook which params should be passed, and enforce type in `params` returned from hook. WDYT :) ?

PR #109, opened 2025-12-18
feat: interface return type is correctly inferred by generateHelpers

### LINKED ISSUE #108: Type inference of widgets does not work with Interface (only Type)
When typing return type of server registerTool / registerWidget outputs with Interface, `generateHelpers` type inference won't work

valentinbeggi:
should fix #108

<h3>Greptile Summary</h3>

This PR fixes TypeScript type inference for `generateHelpers` when using interface declarations for return types. Previously, when developers used interface types (rather than inline object types) for widget/tool return values, the type extraction utilities couldn't properly infer the `structuredContent` and `_meta` properties.

Key changes:
- Wrapped `ExtractStructuredContent` and `ExtractMeta` results with `Simplify<T>` utility to expand interface references into their actual shape
- Improved `ExtractMeta` to properly handle cases where `_meta` doesn't exist using tuple-based `never` check
- Loosened `TReturn` constraint from `CallToolResult` to `{ content: CallToolResult["content"] }` to allow interfaces with additional properties
- Added comprehensive type tests demonstrating the fix works for interface-typed return values

- The changes are purely type-level improvements with comprehensive type tests demonstrating correctness. The modifications properly handle edge cases (like `never` types) and maintain backward compatibility while extending functionality to support interface declarations. No runtime behavior changes.

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -80,12 +80,17 @@ type McpServerOriginalToolConfig = Omit<
   "inputSchema" | "outputSchema"
 >;
 
+type Simplify<T> = { [K in keyof T]: T[K] };
+
 type ExtractStructuredContent<T> = T extends { structuredContent: infer SC }
-  ? SC
+  ? Simplify<SC>
   : never;
 
-type ExtractMeta<T> =
-  Extract<T, { _meta: unknown }> extends { _meta: infer M } ? M : unknown;
+type ExtractMeta<T> = [Extract<T, { _meta: unknown }>] extends [never]
+  ? unknown
+  : Extract<T, { _meta: unknown }> extends { _meta: infer M }
+    ? Simplify<M>
+    : unknown;
 
 /**
  * Type-level marker interface for cross-package type inference.
@@ -99,8 +104,6 @@ type ExtractMeta<T> =
 export interface McpServerTypes<TTools extends Record<string, ToolDef>> {
   readonly tools: TTools;
 }
-
-type Simplify<T> = { [K in keyof T]: T[K] };
 type ShapeOutput<Shape extends ZodRawShapeCompat> = Simplify<
   {
     [K in keyof Shape as undefined extends SchemaOutput<Shape[K]>
@@ -134,7 +137,7 @@ type ToolConfig<TInput extends ZodRawShapeCompat | AnySchema> = {
 
 type ToolHandler<
   TInput extends ZodRawShapeCompat,
-  TReturn extends CallToolResult = CallToolResult,
+  TReturn extends { content: CallToolResult["content"] } = CallToolResult,
 > = (
   args: ShapeOutput<TInput>,
   extra: RequestHandlerExtra<ServerRequest, ServerNotification>,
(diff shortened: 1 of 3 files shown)


src/server/server.ts
```diff
@@ -80,12 +80,17 @@ type McpServerOriginalToolConfig = Omit<
...
+type Simplify<T> = { [K in keyof T]: T[K] };
+
 type ExtractStructuredContent<T> = T extends { structuredContent: infer SC }
-  ? SC
+  ? Simplify<SC>
   : never;
 
-type ExtractMeta<T> =
-  Extract<T, { _meta: unknown }> extends { _meta: infer M } ? M : unknown;
+type ExtractMeta<T> = [Extract<T, { _meta: unknown }>] extends [never]
```
fredericbarthelet, line 89: Why do you need an array for your assertion?

src/server/server.ts
```diff
@@ -134,7 +137,7 @@ type ToolConfig<TInput extends ZodRawShapeCompat | AnySchema> = {
 
 type ToolHandler<
   TInput extends ZodRawShapeCompat,
-  TReturn extends CallToolResult = CallToolResult,
+  TReturn extends { content: CallToolResult["content"] } = CallToolResult,
```
fredericbarthelet, line 140: I also don't get why this has a positive impact 😅

src/server/server.ts
valentinbeggi: This is a typescript trickzzz
```ts                                                                                                                
  // Without array: distributes over empty union → returns never, not boolean                                   
  type Test1 = never extends never ? true : false; // never (!)                                                 
                                                                                                                
  // With array: prevents distribution → works correctly                                                        
  type Test2 = [never] extends [never] ? true : false; // true ✓  
```

src/server/server.ts
valentinbeggi: It isn't required for Interface inference to work correctly

But it is required for good devX, it allows user to not define explicitely [x: string]: unknown in their interfaces (which is cumbersome)


I had a difficult time to fix this devX issue and understand exactly the problem but Claude wrote me a good explanation:

```ts
// Step 1: SDK defines a type with an index signature
// The `[x: string]: unknown` means "any string key is allowed"
type SdkMeta = {
  [x: string]: unknown; // <-- index signature
  taskId?: string;
};

type SdkResult = {
  content: string[];
  _meta?: SdkMeta;
};

// Step 2: User defines their own interface (no index signature)
interface UserMeta {
  requestId: string;
  cached: boolean;
}

interface UserResult {
  content: string[];
  _meta: UserMeta;
}

// Step 3: Try to use UserResult where SdkResult is expected
function processResult<T extends SdkResult>(result: T) {
  return result;
}

// ❌ ERROR: Type 'UserResult' does not satisfy constraint 'SdkResult'
//    Index signature for type 'string' is missing in type 'UserMeta'
//
const result1: UserResult = { content: ["hi"], _meta: { requestId: "123", cached: true } };
processResult<UserResult>(result1);

// WHY? Interfaces are "closed" - they only allow declared keys
// Index signatures are "open" - they allow any string key
// UserMeta says "only requestId and cached exist"
// SdkMeta says "any key can exist"
// These are incompatible!

// Step 4: THE FIX - only constrain on what we actually need
function processResultFixed<T extends { content: string[] }>(result: T) {
  return result;
}

// ✅ Works! We only require `content`, so `_meta` can be whatever
const result: UserResult = {
  content: ["hi"],
  _meta: { requestId: "123", cached: true },
};
processResultFixed<UserResult>(result);

// The actual _meta type is still preserved through inference
// We just don't force it to match the SDK's indexed type
```

PR #110, opened 2025-12-18
Refacto packages organization

harijoe:
## Skybridge Monorepo Refactor

This PR **converts Skybridge from a single-package repository into a proper `pnpm` monorepo**, with a clean separation between:

- **`packages/core`** → the published `skybridge` framework (server + web APIs)
- **`packages/devtools`** → the React-based DevTools UI
- **Root workspace** → orchestration, CI, versioning, and publishing

It also **updates CI, release, and publishing workflows** to align with this structure.

<h3>Greptile Summary</h3>

Refactored repository from single-package to monorepo structure with separate `packages/core` and `packages/devtools` packages.

**Key Changes:**
- Moved all core source code from `src/` to `packages/core/src/` while preserving the package name `skybridge`
- Converted `@skybridge/devtools` from private to publishable package
- Updated `devtoolsStaticServer.ts` to resolve devtools package path using `require.resolve()` instead of relative paths, which correctly handles the new package structure
- Added automation scripts (`version.sh`, `publish-skybridge.sh`, `publish-packages.sh`) to manage versioning and publishing across packages
- Updated CI/CD workflows to build and publish both packages with separate npm tokens (`NPM_TOKEN` for skybridge, `ORG_NPM_TOKEN` for scoped packages)
- Simplified CI workflow from matrix strategy to single test run using workspace-level `pnpm test`
- Removed unused `useQuery` import from devtools mcp module

**Notes:**
- Root `package.json` contains duplicate dependencies that should be removed (already defined in `packages/core/package.json`)
- Build order is correctly handled by pnpm workspace dependency resolution (core builds before devtools)
- Version synchronization ensures both packages are published with matching versions

- This PR is safe to merge with minimal risk after addressing the duplicate dependencies issue
- The refactoring is well-structured with proper monorepo patterns (workspace dependencies, build scripts, CI updates). The only issue is duplicate dependencies in root package.json which won't cause runtime problems but should be cleaned up. All critical changes (devtools path resolution, build order, publishing workflow) are correctly implemented.
- Pay attention to `package.json` (root) - remove duplicate dependencies listed in the style comment


.github/workflows/ci.yml
```diff
@@ -28,24 +28,16 @@ jobs:
...
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - name: Unit Tests
-            command: pnpm run test:unit
-          - name: Type Check
-            command: pnpm run test:type
-          - name: Lint & Format
-            command: pnpm run test:format
```
fredericbarthelet, line 48: The split in install and test workflows was there to avoid reinstalling for each line of the matrix of test. With a single red process, you can remove the cache step and put everything in one workflow

package.json
```diff
@@ -1,32 +1,18 @@
 {
-  "name": "skybridge",
+  "name": "skybridge/monorepo",
```
fredericbarthelet, line 2: Can't keep the name?

packages/devtools/package.json
```diff
@@ -32,7 +36,7 @@
     "react-resizable-panels": "^3.0.6",
     "shadcn": "^3.6.1",
     "shiki": "^3.19.0",
-    "skybridge": "^0.12.0",
+    "skybridge": "workspace:*",
```
fredericbarthelet, line 39: This should be a peerdep and a dev dep now I think

packages/core/src/server/devtoolsStaticServer.ts
```diff
@@ -19,8 +21,17 @@ import express, { type RequestHandler } from "express";
  */
 export const devtoolsStaticServer = async (): Promise<RequestHandler> => {
   const router = express.Router();
-  const currentDir = path.dirname(fileURLToPath(import.meta.url));
-  const devtoolsPath = path.join(currentDir, "..", "devtools");
+  let devtoolsPath: string;
+  const devtoolsPackagePath = require.resolve(
+    "@skybridge/devtools/package.json",
+  );
```
fredericbarthelet, line 27: This looks strange.
Maybe bridging core and dev tools should be done in template rather than in one project or the other

scripts/publish-packages.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -e
+
+TAG=${1:-dev}
+
+pnpm --filter @skybridge/devtools publish --tag "$TAG" --access public --ignore-scripts --no-git-checks
```
fredericbarthelet, line 6: You don't publish core ?

scripts/version.sh
```diff
@@ -0,0 +1,17 @@
...
+if [ -n "$1" ]; then
+  VERSION="$1"
+else
+  SHORT_SHA=$(git rev-parse --short HEAD)
+  VERSION="0.0.0-dev.${SHORT_SHA}"
+fi
+
+echo "Bumping version to: ${VERSION}"
+
+pnpm --filter skybridge exec npm version "${VERSION}" --no-git-tag-version
```
fredericbarthelet, line 14: I think you can remove "old" way of publishing altogether in this PR

fredericbarthelet (commented):
A few comments to address, otherwise very good work. This starts to take shape :)!

package.json
harijoe: It's just a name for the root package.json. It feels weird to have two package.json in the same monorepo with the same name. Maybe there's a better way to handle that.
vite named the root package.json : @vitejs/vite-monorepo

packages/core/src/server/devtoolsStaticServer.ts
harijoe: I'm not a big fan of adding boilerplate to the template just because we have trouble with our implementation.
This require.resolve looks weird I agree, but it works well! 😅

scripts/publish-packages.sh
harijoe: Core is published in a dedicated script because it requires a specific npm_token (the one of your account) compared to the other packages which require the one I created for the skybridge org

scripts/version.sh
harijoe: This is not the old way of publishing! This line is a way to push the new version to every package.json of the monorepo : we need it to make it work.

packages/devtools/package.json
harijoe: Let's discuss it tomorrow : dev dep I agree but not sure about peer dep

package.json
fredericbarthelet: I was more surprised by the missing @ at the beginning of the package name.
I'm good with the version with a @ prefix like vite :) 

`@skybridge/monorepo`

scripts/version.sh
fredericbarthelet: It should be the new package name then, no? Like @skybridge/monorepo

packages/devtools/package.json
fredericbarthelet: My intent was to make devtools optional in fine, but let's go with that for the moment

fredericbarthelet:
Closing in favor of #131

Issue #97, opened 2025-12-15
Polyfill useTheme hook


fredericbarthelet:
Implements a compatibility layer at hook level between [Apps SDK](https://developers.openai.com/apps-sdk/build/chatgpt-ui#understand-the-windowopenai-api) and [MCP Apps](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#transport-layer). The objective of this issue is to commit to a single implementation of `src/web/hooks/use-theme.ts` hook that both works in an OpenAI Apps SDK hosted iFrame as well as in an MCP App hosted iFrame.

This is the first polyfill for MCP Apps API in Skybridge 🎉 

Signature on respective standards for both property is exactly the same (which make the first implementation of a common API easier)

### Remaining points to discuss

- **Single Resource content for all runtimes vs one resource content per runtime** - Should we provide a single implementation that is working in both runtimes (the first one returning something wins, with a parallel or sequential test) or should we provide 2 implementations that are packaged in each runtime content version?

### In Apps SDK
```ts
// visuals
theme: Theme;

//...

type Theme = "light" | "dark";
```
_From https://github.com/openai/openai-apps-sdk-examples/blob/31f00df908a1a98659239a0ed44e87a44327e140/src/types.ts#L41_

### In MCP Apps

```ts
  /** Current color theme preference */
  theme?: "light" | "dark";
````
_From https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#host-context-in-mcpuiinitializeresult_

### Technical implementation recommendations :

- **No dependency to @modelcontextprotocol/ext-apps** - unfortunately they package their own version of @modelcontextprotocol/sdk and are not compatible with Skybridge BYO SDK strategy. Use the no sdk approach detailed in https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#transport-layer in order to implement the base abstraction (similar to external store in `useOpenAiGlobals`).
- Documentation with mappings to MCP App implementation should be added to `docs/docs/skybridge-core-concepts.md`

PR #119, opened 2025-12-19
feat: add use mcp app bridge connector

qchuchu:
- Fixes partially #97

Goal of this PR : 
- Add a class that allows a connexion from a MCP Apps parent host (tested by using basic host example from https://github.com/modelcontextprotocol/ext-apps)
- Test that if the host shares the host context and being able to access it through the bridge (theme for the moment)

<h3>Greptile Summary</h3>

This PR adds the first MCP App bridge connector to Skybridge, enabling compatibility with the MCP Apps standard alongside the existing OpenAI Apps SDK support.

**Key changes:**
- Created `McpAppBridge` class implementing JSON-RPC 2.0 protocol for MCP App host communication
- Refactored hooks to use new `useAppsSdkBridge` abstraction instead of directly calling `useOpenAiGlobal`
- Deprecated `useOpenAiGlobal` in favor of the bridge pattern
- Added `useMcpAppBridge` hook for accessing MCP host context (theme support implemented)
- Added `@modelcontextprotocol/ext-apps` as dev dependency for type definitions

**Critical issues found:**
- Timeout handlers are not cleared when requests complete successfully, causing timer handles to leak until they fire
- The existing comments about memory leaks and singleton behavior have been addressed in the implementation with `cleanup()` method and warning message

- This PR has a critical timeout cleanup bug that needs fixing before merge
- The timeout cleanup issue causes timer handle leaks and will fire unnecessary cleanup code after successful requests. While the architecture is sound and the refactoring is well-structured, this bug affects core request handling logic and should be fixed.
- Pay close attention to `src/web/bridges/use-mcp-app-bridge.ts` - the timeout handling needs to be corrected

### DIFF
--- a/src/web/bridges/use-mcp-app-bridge.ts
+++ b/src/web/bridges/use-mcp-app-bridge.ts
@@ -10,6 +10,7 @@ import { useSyncExternalStore } from "react";
 type PendingRequest<T> = {
   resolve: (value: T | PromiseLike<T>) => void;
   reject: (reason?: unknown) => void;
+  timeout: NodeJS.Timeout;
 };
 
 type McpAppInitializationOptions = Pick<
@@ -52,6 +53,9 @@ export class McpAppBridge {
 
   public cleanup() {
     window.removeEventListener("message", this.handleMessage);
+    this.pendingRequests.forEach((request) => {
+      clearTimeout(request.timeout);
+    });
     this.pendingRequests.clear();
     this.listeners.clear();
   }
@@ -65,12 +69,13 @@ export class McpAppBridge {
     this.pendingRequests.set(id, {
       resolve: resolve as (value: unknown) => void,
       reject,
+      timeout: setTimeout(() => {
+        reject(new Error("Request timed out"));
+        this.pendingRequests.delete(id);
+      }, this.requestTimeout),
     });
     window.parent.postMessage({ jsonrpc: "2.0", id, method, params }, "*");
-    setTimeout(() => {
-      reject(new Error("Request timed out"));
-      this.pendingRequests.delete(id);
-    }, this.requestTimeout);
+
     return promise;
   }
 
@@ -106,6 +111,7 @@ export class McpAppBridge {
 
     const request = this.pendingRequests.get(data.id);
     if (request) {
+      clearTimeout(request.timeout);
       this.pendingRequests.delete(data.id);
       if (data.error) {
         request.reject(new Error(data.error.message));


src/web/hooks/use-mcp-app-bridge.test.ts
```diff
@@ -0,0 +1,57 @@
+import { act, renderHook, waitFor } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+describe("useMcpAppBridge", () => {
+  let mockPostMessage: ReturnType<typeof vi.fn>;
+
+  const useHook = async () => {
+    const { useMcpAppBridge } = await import("./use-mcp-app-bridge.js");
```
fredericbarthelet: Why do you need to import asynchronously?

src/web/hooks/use-mcp-app-bridge.ts
```diff
@@ -0,0 +1,146 @@
...
+    window.addEventListener("message", this.handleMessage);
+    this.connect();
+  }
+
+  private handleMessage = (event: MessageEvent) => {
+    const data = event.data;
+    if (data?.jsonrpc !== "2.0") return;
+
+    if (data.id !== undefined && this.pendingRequests.has(data.id)) {
+      const request = this.pendingRequests.get(data.id);
```
fredericbarthelet: Use only .get to avoid a double value check

src/web/bridges/use-mcp-app-bridge.ts
```diff
@@ -0,0 +1,146 @@
...
+      const request = this.pendingRequests.get(data.id);
+      if (request) {
+        this.pendingRequests.delete(data.id);
+        if (data.error) {
+          request.reject(new Error(data.error.message));
+        } else {
+          request.resolve(data.result);
+        }
+      }
+      return;
```
fredericbarthelet, line 122: Early return, will make it easier to read

src/web/hooks/use-mcp-app-bridge.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.init();
+    return () => this.listeners.delete(onChange);
+  };
+
+  private request<R extends { method: string; params?: unknown }, T>({
+    method,
+    params,
+  }: R): Promise<T> {
+    const id = this.nextId++;
+    return new Promise((resolve, reject) => {
```
fredericbarthelet: It will be easier to understand with Promise.withResolvers()

src/web/bridges/use-mcp-app-bridge.ts
```diff
@@ -0,0 +1,146 @@
...
+    params,
+  }: R): Promise<T> {
+    const id = this.nextId++;
+    return new Promise((resolve, reject) => {
+      this.pendingRequests.set(id, {
+        resolve: resolve as (value: unknown) => void,
+        reject,
+      });
+      window.parent.postMessage({ jsonrpc: "2.0", id, method, params }, "*");
+    });
```
fredericbarthelet, line 76: You should also use a setTimeout method to fire a rejection past a certain duration. Your current implementation waits forever on a request in case no response is given

src/web/hooks/use-mcp-app-bridge.ts
```diff
@@ -0,0 +1,146 @@
...
+      this.pendingRequests.set(id, {
+        resolve: resolve as (value: unknown) => void,
+        reject,
+      });
+      window.parent.postMessage({ jsonrpc: "2.0", id, method, params }, "*");
+    });
+  }
+
+  private emit() {
+    this.listeners.forEach((l) => {
```
fredericbarthelet: No short syntax please
```suggestion
    this.listeners.forEach((listener) => {
```

src/web/hooks/use-mcp-app-bridge.ts
```diff
@@ -0,0 +1,146 @@
...
+  constructor(options: McpAppInitializationOptions) {
+    this.initialized = false;
+    this.appInitializationOptions = {
+      appInfo: options.appInfo,
+      appCapabilities: {},
+      protocolVersion: LATEST_PROTOCOL_VERSION,
+    };
+  }
+
+  subscribe = (onChange: () => void) => {
```
fredericbarthelet: You should also take the `key` parameter as an input to the `subscribe` method in order to index listeners by the value they are watching and only fire onChange when this specific value change. Your current implementation fires all listeners on every notification of host context change

src/web/types/ext-apps.ts
```diff
@@ -0,0 +1,550 @@
+/**
```
fredericbarthelet: It looks like you got this file from @modelcontextprotocol/ext-apps given the comments in file. You shared the idea of using this package as a dev dependency. Why did you version the file instead?

qchuchu: I wasn't sure I could do it - but it turns out it works well now 👍

src/web/hooks/use-mcp-app-bridge.ts
qchuchu: I needed to update the lib to ESNext though but it works fine

src/web/hooks/use-mcp-app-bridge.test.ts
qchuchu: Because I want to reset modules (hence resetting the instance that is not yet set).

PR #120, opened 2025-12-20
feat: add autosave behavior for biome + noUnusedImport as error

qchuchu:
- Add formatOnSave on biome to ease development and format on work
- Consider noUnusedImports as an error

<h3>Greptile Summary</h3>

Added autosave formatting and stricter import checks to improve code quality and developer experience.

- Enabled format-on-save in VSCode to automatically format code using Biome
- Configured Biome to treat unused imports as errors (`noUnusedImports`)
- Removed unused `useQuery` import from `packages/devtools/src/lib/mcp/index.ts`

- This PR is safe to merge with no risk
- Changes are purely configuration and cleanup - VSCode settings affect only developer experience, Biome rules enforce code quality without breaking existing code, and the only code change removes an unused import which has no runtime impact

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,4 @@
 {
   "editor.defaultFormatter": "biomejs.biome",
-  "editor.formatOnSave": true,
-  "editor.codeActionsOnSave": {
-    "source.fixAll.biome": "explicit"
-  },
   "biome.lsp.bin": "${workspaceFolder}/node_modules/.bin/biome"
 }


.vscode/settings.json
```diff
@@ -1,4 +1,8 @@
 {
   "editor.defaultFormatter": "biomejs.biome",
+  "editor.formatOnSave": true,
```
fredericbarthelet: I like this but I know it's not everyone's taste especially when you have autosave. I'd keep this settings on the user settings side

Issue #103, opened 2025-12-16
Polyfill useLocale


rodincave:
Implements a compatibility layer at hook level between [Apps SDK](https://developers.openai.com/apps-sdk/build/chatgpt-ui#understand-the-windowopenai-api) and [MCP Apps](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#host-context-in-mcpuiinitializeresult). The objective of this issue is to commit to a single implementation of `src/web/hooks/use-locale.ts` hook that both works in an OpenAI Apps SDK hosted iFrame as well as in an MCP App hosted iFrame.

Signature on respective standards for the `locale` properties is exactly the same, just like for useTheme in #97 

### In Apps SDK
```ts
window.openai.locale: en-US
//The user's locale string in BCP 47 format (e.g., "en-US", "fr-FR", "ja-JP").

```
_from https://www.skybridge.tech/api-reference/use-locale_

### In MCP Apps
```ts
  /** User's language/region preference (BCP 47, e.g., "en-US") */
  locale?: string;
```
_From https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx#host-context-in-mcpuiinitializeresult_

### Technical implementation recommendations :
 
- Documentation with mappings to MCP App implementation should be added to `docs/docs/skybridge-core-concepts.md`.

PR #126, opened 2025-12-23
Polyfill theme & locale for MCP App

qchuchu:
Fixes (partially as we don't expose the MCP App resources yet) : 
- #103
- #97
- Polyfill useDisplay Mode

There is one important take here : as MCPUIHostContext type definition from [ext-apps](https://github.com/modelcontextprotocol/ext-apps/blob/main/src/spec.types.ts#L309) does not ensure keys are going to be set for each Host Data, hence exposing them in different hooks (useTheme, useDisplayMode & useLocale), makes the current hook implementation "worse" as before people were sure the values were set.

Now all those hooks returns types like "DisplayMode | undefined"

```typescript
DisplayMode -> DisplayMode | undefined
string -> string | undefined
Theme -> Theme | undefined
```

I don't think we want to downgrade users DevX as this is the promise of Skybridge initially. I'm trying to look some inspiration from React Native, but happy to chat & discuss about this !

<h3>Greptile Summary</h3>

This PR adds polyfill support for `useTheme`, `useLocale`, and `useDisplayMode` hooks to work with both ChatGPT Apps SDK and MCP Apps runtime environments.

- Introduced NOOP helpers (`NOOP_SUBSCRIBE`, `NOOP_GET_SNAPSHOT`) to conditionally bypass bridge operations based on host type
- Refactored both `useAppsSdkBridge` and `useMcpAppBridge` to check `window.skybridge.hostType` and use NOOP functions when not applicable
- The pattern of calling both bridge hooks unconditionally and selecting the value based on hostType is correct and doesn't violate React hook rules
- **Critical issue**: `setDisplayMode` in `use-display-mode.ts` calls `window.openai.requestDisplayMode` unconditionally - this will throw a runtime error in MCP App context where `window.openai` doesn't exist
- Minor style inconsistencies in import extensions (some files use `.js`, others don't)

- PR has a runtime error bug in setDisplayMode for MCP App context, but reading values works correctly
- The core polyfill logic for reading theme/locale/displayMode is correct and well-implemented. However, the setDisplayMode function will fail in MCP App context since it unconditionally calls window.openai.requestDisplayMode. This is a functional bug that affects a key use case.
- src/web/hooks/use-display-mode.ts - setDisplayMode will throw in MCP App context

### DIFF
--- a/src/web/bridges/hooks/use-bridge.ts
+++ b/src/web/bridges/hooks/use-bridge.ts
@@ -1,0 +1,51 @@
+import type { McpUiHostContext } from "@modelcontextprotocol/ext-apps";
+import { useSyncExternalStore } from "react";
+import { AppsSdkBridge } from "../apps-sdk-bridge";
+import { McpAppBridge } from "../mcp-app-bridge";
+
+export type BridgeInterface = Required<
+  Pick<McpUiHostContext, "theme" | "locale">
+>;
+
+type BridgeExternalStore<K extends keyof BridgeInterface> = {
+  subscribe: (onChange: () => void) => () => void;
+  getSnapshot: () => BridgeInterface[K];
+};
+
+const DEFAULT_VALUE_FOR_MCP_APP_BRIDGE: BridgeInterface = {
+  theme: "light",
+  locale: "en-US",
+};
+
+const getExternalStore = <K extends keyof BridgeInterface>(
+  key: K,
+  defaultValue: BridgeInterface[K] = DEFAULT_VALUE_FOR_MCP_APP_BRIDGE[key],
+): BridgeExternalStore<K> => {
+  const hostType = window.skybridge.hostType;
+  if (hostType === "apps-sdk") {
+    const bridge = AppsSdkBridge.getInstance();
+    return {
+      subscribe: bridge.subscribe(key),
+      getSnapshot: () => bridge.getSnapshot(key) as BridgeInterface[K],
+    };
+  }
+  const bridge = McpAppBridge.getInstance();
+  return {
+    subscribe: bridge.subscribe(key),
+    getSnapshot: () =>
+      (bridge.getSnapshot(key) ?? defaultValue) as BridgeInterface[K],
+  };
+};
+
+export const useBridge = <K extends keyof BridgeInterface>(
+  key: K,
+  defaultValue: BridgeInterface[K] = DEFAULT_VALUE_FOR_MCP_APP_BRIDGE[key],
+): BridgeInterface[K] => {
+  const externalStore = getExternalStore(key, defaultValue);
+
+  return useSyncExternalStore(
+    externalStore.subscribe,
+    externalStore.getSnapshot,
+    () => DEFAULT_VALUE_FOR_MCP_APP_BRIDGE[key],
+  );
+};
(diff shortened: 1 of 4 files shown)


fredericbarthelet:
Thanks for putting together this first proposal of complete implementation from end-user hook to Apps SDK and MCP Apps bridges :)

I like your last proposition, and I think it is complementary to your current PR. I'd separate 2 concerns - the implementation strategy to avoid duplication of logic due to rule of hook and the _shitty common interface_ imposed by MCP App.

For the implementation strategy, my biggest concern when looking at your code changes was to see `window.skybridge.hostType` used in both bridge hooks `useAppsSdkBridge` and `useMcpAppBridge`, as well as in end-user hooks like `useLocale`. My initial impression was that `hostType` will only be used in the "selector" layer called `useBridge` in the following diagram:

```mermaid
flowchart TD
    useLocale --> useBridge
    useTheme --> useBridge
    useUserAgent --> useBridge
    useBridge --> useAppsSdkBridge
    useBridge --> useMcpAppBridge
```

I'd stick with the additional following requirements:
- simple to implement L2/L3 hooks (like `useLocale`, `useTheme`...). This translates for me with the following requirement: **there can't be any trace of logic related to host selection in those hooks**. The value returned by the hook must purely be functional and run against the _common interface_. _useLocal.ts_ should still be as simple as `export function useLocale() { return useBridge("locale"); }`

- `useBridge` must export the _common interface_ as an external store. To avoid the issue of rules of hooks, I like the idea of exploiting classes `AppsSdkBridge` and `McpAppBridge` and instanciate them in `useBridge` instead of relying on the bridge specific hook. You can conditionally create the bridge based on hostType and use it in the returned `useSyncExternalStore`. Transposing from host-specific interface to common interface should also be done in `useBridge`.

- host-specific hooks `useAppsSdkBridge` and `useMcpAppBridge` must still be accessible for feature that will work only in one env

WDYT?

As for the _shitty_ common interface because all properties in MCP App can be undefined, I'm happy to "decide" default values waiting for initialization. Those should be designed in `useBridge` in order to improve as much as possible common interface types (and avoid `| undefined` everywhere)

WDYT?


Last small detail - I'd remove all mention of `chatgpt-app` and `AppsSdk`. Use the same naming everywhere :)

src/web/bridges/hooks/use-bridge.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {
+      subscribe: bridge.subscribe(key),
+      getSnapshot: () => bridge.getSnapshot(key) as BridgeInterface[K],
+    };
+  }
+  const bridge = McpAppBridge.getInstance();
+  return {
+    subscribe: bridge.subscribe(key),
+    getSnapshot: () =>
+      (bridge.getSnapshot(key) ?? defaultValue) as BridgeInterface[K],
```
qchuchu, line 36: Casting here

fredericbarthelet: You may be able not to cast if getExternalStore does not take key as an argument and return uncalled bridge methods, don't you think?

fredericbarthelet (approved):
Great implementation @qchuchu ! One small comment on use bridge that may help on avoiding type casting, but otherwise good to go!

src/web/bridges/hooks/use-bridge.ts
qchuchu: Hmmm for the moment it does not seem to make the trick.. I think I need to better rethink how I leverage the interface. I think I'll make a proposal on ext-apps, because I think what happen is that : 

- The keys of the host context can not be defined (theme?:) 
- But it also has a [key: string]: unknown. So I think that the issue is that in that case, it can accept other type definition of theme. I'll dig further on this next week :)

PR #115, opened 2025-12-18
Create skybridge command

harijoe:
<h3>Greptile Summary</h3>

Added new `create-skybridge` CLI tool that scaffolds Skybridge projects by cloning a template repository

**Key Changes:**
- Created new `create-skybridge` package with interactive CLI for project scaffolding
- Added support for project name prompts, directory overwrite handling, and help messages
- Integrated into CI/CD workflows with separate npm token for publishing
- Renamed `publish:skybridge` to `publish:core` for better clarity

**Issues Found:**
- Critical: bin entry path mismatch will prevent the CLI from working after installation

<h3>Confidence Score: 2/5</h3>

- This PR has a critical issue that will prevent the CLI from working after installation
- While the implementation logic is sound and the CI/CD integration is correct, there's a critical bin path mismatch in package.json that will cause the published package to fail when users try to run it. The bin entry points to `dist/index.mjs` but TypeScript compiles to `dist/index.js`
- packages/create-skybridge/package.json requires immediate attention to fix the bin entry path

PR #134, opened 2025-12-29
feat: add create-skybridge

paulleseute:
To clone alpic-ai/apps-sdk-template with a npx command. Can be tested with `pnpx create-skybridge@dev`.

Adapted from #115. `pnpm build` failure is not related to this PR and exists since b2c1d21.

### DIFF
--- a/packages/create-skybridge/package.json
+++ b/packages/create-skybridge/package.json
@@ -23,7 +23,6 @@
   },
   "devDependencies": {
     "@types/node": "^25.0.3",
-    "picocolors": "^1.1.1",
     "typescript": "^5.9.3"
   }
 }
--- a/packages/create-skybridge/tsconfig.json
+++ b/packages/create-skybridge/tsconfig.json
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "target": "ES2020",
-    "module": "ESNext",
-    "moduleResolution": "Node",
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
     "outDir": "dist",
     "rootDir": "src",
     "esModuleInterop": true,
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -152,9 +152,6 @@ importers:
       '@types/node':
         specifier: ^25.0.3
         version: 25.0.3
-      picocolors:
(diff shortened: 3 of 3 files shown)


packages/create-skybridge/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "test:format": "biome ci",
+    "prepublishOnly": "pnpm run build"
+  },
+  "dependencies": {
+    "@clack/prompts": "^0.11.0",
+    "mri": "^1.2.0"
+  },
+  "devDependencies": {
+    "@types/node": "^25.0.3",
+    "picocolors": "^1.1.1",
```
valentinbeggi: Do we need this dep ? I don't see it in imports

packages/create-skybridge/src/index.ts
```diff
@@ -0,0 +1,178 @@
...
+        return;
+    }
+  }
+
+  const root = path.join(cwd, targetDir);
+
+  // 3. Clone the repository
+  prompts.log.step(`Cloning template from ${TEMPLATE_REPO}...`);
+
+  try {
```
valentinbeggi, line 131: not sure the try catch here is necessary since run already exits with code = 1 in case of error, maybe run should throw ?

packages/create-skybridge/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
+    "module": "ESNext",
+    "moduleResolution": "Node",
```
valentinbeggi: Node is deprecated, should use NodeNext
I think you won't have anything to update in your code since you're not importing from other files in index.ts

packages/create-skybridge/src/index.ts
paulleseute: I think it's because we want to want to have something nice to display ("Failed to clone repository") along the actual error. The script is from #115.

fredericbarthelet: Which has been inspired from https://github.com/vitejs/vite/blob/main/packages/create-vite/src/index.ts

fredericbarthelet (approved):

PR #141, opened 2025-12-30
[POLYFILL] feat: add user agent bridge for mcp app

qchuchu:
- Use OpenAI interface for the moment
- Add two subscriber : but we should think of a better behavior for our listeners. Currently this is OK so I don't want to complexify, but for future interface it might be relevant to change the way we store listeners

<h3>Greptile Summary</h3>

Added `userAgent` support to the MCP app bridge by creating an adapter that maps MCP host context properties (`platform`, `deviceCapabilities`) to the unified `BridgeInterface`. The PR also refactored the `subscribe` method in `McpAppBridge` to support subscribing to multiple keys at once, and extracted bridge-related types to a separate file for better organization.

Key changes:
- Implemented `getMcpAppExternalStore` in `mcp-app-adapter.ts` that provides external store adapters for all bridge properties including the new `userAgent`
- Enhanced `McpAppBridge.subscribe()` with overloaded signatures to accept either a single key or array of keys
- Simplified `use-bridge.ts` by delegating MCP-specific logic to the adapter
- Updated `use-user.ts` to use the unified `useBridge` hook instead of `useAppsSdkBridge` for `userAgent`

- This PR is mostly safe to merge with one logical issue that needs fixing
- The implementation is well-structured and the refactoring improves code organization. However, there's a logical error in the platform-to-device-type mapping that could cause incorrect device type values for iOS/Android platforms.
- Pay close attention to `packages/core/src/web/bridges/mcp-app-adapter.ts` for the platform mapping logic

### DIFF
--- a/packages/core/src/web/bridges/mcp-app-adapter.ts
+++ b/packages/core/src/web/bridges/mcp-app-adapter.ts
@@ -3,7 +3,7 @@ import type {
   McpUiRequestDisplayModeRequest,
   McpUiRequestDisplayModeResult,
 } from "@modelcontextprotocol/ext-apps";
-import type { BridgeExternalStore, BridgeSubscribe } from "./hooks/types.js";
+import type { BridgeExternalStore } from "./hooks/types.js";
 import { McpAppBridge } from "./mcp-app-bridge.js";
 import type { BridgeInterface, Methods } from "./types.js";
 


packages/core/src/web/bridges/hooks/use-bridge.ts
```diff
@@ -56,6 +65,40 @@ const getExternalStore = <K extends keyof BridgeInterface>(
...
+      subscribe: (onChange) => {
+        const deviceCapabilitiesUnsubscribe =
+          bridge.subscribe("deviceCapabilities")(onChange);
+        const platformUnsubscribe = bridge.subscribe("platform")(onChange);
+
+        return () => {
+          deviceCapabilitiesUnsubscribe();
+          platformUnsubscribe();
+        };
+      },
```
fredericbarthelet: I'd instead update subscribe interface to accept an array of values. Please use an overloading to avoid having to change syntax on all subscribe calls that currently uses only a single string argument :)

```suggestion
      subscribe: bridge.subscribe(["deviceCapabilities", "platform"])
```

qchuchu: Done :)

packages/core/src/web/bridges/mcp-app-adapter.ts
fredericbarthelet, line 1: Almost there I think :) Well done
The adapter subscribe and getSnapshot methods are both returning stuff that are indexed by the common interface. I believe it would be easier to read if both were at the same location.

```ts
const adapter = () => {
  const bridge = McpAppBridge.getInstance();

  return {
    theme: {
      subscribe: bridge.subscribe("theme"),
      getSnapshot: () => bridge.getSnapshot("theme") ?? "light"
    },
    //...
```

This could even be auto improved later on (subscribe arguments are the one used in getSnapshot). You could have something like

```ts
const adapter = () => {
  return {
    theme: (["theme"], ({ theme }) => theme ?? "light"
    //...
```

where the callback first arg would be used to subscribe and inject args in callback in second arg callbacks args with bridge.getSnapshot

WDYT ?

packages/core/src/web/bridges/mcp-app-adapter.ts
```diff
@@ -19,3 +21,82 @@ export const requestDisplayMode: Methods["requestDisplayMode"] = ({ mode }) => {
...
+    userAgent: () =>
+      createExternalStore(
+        ["platform", "deviceCapabilities"],
+        ({ platform, deviceCapabilities }) => ({
+          device: {
+            type: platform
+              ? platform === "web"
+                ? "desktop"
+                : platform
+              : "unknown",
```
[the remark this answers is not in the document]
qchuchu: Actually McpUiHostContext does not allow this interface values for platform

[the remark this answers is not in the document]
qchuchu: but platform does not propose those values.

image

here is the interface of platform from McpUiHostContext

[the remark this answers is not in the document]
qchuchu: don't you see in the image? the platform can be either web | desktop | mobile or undefined

packages/core/src/web/bridges/mcp-app-adapter.ts
qchuchu: @fredericbarthelet Found something that might work :)

packages/core/src/web/bridges/mcp-app-adapter.ts
```diff
@@ -19,3 +21,78 @@ export const requestDisplayMode: Methods["requestDisplayMode"] = ({ mode }) => {
 
   throw new Error("Modal display mode is not accessible in MCP App.");
 };
+
+type PickContext<K extends readonly (keyof McpUiHostContext)[]> = {
+  [P in K[number]]: McpUiHostContext[P];
+};
+
+const createExternalStore = <
+  const Keys extends readonly (keyof McpUiHostContext)[],
```
fredericbarthelet, line 30: Why do you need a const keyword here? First time I see it used for a type generic

packages/core/src/web/bridges/mcp-app-adapter.ts
```diff
@@ -19,3 +21,78 @@ export const requestDisplayMode: Methods["requestDisplayMode"] = ({ mode }) => {
...
+    getSnapshot: () => {
+      const context = Object.fromEntries(
+        keys.map((k) => [k, bridge.getSnapshot(k)]),
+      ) as PickContext<Keys>;
+      return getSnapshot(context);
+    },
+  };
+};
+
+export const getMcpAppExternalStore = <K extends keyof BridgeInterface>(
```
fredericbarthelet: you can export the `mcpAppAdaptor` object directly I believe. The `getMcpAppExternalStore` does not add any value here

qchuchu: Nop because of the window variable that is not defined in server time - this cause vite to fail

qchuchu: (but we could change to set the maxheight by default to 400 - currently i'm not fixed on this)

fredericbarthelet: I think you simplify adapter  (you don't need each adapter value to be a callback, you can call createExternalStore directly, no?)

```suggestion
export const mcpAppAdaptor = {
    theme: createExternalStore(["theme"], ({ theme }) => theme ?? "light"),

```

packages/core/src/web/bridges/mcp-app-adapter.ts
qchuchu: Well it's necessary to narrow down the Keys accessible in the context. 

If I don't do this actually, the keys accessible by the callback are not defined

image

packages/core/src/web/bridges/mcp-app-adapter.ts
qchuchu: Indeed, cleaner :)

qchuchu: I needed still to put it inside a function, because otherwise Vite tries to calls the bridge instance which requires to check window which does not exist.

qchuchu:
> LGTM! Fantastic work @qchuchu , this will make adapters much easier to read. Some minor comments though on making less anonymous callbacks
> 
> In a follow up PR, I'd prioritize the following :
> 
> * move utility methods to a separate file (`createExternalStore`)
> * add an abstract base class or interface for subscribe and getSnapshot methods on each bridge
> * add multiple keys subscribe method on apps sdk bridge too
> * implement same adaptor (even if 1-1 mapping ATM) on apps sdk bridge
> 
> WDYT ?

Indeed agree to make it more generic on the openai side

PR #142, opened 2025-12-30
Remove defaultValue override option

fredericbarthelet:
No useBridge implementation uses optional argument `defaultValue` to override the defaults specified in `useBridge`. This argument is unnecessary ATM.

<h3>Greptile Summary</h3>

Removed the unused `defaultValue` parameter from both `useBridge` and `getExternalStore` functions. The default value is now always sourced from `DEFAULT_VALUE_FOR_MCP_APP_BRIDGE[key]` internally within `getExternalStore`, maintaining the same behavior while simplifying the API.

- Simplified function signatures by removing optional `defaultValue` parameter
- Default values are still used internally via `DEFAULT_VALUE_FOR_MCP_APP_BRIDGE`
- No consumers in the codebase were passing the second parameter
- Behavior remains unchanged - defaults are still applied when bridge values are unavailable

- This PR is safe to merge with no risk
- The change removes an unused optional parameter that was never utilized by any consumer in the codebase. The default value behavior is preserved by moving the default assignment inside the function. All existing usages of `useBridge` only pass a single argument, so this is a non-breaking change.

### DIFF
--- a/packages/core/src/web/bridges/hooks/use-bridge.ts
+++ b/packages/core/src/web/bridges/hooks/use-bridge.ts
@@ -25,8 +25,8 @@ const DEFAULT_VALUE_FOR_MCP_APP_BRIDGE: BridgeInterface = {
 
 const getExternalStore = <K extends keyof BridgeInterface>(
   key: K,
-  defaultValue: BridgeInterface[K] = DEFAULT_VALUE_FOR_MCP_APP_BRIDGE[key],
 ): BridgeExternalStore<K> => {
+  const defaultValue = DEFAULT_VALUE_FOR_MCP_APP_BRIDGE[key];
   const hostType = window.skybridge.hostType;
   if (hostType === "apps-sdk") {
     const bridge = AppsSdkBridge.getInstance();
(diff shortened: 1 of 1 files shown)


packages/core/src/web/bridges/hooks/use-bridge.ts
```diff
@@ -25,8 +25,8 @@ const DEFAULT_VALUE_FOR_MCP_APP_BRIDGE: BridgeInterface = {
 
 const getExternalStore = <K extends keyof BridgeInterface>(
   key: K,
-  defaultValue: BridgeInterface[K] = DEFAULT_VALUE_FOR_MCP_APP_BRIDGE[key],
 ): BridgeExternalStore<K> => {
+  const defaultValue = DEFAULT_VALUE_FOR_MCP_APP_BRIDGE[key];
```
qchuchu, line 29: Fine by me, but in the #141 I'll move it directly in the external store definition, as it'll reduce type safety declaration.

qchuchu (approved):

packages/core/src/web/bridges/hooks/use-bridge.ts
fredericbarthelet: In the `getServerSnapshot` implementation :) ? Agreed that it'll be better

PR #148, opened 2025-12-30
Fix Polyfill on ChatGPT

qchuchu:
<h3>Greptile Summary</h3>

Changed the `window.skybridge` initialization script from a regular script to a module script (`type="module"`) to comply with ChatGPT's Content Security Policy that blocks non-module scripts. Added SSR safety check to `DEFAULT_VALUE_FOR_MCP_APP_BRIDGE.maxHeight` using `typeof window !== "undefined"` guard with fallback to 400.

Key changes:
- Template scripts now use `type="module"` for ChatGPT CSP compatibility
- `maxHeight` default value now safely handles server-side rendering scenarios
- Both development and production templates updated consistently

- This PR is safe to merge with minor considerations about module script execution timing
- The changes correctly address ChatGPT CSP requirements and add proper SSR safety. Module scripts execute in document order so `window.skybridge` will be available before widget code runs. However, there's no defensive check for `window.skybridge` existence in runtime code (line 30 of use-bridge.ts), which could theoretically cause issues if the execution order assumptions are violated in edge cases
- No files require special attention - the changes are minimal and focused

### DIFF
--- a/packages/core/src/web/bridges/hooks/use-bridge.ts
+++ b/packages/core/src/web/bridges/hooks/use-bridge.ts
@@ -8,7 +8,9 @@ type BridgeExternalStore<K extends keyof BridgeInterface> = {
   getSnapshot: () => BridgeInterface[K];
 };
 
-const getDefaultValueFromMcpAppBridge = <K extends keyof BridgeInterface>(key: K): BridgeInterface[K] => {
+const getDefaultValueFromMcpAppBridge = <K extends keyof BridgeInterface>(
+  key: K,
+): BridgeInterface[K] => {
   const DEFAULT_VALUES_FOR_MCP_APP_BRIDGE: BridgeInterface = {
     theme: "light",
     locale: "en-US",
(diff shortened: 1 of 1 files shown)


valentinbeggi (approved):

packages/core/src/web/bridges/hooks/use-bridge.ts
```diff
@@ -20,7 +20,7 @@ const DEFAULT_VALUE_FOR_MCP_APP_BRIDGE: BridgeInterface = {
       left: 0,
     },
   },
-  maxHeight: window.innerHeight,
+  maxHeight: typeof window !== "undefined" ? window.innerHeight : 400,
```
fredericbarthelet:
```suggestion
  maxHeight: window?.innerHeight ?? 400,
```

packages/core/src/server/templates/development.hbs
```diff
@@ -1,5 +1,7 @@
 <base href="{{serverUrl}}" />
-<script> window.skybridge = { hostType: "{{hostType}}" }; </script>
+<script type="module"> 
+  window.skybridge = { hostType: "{{hostType}}" };
+</script>
```
fredericbarthelet: why multiline ?

packages/core/src/web/bridges/hooks/use-bridge.ts
qchuchu: image

However I get this when I run the dev environment of vite

qchuchu: But that's because I'm setting a constant with a window default value. As vite is evaluating the file for ssr, then it tries to access window which indeed not defined (no declaration AT all).

A solution is to put the default value getter into a function

PR #151, opened 2025-12-31
feat: add create-skybridge smoke test

paulleseute:
We're adding a unit test to make sure the creation script does not break on further changes.

This PR addresses few things:
- the script main function is explicitly invoked by the actual entrypoint, not the underlying modules: IMO it's cleaner and makes it easier to test
- in the same fashion, `argv` are read when the script is actually run, not when the module is loaded
- we were not properly sanitizing user input regarding project dir, this is fixed with `formatDir` becoming `sanitizeDir`

<h3>Greptile Summary</h3>

Adds unit test for `create-skybridge` script to prevent regressions from future changes. The PR includes three main improvements:

- **Refactored entry point**: The `init()` function is now explicitly invoked by `index.js`, making it easier to test
- **Parameterized argv**: Arguments are read when the script runs (not at module load), allowing the test to pass custom arguments
- **Enhanced input sanitization**: Renamed `formatTargetDir` to `sanitizeTargetDir` with additional security measures to prevent path traversal and filter special characters

The test verifies that the script can scaffold a project with sanitized input (e.g., `../../test-1234//project$` becomes `test-1234/project`).

- The changes are well-structured and improve code quality by making the script testable and adding security measures. The refactoring is minimal and non-breaking, with clear improvements to input sanitization. The test validates the core functionality.

### DIFF
--- a/packages/create-skybridge/src/index.ts
+++ b/packages/create-skybridge/src/index.ts
@@ -4,15 +4,6 @@ import { fileURLToPath } from "node:url";
 import * as prompts from "@clack/prompts";
 import mri from "mri";
 
-const argv = mri<{
-  help?: boolean;
-  overwrite?: boolean;
-}>(process.argv.slice(2), {
-  boolean: ["help", "overwrite"],
-  alias: { h: "help" },
-});
-
-const cwd = process.cwd();
 const defaultProjectName = "skybridge-project";
 
 // prettier-ignore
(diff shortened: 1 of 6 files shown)


packages/create-skybridge/src/index.ts
fredericbarthelet, line 123: shouldn't you keep unsatinized project name here - sanitizing is only required for the path no?

fredericbarthelet: In fact package name comes with its own sets of rules: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#name

paulleseute: Seen with @fredericbarthelet: this is what vite does, let's keep this behavior for now.

PR #155, opened 2025-12-31
[POLYFILL] callTool

qchuchu:
Solves #145 

<h3>Greptile Summary</h3>

This PR adds a `callTool` polyfill to enable MCP apps to use the `useCallTool` hook with the same interface as Apps SDK apps.

**Key changes:**
- Implemented `callTool` method in both `apps-sdk-adapter.ts` (delegates to `window.openai.callTool`) and `mcp-app-adapter.ts` (uses MCP bridge to call `tools/call` method)
- Updated `useCallTool` hook to use `getBridgeMethods().callTool` instead of directly accessing `window.openai.callTool`, enabling runtime bridge selection
- Changed `CallToolResponse.content` type from restricted text-only array to full `CallToolResult["content"]` union type to support all MCP content types (text, image, resource, embedded_resource)
- MCP adapter extracts text content from response and joins into result string, while preserving full content array and metadata
- Updated test suite to mock `skybridge.hostType` and adapted type assertions to match new `CallToolResponse` signature

- This PR is safe to merge with no identified issues
- The implementation follows established patterns in the codebase for bridge polyfills. The changes are well-structured, properly typed, and include comprehensive test coverage. The type changes from text-only to full `CallToolResult["content"]` correctly reflect the MCP SDK's capabilities and maintain backward compatibility since the changes are additive.

### DIFF
--- a/packages/core/src/web/bridges/mcp-app-adapter.ts
+++ b/packages/core/src/web/bridges/mcp-app-adapter.ts
@@ -30,15 +30,13 @@ export const callTool = async <
     },
   });
 
-  const text: string[] = [];
-
-  response.content.forEach((content) => {
-    if (content.type === "text") {
-      text.push(content.text);
-    }
-  });
-
-  const result = text.join("\n");
+  const result = response.content
+    .filter(
+      (content): content is { type: "text"; text: string } =>
+        content.type === "text",
+    )
+    .map(({ text }) => text)
+    .join("\n");
 
   return {
     content: response.content,
--- a/packages/core/src/web/hooks/use-call-tool.ts
+++ b/packages/core/src/web/hooks/use-call-tool.ts
@@ -114,8 +114,8 @@ export const useCallTool = <
     toolArgs: ToolArgs,
   ): Promise<CombinedCallToolResponse> => {
     setCallToolState({ status: "pending", data: undefined, error: undefined });
+    const { callTool } = getBridgeMethods();
     try {
-      const { callTool } = getBridgeMethods();
       const data = await callTool<ToolArgs, CombinedCallToolResponse>(
         name,
         toolArgs,


packages/core/src/web/bridges/mcp-app-adapter.ts
```diff
@@ -5,9 +5,49 @@ import type {
...
+
+  const text: string[] = [];
+
+  response.content.forEach((content) => {
+    if (content.type === "text") {
+      text.push(content.text);
+    }
+  });
+
+  const result = text.join("\n");
```
fredericbarthelet:
```suggestion
  const result = response.content.filter(({ type }) => type === "text").map(({ text }) => text).join("\n");
```

packages/core/src/web/bridges/mcp-app-adapter.ts
```diff
@@ -5,9 +5,49 @@ import type {
...
+
+  return {
+    content: response.content,
+    structuredContent: response.structuredContent ?? {},
+    isError: response.isError ?? false,
+    result,
+    meta: response._meta ?? {},
+  } as ToolResponse;
```
fredericbarthelet, line 47: Why do you need typecasting here?

packages/core/src/web/hooks/use-call-tool.ts
```diff
@@ -113,10 +115,11 @@ export const useCallTool = <
   ): Promise<CombinedCallToolResponse> => {
     setCallToolState({ status: "pending", data: undefined, error: undefined });
     try {
-      const data = await window.openai.callTool<
-        ToolArgs,
-        CombinedCallToolResponse
-      >(name, toolArgs);
+      const { callTool } = getBridgeMethods();
```
fredericbarthelet: must be outside of the try

packages/core/src/web/bridges/mcp-app-adapter.ts
qchuchu: Yeah I know but ts is the worst about that. Using a typeguard will be cleaner

packages/core/src/web/bridges/mcp-app-adapter.ts
qchuchu: Well it's something we can see in the implementation - calling a tool is like does a fetch call (external) so I cannot ensure programmatically that the response will be of this type.

I could add a type hint directly on the callTool Request. But for now it don't think it's worth it.

WDYT ?

PR #158, opened 2026-01-02
[Polyfill] feat: add use tool info support for MCP App

### LINKED ISSUE #144: Polyfill - useToolInfo

qchuchu:
Fixes #144 

<h3>Greptile Summary</h3>

Added MCP App support for the `useToolInfo` hook by implementing tool state tracking in `McpAppBridge` and creating adapter functions for `toolInput`, `toolOutput`, and `toolResponseMetadata`. The implementation handles three MCP notification types: `tool-input`, `tool-result`, and `tool-cancelled`, allowing the hook to work across both apps-sdk and mcp-app runtime environments.

- Extended `McpAppBridge` to track tool state (`input`, `result`, `cancelled`)
- Added `toolInput`, `toolOutput`, and `toolResponseMetadata` to `BridgeInterface` type
- Implemented notification handlers for tool lifecycle events
- Updated `useToolInfo` to use generic `useBridge` instead of `useAppsSdkBridge`
- Added comprehensive test coverage for MCP App host environment

- The implementation follows established patterns in the codebase (similar to previous polyfill PRs), has comprehensive test coverage, correctly handles null/undefined cases, and properly integrates with the existing bridge architecture. The previously identified test issue was already fixed.

### DIFF
--- a/packages/core/src/web/bridges/mcp-app-bridge.ts
+++ b/packages/core/src/web/bridges/mcp-app-bridge.ts
@@ -21,7 +21,9 @@ type McpAppInitializationOptions = Pick<
 >;
 
 export type McpToolState = {
-  toolInput: McpUiToolInputNotification["params"]["arguments"] | null;
+  toolInput: NonNullable<
+    McpUiToolInputNotification["params"]["arguments"]
+  > | null;
   toolResult: McpUiToolResultNotification["params"] | null;
   toolCancelled: McpUiToolCancelledNotification["params"] | null;
 };
@@ -220,7 +222,7 @@ export class McpAppBridge {
         return;
       case "ui/notifications/tool-input":
         this.updateContext({
-          toolInput: data.params.arguments,
+          toolInput: data.params.arguments ?? {},
         });
         return;
       case "ui/notifications/tool-result":


packages/core/src/web/hooks/use-tool-info.test.ts
```diff
@@ -8,70 +9,195 @@ import {
...
   });
 
-  it("should eventually return tool output and response metadata once tool call completes", async () => {
-    const toolOutput = {
-      name: "pikachu",
-      color: "yellow",
-      description:
-        "When several of these POKéMON gather, their\felectricity could build and cause lightning storms.",
+  describe("mcp-app host", () => {
+    let mockPostMessage: ReturnType<typeof vi.fn>;
```
fredericbarthelet:
```suggestion
    let mockPostMessage: Mock;
```

fredericbarthelet: + you never reassign it

```suggestion
    const mockPostMessage = vi.fn();
```

packages/core/src/web/hooks/use-tool-info.test.ts
```diff
@@ -8,70 +9,195 @@ import {
...
+      OpenAiProperties,
+      "toolInput" | "toolOutput" | "toolResponseMetadata"
+    >;
 
-  afterEach(() => {
-    vi.unstubAllGlobals();
-    vi.resetAllMocks();
-  });
+    beforeEach(() => {
+      OpenaiMock = {
```
fredericbarthelet, line 24: this can be defined as const outside the scope of the beforeEach too

packages/core/src/web/hooks/use-tool-info.test.ts
```diff
@@ -8,70 +9,195 @@ import {
...
-      name: "pikachu",
-      color: "yellow",
-      description:
-        "When several of these POKéMON gather, their\felectricity could build and cause lightning storms.",
+  describe("mcp-app host", () => {
+    let mockPostMessage: ReturnType<typeof vi.fn>;
+
+    beforeEach(() => {
+      mockPostMessage = vi.fn();
+      Object.defineProperty(window, "parent", {
```
fredericbarthelet: you should use `vi.stubGlobal` to stub something from the `window` space

packages/core/src/web/hooks/use-tool-info.test.ts
```diff
@@ -8,70 +9,195 @@ import {
...
+        value: { postMessage: mockPostMessage },
+        writable: true,
+        configurable: true,
+      });
+      vi.stubGlobal("skybridge", { hostType: "mcp-app" });
+      McpAppBridge.resetInstance();
```
fredericbarthelet: reset should be in afterEach, not beforeEach (cleanup at the end, not at the beginning)

packages/core/src/web/hooks/use-tool-info.test.ts
```diff
@@ -8,70 +9,195 @@ import {
...
+    });
+
+    const initializeBridge = () => {
+      const initCall = mockPostMessage.mock.calls.find(
+        (call) => call[0].method === "ui/initialize",
+      );
+
+      if (initCall) {
+        act(() => {
+          window.dispatchEvent(
```
fredericbarthelet: you should use `fireEvent` API instead - it does the same `window.dispatchEvent` under the hood, but it adds the React testing library event wrapper

packages/core/src/web/hooks/use-tool-info.test.ts
```diff
@@ -8,70 +9,195 @@ import {
...
+      vi.resetAllMocks();
+    });
+
+    const initializeBridge = () => {
+      const initCall = mockPostMessage.mock.calls.find(
+        (call) => call[0].method === "ui/initialize",
+      );
+
+      if (initCall) {
+        act(() => {
```
fredericbarthelet: this should be implemented in the mock behavior 

```suggestion
     const mockPostMessage = vi.fn((message: unknown) => {
    if (typeof message === "object" && message !== null && "method" in message && message.method === "ui/initialize") {
      act(() => fireEvent(window, (
        new MessageEvent("message", {
          data: {
            jsonrpc: "2.0",
            result: {
              protocolVersion: "2025-06-18",
              hostInfo: { name: "test-host", version: "1.0.0" },
              hostCapabilities: {},
              hostContext: { theme: "light" },
            },
          },
        }))
      ));
    }
  });
```

packages/core/src/web/hooks/use-tool-info.test.ts
```diff
@@ -8,70 +9,195 @@ import {
...
+      });
+    });
+
+    it("should return tool input from tool-input notification", async () => {
+      const { result } = renderHook(() => useToolInfo());
+
+      initializeBridge();
+
+      act(() => {
+        window.dispatchEvent(
```
fredericbarthelet: same here

packages/core/src/web/hooks/use-tool-info.test.ts
```diff
@@ -8,70 +9,195 @@ import {
...
+      const { result } = renderHook(() => useToolInfo());
+
+      initializeBridge();
+
+      await waitFor(() => {
+        expect(result.current).toMatchObject({
+          status: "pending",
+          isPending: true,
+          isSuccess: false,
+          input: {},
```
fredericbarthelet: does this mean you always get an empty object as initial input? This behavior differs from Apps SDK implementation where input is always defined, and always equal to tool input args. Having an empty object could trigger unwanted side effects (especially on all code that rely on input being typed at the tool input, i.e. accessing nested props without `?`). I think we should either update the hook type to reflect that input can be undefined (in that case, before the `ui/notifications/tool-input` notification is received in the MCP App environment)

packages/core/src/web/bridges/mcp-app-bridge.ts
```diff
@@ -16,12 +19,30 @@ type McpAppInitializationOptions = Pick<
...
+};
+
+export type McpAppBridgeContext = McpUiHostContext & McpToolState;
+
+export type McpAppBridgeKey = keyof McpAppBridgeContext;
+
 const LATEST_PROTOCOL_VERSION = "2025-11-21";
 
+const isToolStateKey = (key: McpAppBridgeKey): key is keyof McpToolState =>
+  key === "input" || key === "result" || key === "cancelled";
```
fredericbarthelet:
```suggestion
  ["input", "result", "cancelled"].includes(key)
```

packages/core/src/web/bridges/mcp-app-bridge.ts
```diff
@@ -16,12 +19,30 @@ type McpAppInitializationOptions = Pick<
...
+
 export class McpAppBridge {
   private static instance: McpAppBridge | null = null;
   public context: McpUiHostContext | null = null;
-  private listeners = new Map<keyof McpUiHostContext, Set<() => void>>();
+  public toolState: McpToolState = {
+    input: null,
+    result: null,
+    cancelled: null,
+  };
```
fredericbarthelet: Given the fact that you now define an `McpAppBridgeContext` that joins toolState and host context, you can use this type for your context:
```ts
public context: McpAppBridgeContext = {
    input: null,
    result: null,
    cancelled: null,
  }
```

WDYT?

packages/core/src/web/hooks/use-tool-info.test.ts
qchuchu: Ok yes I see - thanks for the heads up. 

In our case currently the hook says (for the output for example) that it's either null or defined if it's success.

We could change the interface and say the input & the output are accessible only on success - or we add another partial state where only the input is available. 

WDYT ?

packages/core/src/web/bridges/mcp-app-bridge.ts
```diff
@@ -175,6 +203,27 @@ export class McpAppBridge {
 
     if (data.method === "ui/notifications/host-context-changed") {
       this.setContext({ ...this.context, ...data.params });
+      return;
+    }
+
+    if (data.method === "ui/notifications/tool-input") {
```
fredericbarthelet: you should use a switch on `data.method`. It will improve readability now that you have multiple conditional checks on value

packages/core/src/web/bridges/mcp-app-bridge.ts
qchuchu: I was wondering what was best - but yes we could have one unified context.

packages/core/src/web/hooks/use-tool-info.test.ts
qchuchu: Seen together -> We are going to add a state in our hook that represents an intermediary state when the ToolInput is not yet set

packages/core/src/web/hooks/use-tool-info.test.ts
qchuchu: I did a change

packages/core/src/web/bridges/mcp-app-bridge.ts
qchuchu: Done

packages/core/src/web/bridges/mcp-app-bridge.ts
qchuchu: Done

packages/core/src/web/bridges/mcp-app-bridge.ts
qchuchu: Not needed now that we merged states

packages/core/src/web/hooks/use-tool-info.test.ts
qchuchu: I've added then an idle state to represent when the input are yet not available (we consider that the tool call haven't yet be done).

packages/core/src/web/hooks/use-tool-info.ts
```diff
@@ -38,19 +50,28 @@ type ToolSignature = {
   responseMetadata: UnknownObject;
 };
 
+function deriveStatus(
+  input: Record<string, unknown> | null,
+  output: Record<string, unknown> | null,
+  responseMetadata: Record<string, unknown> | null,
+): "idle" | "pending" | "success" {
+  if (input === null) {
```
fredericbarthelet, line 58: just to make sure: apps sdk still exposes `{}` as input when the tool does not require any inputs. Does ext-apps does that as well? If it exposes null as a valid tool input (for tools that do not expect any input), we will be stuck in idle

qchuchu: Good catch. I'm going to set to an empty object when we receive the notification if it's undefined (based on the ext-apps definition it can be.

Issue #152, opened 2025-12-31
Make bridge adapter on Apps SDK + create a shared interface on getSnapshot & subscribe of the bridges


qchuchu:
In a follow up PR, I'd prioritize the following :
- move utility methods to a separate file (`createExternalStore`)
- add an abstract base class or interface for subscribe and getSnapshot methods on each bridge
- add multiple keys subscribe method on apps sdk bridge too
- implement same adaptor (even if 1-1 mapping ATM) on apps sdk bridge

WDYT ?

_Originally posted by @fredericbarthelet in https://github.com/alpic-ai/skybridge/pull/141#pullrequestreview-3620735643_

PR #164, opened 2026-01-02
[BUG][POLYFILL] Add better value modification handling in adapter + create class

### LINKED ISSUE #160: [BUG][POLYFILL] useLayout & useUser cause infinite rerender on MCP Apps host
- This is due to the `mcp-app-renderer.ts` that always create a new object when calling the getSnapshot method
- useSyncExternalStore compares object and decide to rerender if the object is different
- This is only on the userAgent & safeArea

qchuchu:
Fixes #160 + #152

<h3>Greptile Summary</h3>

Refactors bridge adapters from module exports to singleton classes, improving architecture and enabling better value management. The PR introduces a `BaseAdapter` class that both `AppsSdkAdapter` and `McpAppAdapter` extend, and adds deep equality checking with `dequal` to maintain referential stability for complex objects in the MCP adapter. This prevents unnecessary re-renders when object values haven't actually changed. The `requestDisplayMode` API is also simplified to accept the mode directly rather than wrapped in an object. Tests are expanded to cover both adapter types and verify the referential stability behavior.

- This PR is safe to merge with only minor optimization opportunities
- The refactoring is well-structured with proper singleton patterns, comprehensive tests covering both adapters, and the core functionality appears correct. The main issue is an inconsistency in how AppsSdkAdapter handles store caching compared to McpAppAdapter - it creates new objects on each call rather than caching them. While React's useSyncExternalStore should handle this gracefully, it's inefficient and inconsistent with the MCP adapter's approach. The deep equality checking with dequal is properly implemented for MCP, and all method signatures are correctly adapted.
- Pay attention to `packages/core/src/web/bridges/apps-sdk-adapter.ts` for the store caching optimization

### DIFF
--- a/packages/core/src/web/bridges/hooks/use-bridge.ts
+++ b/packages/core/src/web/bridges/hooks/use-bridge.ts
@@ -1,23 +1,12 @@
 import { useSyncExternalStore } from "react";
-import { AppsSdkAdaptor } from "../adaptors/apps-sdk-adaptor.js";
-import { McpAppAdaptor } from "../adaptors/mcp-app-adaptor.js";
-import type { BridgeInterface, ExternalStore } from "../types.js";
-
-const getExternalStore = <K extends keyof BridgeInterface>(
-  key: K,
-): ExternalStore<K> => {
-  const hostType = window.skybridge.hostType;
-  if (hostType === "apps-sdk") {
-    return AppsSdkAdaptor.getInstance().getExternalStore(key);
-  }
-
-  return McpAppAdaptor.getInstance().getExternalStore(key);
-};
+import type { BridgeInterface } from "../types.js";
+import { useAdaptor } from "./use-adaptor.js";
 
 export const useBridge = <K extends keyof BridgeInterface>(
   key: K,
 ): BridgeInterface[K] => {
-  const externalStore = getExternalStore(key);
+  const adaptor = useAdaptor();
+  const externalStore = adaptor.getExternalStore(key);
 
   return useSyncExternalStore(
     externalStore.subscribe,


packages/core/src/web/bridges/adaptors/apps-sdk-adaptor.ts
```diff
@@ -1,19 +1,53 @@
...
+
+  public getExternalStore<K extends keyof BridgeInterface>(
+    key: K,
+  ): BridgeExternalStore<K> {
+    const bridge = AppsSdkBridge.getInstance();
+    return {
+      subscribe: bridge.subscribe(key),
+      getSnapshot: () => bridge.getSnapshot(key),
+    };
+  }
```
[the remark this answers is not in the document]
qchuchu: Yes but it was already the case before no ? As we are destructuring after in the useSyncExternalStore we pass directly the methods. The only thing I see is that we always send a new getSnapshot method.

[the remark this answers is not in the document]
qchuchu: It's too verbose and does not have any impact

packages/core/src/web/hooks/use-layout.test.ts
```diff
@@ -1,61 +1,152 @@
...
+  describe("mcp-app host type", () => {
+    let mockPostMessage: ReturnType<typeof vi.fn>;
+
+    beforeEach(() => {
+      mockPostMessage = vi.fn();
+      Object.defineProperty(window, "parent", {
+        value: { postMessage: mockPostMessage },
+        writable: true,
+        configurable: true,
+      });
```
fredericbarthelet: why not use your new host initializer mock here?

packages/core/src/web/hooks/use-layout.test.ts
```diff
@@ -1,61 +1,152 @@
...
+      vi.stubGlobal("skybridge", { hostType: "mcp-app" });
+      McpAppBridge.resetInstance();
+      McpAppAdapter.resetInstance();
```
fredericbarthelet: should be in after only

packages/core/src/web/hooks/use-layout.test.ts
qchuchu: My bad I didn't adapt the test

packages/core/src/web/hooks/use-layout.test.ts
```diff
@@ -1,61 +1,152 @@
...
+        theme: "light",
+        viewport: { maxHeight: 500 },
+        safeAreaInsets: { top: 44, right: 0, bottom: 34, left: 0 },
+      });
+
+      await waitFor(() => {
+        expect(result.current.safeArea).toBeDefined();
+      });
+
+      const firstSafeAreaRef = result.current.safeArea;
```
fredericbarthelet: I thought you were mentionning a React ref here. WDYT of `initialSafeArea` instead?

packages/core/src/web/hooks/use-user.test.ts
```diff
@@ -1,56 +1,147 @@
...
+  describe("mcp-app host type", () => {
+    let mockPostMessage: ReturnType<typeof vi.fn>;
+
+    beforeEach(() => {
+      mockPostMessage = vi.fn();
+      Object.defineProperty(window, "parent", {
+        value: { postMessage: mockPostMessage },
+        writable: true,
+        configurable: true,
+      });
```
fredericbarthelet: same here

packages/core/src/web/hooks/use-user.test.ts
```diff
@@ -1,56 +1,147 @@
...
+        locale: "en-US",
+        platform: "web",
+        deviceCapabilities: { hover: true, touch: false },
+      });
+
+      await waitFor(() => {
+        expect(result.current.userAgent).toBeDefined();
+      });
+
+      const firstUserAgentRef = result.current.userAgent;
```
fredericbarthelet: same here -> `initialUserAgent` ?

packages/core/src/web/bridges/apps-sdk-adapter.ts
```diff
@@ -1,19 +1,58 @@
...
+    return window.openai.requestDisplayMode({ mode });
+  };
+
+  public sendFollowUpMessage = (prompt: string): Promise<void> => {
+    return window.openai.sendFollowUpMessage({ prompt });
+  };
+
+  public getMethod<K extends keyof Methods>(key: K) {
+    return this[key];
+  }
```
fredericbarthelet: I don't understand the need for this method

packages/core/src/web/bridges/adaptors/mcp-app-adaptor.ts
```diff
@@ -0,0 +1,197 @@
...
+        ) as PickContext<Keys>;
+        const newValue = computeSnapshot(context);
+
+        if (cachedValue !== undefined && dequal(cachedValue, newValue)) {
+          return cachedValue;
+        }
+
+        cachedValue = newValue;
+        return newValue;
+      },
```
fredericbarthelet, line 194: can't you just use `useCallback` from React on this monad in `getSnapshot` with a dependency to `context` to compute if a re-render is necessary?

packages/core/src/web/bridges/adaptors/mcp-app-adaptor.ts
```diff
@@ -0,0 +1,197 @@
...
+          keys.map((k) => [k, bridge.getSnapshot(k)]),
+        ) as PickContext<Keys>;
+        const newValue = computeSnapshot(context);
+
+        if (cachedValue !== undefined && dequal(cachedValue, newValue)) {
+          return cachedValue;
+        }
+
+        cachedValue = newValue;
+        return newValue;
```
fredericbarthelet, line 193:
```suggestion
        return cachedValue;
```

fredericbarthelet: I don't understand where you set cachedValue in fact, and why it's a let

fredericbarthelet (approved):

PR #176, opened 2026-01-05
Reenable devtools

harijoe:
### Goals

- Enabling publish of devtools
- Change the way the static server resolves the devtools html file (using node resolution instead of readFileSync)
- Fix biome issues in devtools

### Notes

- Could not declare @skybridge/devtools as a peer dependency of skybridge as it creates a circular dependency that doesn't resolve properly 
- It requires the user to keep in sync skybridge and @skybridge/devtools
- The CI script had to be changed to enable :workspace tags to be resolved. Now we set the versions of all the packages first, then we use pnpm publish which resolves workspace links before publishing
- Chicken&Egg problem : Removed the test on create-skybridge-template because it can not resolve devtools version (as it's not published), and to publish devtools the CI has to pass. Once a first version of the devtools is published, we'll be able to put the test back. Also removed temporarily the template from skybridge pnpm workspace.

<h3>Greptile Summary</h3>

Re-enabled publishing of `@skybridge/devtools` package and changed how devtools HTML files are resolved. The static server now uses Node's module resolution (`require.resolve`) instead of relative path resolution (`readFileSync`), which correctly locates the devtools package when installed as a dependency. Fixed biome formatting issues across devtools components (added `.js` extensions to imports and improved code formatting). Modified CI/CD to set versions for all packages before publishing (fixing workspace reference resolution) and added `--no-git-checks` flag to handle modified package.json files during publish. Removed the `create-skybridge-template` CI test temporarily to break the chicken-and-egg problem where the test can't resolve devtools version until it's published.

- This PR is safe to merge with minor considerations about the removed CI test
- The changes are well-structured and solve the circular dependency problem. The approach of using `require.resolve` is correct for locating installed packages. The CI workflow changes properly handle workspace dependencies. However, the removal of the `create-skybridge-template` test creates a gap in CI coverage that should be restored after the first publish.
- Ensure the `create-skybridge-template` test is re-enabled after first devtools publish completes

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -159,6 +159,61 @@ importers:
         specifier: ^2.1.9
         version: 2.1.9(@types/node@25.0.3)(@vitest/ui@2.1.9)(jsdom@25.0.1)(lightningcss@1.30.2)(msw@2.12.4(@types/node@25.0.3)(typescript@5.9.3))(terser@5.44.1)
 
+  packages/create-skybridge/template:
+    dependencies:
+      '@modelcontextprotocol/sdk':
+        specifier: ^1.24.3
+        version: 1.24.3(zod@4.1.13)
+      express:
+        specifier: ^5.1.0
+        version: 5.1.0
+      react:
+        specifier: ^19.1.1
+        version: 19.2.0
+      react-dom:
+        specifier: ^19.1.1
+        version: 19.2.0(react@19.2.0)
+      skybridge:
+        specifier: '>=0.16.0 <1.0.0'
+        version: 0.16.1(@modelcontextprotocol/sdk@1.24.3(zod@4.1.13))(@types/node@22.18.12)(@types/react@19.2.7)(jiti@2.6.1)(lightningcss@1.30.2)(react-dom@19.2.0(react@19.2.0))(react@19.2.0)(terser@5.44.1)(tsx@4.21.0)(use-sync-external-store@1.6.0(react@19.2.0))
+      vite:
+        specifier: ^7.1.11
+        version: 7.2.7(@types/node@22.18.12)(jiti@2.6.1)(lightningcss@1.30.2)(terser@5.44.1)(tsx@4.21.0)
+      zod:
+        specifier: ^4.1.13
+        version: 4.1.13
+    devDependencies:
+      '@modelcontextprotocol/inspector':
+        specifier: ^0.17.5
+        version: 0.17.5(@types/node@22.18.12)(@types/react-dom@19.2.3(@types/react@19.2.7))(@types/react@19.2.7)(typescript@5.9.3)
+      '@types/express':
+        specifier: ^5.0.3
+        version: 5.0.3
+      '@types/node':
+        specifier: ^22.15.30
+        version: 22.18.12
+      '@types/react':
+        specifier: ^19.1.16
+        version: 19.2.7
+      '@types/react-dom':
+        specifier: ^19.1.9
+        version: 19.2.3(@types/react@19.2.7)
+      '@vitejs/plugin-react':
+        specifier: ^5.0.4
+        version: 5.1.2(vite@7.2.7(@types/node@22.18.12)(jiti@2.6.1)(lightningcss@1.30.2)(terser@5.44.1)(tsx@4.21.0))
+      nodemon:
+        specifier: ^3.1.10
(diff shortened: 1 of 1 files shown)


packages/create-skybridge/template/pnpm-workspace.yaml
fredericbarthelet, line 1: There is only a single package.json in seed template now

packages/core/src/server/devtoolsStaticServer.ts
```diff
@@ -19,22 +24,33 @@ import express, { type RequestHandler } from "express";
...
 export const devtoolsStaticServer = async (): Promise<RequestHandler> => {
   const router = express.Router();
-  const currentDir = path.dirname(fileURLToPath(import.meta.url));
-  const devtoolsPath = path.join(currentDir, "..", "devtools");
+  let devtoolsPath: string;
+  try {
+    const devtoolsPackagePath = require.resolve(
+      "@skybridge/devtools/package.json",
+    );
+    devtoolsPath = path.join(path.dirname(devtoolsPackagePath), "dist");
```
fredericbarthelet, line 32: if this is exported by @skybridge/devtools, wouldn't it be easier :)?

.github/workflows/ci.yml
```diff
@@ -72,27 +72,3 @@ jobs:
 
       - name: Run ${{ matrix.name }}
         run: ${{ matrix.command }}
-
-  create-skybridge-template:
```
fredericbarthelet, line 76: why did you remove this test? The goal is to ensure the template is starting correctly :)

harijoe: It's a temporary thing : There's a chicken and egg problem (see the PR notes for details)

packages/core/src/server/devtoolsStaticServer.ts
harijoe: Not really easier actually. Devtools is a vite app so exporting an express middleware from it would require to add some deps and an extra build step. I think it's better to keep it in core for simplicity. WDYT?

.github/workflows/ci.yml
fredericbarthelet: If you don't add just yet the dependency to `@skybridge/devtools` in `server/src/package.json` and if you don't reference it in `server/src/index.ts` you should be fine

packages/core/src/server/devtoolsStaticServer.ts
fredericbarthelet: The error you're throwing just after makes me think we're in between 2 strategies: we can't reference @skybridge/devtools as a peer dep from core, but we're acting as if it was one. The 2 options I see best fit are:
- either ship everything in a single package core (that includes the devtools)
- either ship independent packages, but then @skybridge/devtools shouldn't be required in order for core to work (the direction we're going towards)

I might be missing something here, but AFAIK, packaging this from devtools "only" require building @skybridge/devtools like a normal app with a rollupOptions to give 2 inputs, one for the express router export and one for the app entrypoint. What am I missing here?

pnpm-workspace.yaml
```diff
@@ -1,4 +1,3 @@
 packages:
   - "docs"
   - "packages/*"
-  - "packages/create-skybridge/template"
```
fredericbarthelet, line 4: why do you remove the template from the monorepo too now?

.github/workflows/ci.yml
harijoe: This kinda defeats the name of this PR haha, but alright, enabling the emulator will go to another PR after this one has been released.

pnpm-workspace.yaml
harijoe: > why do you remove the template from the monorepo too now?

Exactly the same problem, moving it back now : the devtools will be added to the template in another PR

packages/core/src/server/devtoolsStaticServer.ts
harijoe: I tried but it doesn't simply work : vite bundles for the browser so bundling another input for node requires a different vite config (e.g. to externalize node built-ins).
Are we sure we want to go this way ? Maybe we can jump on a call to clarify this.

fredericbarthelet: OK to proceed with moving this to @skybridge/devtools in another PR

PR #183, opened 2026-01-05
feat: support user package manager

paulleseute:
Builds on #165 by detecting and using the user's preferred package manager for installation and execution.


packages/create-skybridge/src/index.ts
```diff
@@ -142,12 +140,15 @@ export async function init(args: string[] = process.argv.slice(2)) {
     process.exit(1);
   }
 
+  const userAgent = process.env.npm_config_user_agent;
```
qchuchu, line 143: Didn't know this env variable existed ! What sets it ?

paulleseute: The package manager I guess - I stole it from Vite

qchuchu: https://github.com/pnpm/pnpm/issues/3985 -> Ok, found this. Interesting to know :)

PR #193, opened 2026-01-05
feat: add modules import check in hot module reload to allow HMR on subfiles

### LINKED ISSUE #127: [Bug] HMR is not working when outside of widget entrypoint file
HMR works if editing widget.tsx, but doesn't work if editing a .tsx file imported by widget.tsx

qchuchu:
Fixes #127

<h3>Greptile Summary</h3>

Added HMR support for widget subfiles by implementing `handleHotUpdate` hook that traverses the module import graph to find and reload widget entry points when their dependencies change.

- Extracted widget pattern matching into reusable helper functions (`isWidgetEntryPoint`, `getWidgetName`)
- Added `collectWidgetEntryPoints` function that recursively walks up the module importer chain to find widget entry points
- Implemented `handleHotUpdate` hook to trigger HMR for both changed modules and their widget entry points
- Extended widget pattern matching to support both `.tsx` and `.jsx` extensions

- The implementation correctly handles HMR by traversing the module graph to find widget entry points. The logic includes proper cycle detection with a visited Set, handles edge cases like null file paths, and uses deduplication to avoid returning duplicate modules. The refactoring also improves code reusability by extracting widget detection logic into helper functions.

### DIFF
--- a/packages/core/src/test/widget.test.ts
+++ b/packages/core/src/test/widget.test.ts
@@ -112,7 +112,9 @@ describe("McpServer.registerWidget", () => {
     });
 
     // Check development-specific content
-    expect(result.contents[0]?.text).toContain(`${serverUrl}/@react-refresh`);
+    expect(result.contents[0]?.text).toContain(
+      `${serverUrl}/assets/@react-refresh`,
+    );
     expect(result.contents[0]?.text).toContain(`${serverUrl}/@vite/client`);
     expect(result.contents[0]?.text).toContain(
       `${serverUrl}/src/widgets/my-widget`,
@@ -167,7 +169,7 @@ describe("McpServer.registerWidget", () => {
 
     // Check production-specific content
     expect(result.contents[0]?.text).not.toContain(
-      `${serverUrl}@react-refresh`,
+      `${serverUrl}/assets/@react-refresh`,
     );
     expect(result.contents[0]?.text).not.toContain(`${serverUrl}@vite/client`);
     expect(result.contents[0]?.text).toContain(


packages/core/src/web/plugin/plugin.ts
```diff
@@ -1,6 +1,42 @@
...
+  if (!filePath) {
+    return false;
+  }
+  return FLAT_WIDGET_REGEX.test(filePath) || DIR_WIDGET_REGEX.test(filePath);
+};
+
+const getWidgetName = (filePath: string): string | undefined => {
+  return (
+    filePath.match(FLAT_WIDGET_REGEX)?.[1] ||
+    filePath.match(DIR_WIDGET_REGEX)?.[1]
```
fredericbarthelet: now that you do one **or** the other, I'd rather use a single regex that match both widget file entry and widget index entry: `/src\/widgets\/([^/]+)(?:\/index)?\.(tsx|jsx)$/`

packages/core/src/web/plugin/plugin.ts
```diff
@@ -17,15 +53,14 @@ export function skybridge(): Plugin {
       );
       const dirWidgetPattern = resolve(projectRoot, "src/widgets/*/index.tsx");
```
fredericbarthelet, line 18: you should use a single pattern as well -> `"src/widgets/*{.jsx,.tsx,/index.jsx,/index.tsx}"`

packages/core/src/web/plugin/plugin.ts
```diff
@@ -39,9 +74,20 @@ export function skybridge(): Plugin {
...
       return await dataLlmTransform(code, id);
     },
+
+    handleHotUpdate({ modules }) {
+      const visited = new Set<ModuleNode>();
+      const widgetEntryPoints = modules.flatMap((moduleNode) =>
+        collectWidgetEntryPoints(moduleNode, visited),
+      );
+
+      return [...new Set([...modules, ...widgetEntryPoints])];
```
fredericbarthelet: Maybe we don't need modules here as well

harijoe (approved):
Good catch Quentin! 🎯

PR #206, opened 2026-01-05
Actually enable devtools

harijoe:
Following : https://github.com/alpic-ai/skybridge/pull/176
Also includes opening the emulator on startup

<h3>Greptile Summary</h3>

This PR completes the devtools enablement by adding the `devtoolsStaticServer()` middleware and automatically opening the devtools UI on server startup in non-production environments.

- Added `@skybridge/devtools` package dependency to enable devtools UI
- Added `open` package to automatically launch browser
- Imported and mounted `devtoolsStaticServer()` middleware before `widgetsDevServer()`
- Added auto-open functionality that launches browser to `http://localhost:3000` on startup

The implementation follows the expected pattern from PR #176 and correctly guards both the middleware and browser opening with environment checks.

- This PR is safe to merge with one minor improvement needed for error handling
- The implementation is straightforward and correctly follows the expected pattern from the preparatory PR #176. All dependencies are properly added, and the environment checks are in place. The only issue is the unhandled promise from `open()` which could cause warnings in headless environments, but won't break functionality.
- The `index.ts` file needs attention for the unhandled promise in the `open()` call

### DIFF
--- a/packages/create-skybridge/template/package.json
+++ b/packages/create-skybridge/template/package.json
@@ -19,20 +19,19 @@
     "express": "^5.1.0",
     "react": "^19.1.1",
     "react-dom": "^19.1.1",
-    "skybridge": ">=0.16.0 <1.0.0",
+    "skybridge": ">=0.16.2 <1.0.0",
     "vite": "^7.1.11",
     "zod": "^4.1.13"
   },
   "devDependencies": {
     "@modelcontextprotocol/inspector": "^0.17.5",
-    "@skybridge/devtools": ">=0.16.0 <1.0.0",
+    "@skybridge/devtools": ">=0.16.2 <1.0.0",
     "@types/express": "^5.0.3",
     "@types/node": "^22.15.30",
     "@types/react": "^19.1.16",
     "@types/react-dom": "^19.1.9",
     "@vitejs/plugin-react": "^5.1.2",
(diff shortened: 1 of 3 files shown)


packages/create-skybridge/template/server/src/index.ts
```diff
@@ -27,6 +28,11 @@ app.listen(3000, (error) => {
   console.log(
     "Make your local server accessible with 'ngrok http 3000' and connect to ChatGPT with URL https://xxxxxx.ngrok-free.app/mcp",
   );
+
+  if (env !== "production") {
+    console.log("Opening devtools at http://localhost:3000");
+    open(`http://localhost:3000`);
```
paulleseute: IMO it'd less invasive to just have a clickable link then having a browser window open each time the server is started.

packages/create-skybridge/template/package.json
```diff
@@ -25,12 +25,14 @@
   },
   "devDependencies": {
     "@modelcontextprotocol/inspector": "^0.17.5",
+    "@skybridge/devtools": ">=0.16.0 <1.0.0",
```
paulleseute: we need at least 0.16.2

paulleseute (approved):
The @skybridge/devtools need to be upgraded, otherwise one non-blocking remark on the DX.

PR #224, opened 2026-01-06
Update every reference to nodejs 24

fredericbarthelet:
<h3>Greptile Summary</h3>

This PR attempts to update the codebase from Node.js 22 to Node.js 24, however **Node.js 24 has not been released yet**. The changes reference `lts/krypton` (a speculative codename) and `@types/node@24.10.1`, but Node.js 24 doesn't exist as a stable or LTS release.

**Critical Issues:**
- `.nvmrc` and CI workflow reference `lts/krypton` which is not a valid Node.js LTS version
- `@types/node@24.10.1` is installed but there's no corresponding Node.js 24 runtime
- This will cause CI failures and create mismatches between type definitions and actual runtime behavior
- Documentation updated to require "Node.js 24+" which users cannot install

**Current Status:**
- Node.js 22 (codename "Jod") is the latest stable version
- Node.js 22 will become LTS in October 2024
- Node.js 20 (codename "Iron") is the current Active LTS

The PR should either revert to Node.js 22 or wait until Node.js 24 is officially released.

<h3>Confidence Score: 0/5</h3>

- This PR is not safe to merge - it references a non-existent Node.js version
- The PR updates references to Node.js 24 which hasn't been released. The `lts/krypton` codename is not valid, CI will fail, and users cannot install the required Node.js version. This creates a critical mismatch between documentation, dependencies, and available software.
- All files need attention: `.nvmrc`, `.github/workflows/ci.yml`, `packages/create-skybridge/template/package.json`, and all documentation files must be corrected to reference an actual Node.js version (22 or 20)

### DIFF
--- a/packages/create-skybridge/template/package.json
+++ b/packages/create-skybridge/template/package.json
@@ -27,7 +27,6 @@
     "@modelcontextprotocol/inspector": "^0.18.0",
     "@skybridge/devtools": ">=0.16.2 <1.0.0",
     "@types/express": "^5.0.6",
-    "@types/node": "^22.19.3",
     "@types/react": "^19.2.7",
     "@types/react-dom": "^19.2.3",
     "@vitejs/plugin-react": "^5.1.2",
@@ -36,5 +35,8 @@
     "tsx": "^4.21.0",
     "typescript": "^5.9.3"
   },
-  "workspaces": []
+  "workspaces": [],
+  "engines": {
+    "node": ">=24.0.0"
+  }
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
 
       - uses: actions/setup-node@v6
         with:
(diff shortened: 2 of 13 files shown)


packages/create-skybridge/template/package.json
qchuchu, line 6: Should force the engine no then ?

packages/create-skybridge/template/package.json
```diff
@@ -36,5 +35,8 @@
     "tsx": "^4.21.0",
     "typescript": "^5.9.3"
   },
-  "workspaces": []
+  "workspaces": [],
+  "engines": {
```
qchuchu, line 39: Is it smth we can share at workspace level ? Or we still need to have it per example (I guess when copying the template) ? cc @paulleseute.

Otherwise I would set this engine at root level maybe ?

paulleseute: We need it in the templates since they are copied as standalone projects.

PR #227, opened 2026-01-06
[Example] Example real app capitals of the world

harijoe:
image

## Notes:
- Based on the create-skybridge template
- Unsplash removed in favor of image provided by wikipedia api
- Minor UI changes compared to [original repo](https://github.com/fredericbarthelet/the-context-demo)

<h3>Greptile Summary</h3>

Adds a comprehensive capitals of the world example application demonstrating Skybridge's widget capabilities including interactive maps, display mode switching, and external API integrations. The `.env` file with an Unsplash API key was properly removed (commit 722b07f). The implementation uses RestCountries API and Wikipedia for data, with a clean architecture separating server logic from UI components.

- Implements interactive map with Mapbox showing 200+ world capitals
- Features multi-panel UI with nearby capitals list and detailed capital information
- Uses proper caching strategy for RestCountries API calls  
- Includes comprehensive README with setup instructions
- Minor type improvement to `devtoolsStaticServer` return type for better TypeScript accuracy

- This PR is safe to merge with minimal risk - it adds a self-contained example application.
- The code is well-structured with proper error handling, caching mechanisms, and follows React best practices. The previously committed `.env` file with secrets was properly removed. The only concern is a hardcoded Mapbox token, but this is a public token used for map display and is commonly hardcoded in demo applications. The example is isolated to the `examples/capitals` directory and includes only one minor change to core code (TypeScript type improvement).
- No files require special attention - the example is self-contained and well-documented.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -12464,15 +12464,6 @@ snapshots:
     dependencies:
       react: 19.2.3
 
-  '@rjsf/core@6.1.2(@rjsf/utils@6.1.2(react@19.2.0))(react@19.2.0)':
-    dependencies:
-      '@rjsf/utils': 6.1.2(react@19.2.3)
-      lodash: 4.17.21
-      lodash-es: 4.17.22
-      markdown-to-jsx: 8.0.0(react@19.2.3)
-      prop-types: 15.8.1
-      react: 19.2.0
-
   '@rjsf/core@6.1.2(@rjsf/utils@6.1.2(react@19.2.3))(react@19.2.3)':
     dependencies:
       '@rjsf/utils': 6.1.2(react@19.2.3)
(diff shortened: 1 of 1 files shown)


examples/capitals/README.md
```diff
@@ -0,0 +1,80 @@
...
+- Node.js 22+
+- HTTP tunnel such as [ngrok](https://ngrok.com/download)
+
+### Local Development
+
+#### 1. Install
+
+```bash
+pnpm install
+```
```
paulleseute, line 42: This has changed, we now support all package manager - see template README

examples/capitals/web/src/widgets/capital.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { BrowserRouter, Route, Routes } from "react-router-dom";
+import { mountWidget, useDisplayMode } from "skybridge/web";
+import { Spinner } from "@/components/ui/shadcn-io/spinner/index.js";
+import { useCallTool, useToolInfo } from "../helpers.js";
+import { CapitalDetail } from "./components/CapitalDetail.js";
+import { MapView } from "./components/MapView.js";
+import { NearbyList } from "./components/NearbyList.js";
+
+// biome-ignore lint/correctness/useImportExtensions: biome doesn't support .css imports
+import "@/index.css";
```
paulleseute, line 11: when it's the first first import it does not complain

paulleseute (commented):
Maybe I'm missing something but I'm not able to render the widget in MCPJam or the dev tools

examples/capitals/README.md
harijoe: Even if we do, it felt very noisy in this example to give commands for every package manager, as the purpose of this readme is to introduce the specificities of this example I believe, let's discuss it!

examples/capitals/web/src/widgets/capital.tsx
harijoe: I tried but it still complains for some reason… I think we need to fix it on biome level : importing a .css is completely fine and should not force us to change imports order

examples/capitals/README.md
paulleseute: My understanding is those examples will end up as template at some point, so I guess it'd be better to have all package manager there as well. That said, it's your call.

examples/capitals/web/src/widgets/capital.tsx
qchuchu: I think we can specially remove this rule for the examples. 

It does not have any sense toe enforce the rules here - we should better only apply the rules for the packages we build

PR #228, opened 2026-01-06
feat: add ecom carousel to examples

paulleseute:
A minimal, no-dependency implementation of an ecommerce carousel. LLMs can filter product by category or price.

I'm planning to add the following to demonstrate more Skybridge capabilities:
- dark/light mode: useLayout
- traductions: useUser
- more info on click (instead on the description pane): sendFollowUpMessage
- basket: useWidgetState
- checkout: useOpenExternal

Or we can keep it simple and leave it as is.

### DIFF
--- a/examples/ecom-carousel/server/src/server.ts
+++ b/examples/ecom-carousel/server/src/server.ts
@@ -57,7 +57,7 @@ const server = new McpServer(
 
       return {
         structuredContent: { products: filtered },
-        content: [],
+        content: [{ type: "text", text: JSON.stringify(filtered) }],
         isError: false,
       };
     } catch (error) {
--- a/examples/ecom-carousel/web/src/widgets/ecom-carousel.tsx
+++ b/examples/ecom-carousel/web/src/widgets/ecom-carousel.tsx
@@ -2,14 +2,14 @@ import "@/index.css";
 
 import { useState } from "react";
 import { mountWidget } from "skybridge/web";
-import { useToolInfo } from "../helpers";
+import { useToolInfo } from "../helpers.js";
 
 function EcomCarousel() {
-  const { output } = useToolInfo<"ecom-carousel">();
+  const { output, isPending } = useToolInfo<"ecom-carousel">();
   type Product = NonNullable<typeof output>["products"][number];
   const [selected, setSelected] = useState<Product | null>(null);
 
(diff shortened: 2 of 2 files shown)


examples/ecom-carousel/web/src/widgets/ecom-carousel.tsx
```diff
@@ -0,0 +1,70 @@
+import "@/index.css";
+
+import { useState } from "react";
+import { mountWidget } from "skybridge/web";
+import { useToolInfo } from "../helpers";
+
+function EcomCarousel() {
+  const { output } = useToolInfo<"ecom-carousel">();
+  type Product = NonNullable<typeof output>["products"][number];
```
valentinbeggi, line 9: You can use

`type Product = ToolOutput<typeof server, "ecom-carousel">["products"][number];`

valentinbeggi: so you don't have to define type within component

examples/ecom-carousel/server/src/server.ts
```diff
@@ -0,0 +1,73 @@
...
+        }
+        if (maxPrice !== undefined && product.price > maxPrice) {
+          continue;
+        }
+        filtered.push(product);
+      }
+
+      return {
+        structuredContent: { products: filtered },
+        content: [],
```
valentinbeggi: from MCP spec:
> a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

examples/ecom-carousel/web/src/widgets/ecom-carousel.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const [selected, setSelected] = useState<Product | null>(null);
+
+  if (!output) {
```
valentinbeggi: toolInfo.isPending ?

valentinbeggi (approved):

examples/ecom-carousel/web/src/widgets/ecom-carousel.tsx
paulleseute: Yes but I then I'd have to import a bunch of other stuff - also this type is only used within the fn, so I think it's valid

PR #230, opened 2026-01-06
feat: add theme and language support to ecom example

paulleseute:
So we can showcase the useLayout() and useUser() hooks

### DIFF
--- a/examples/ecom-carousel/web/src/widgets/ecom-carousel.tsx
+++ b/examples/ecom-carousel/web/src/widgets/ecom-carousel.tsx
@@ -1,11 +1,37 @@
 import "@/index.css";
 
 import { useState } from "react";
-import { mountWidget, useLayout } from "skybridge/web";
+import { mountWidget, useLayout, useUser } from "skybridge/web";
 import { useToolInfo } from "../helpers.js";
+
+const translations: Record<string, Record<string, string>> = {
+  en: {
+    loading: "Loading products...",
+    noProducts: "No product found",
+  },
+  fr: {
+    loading: "Chargement des produits...",
+    noProducts: "Aucun produit trouvé",
+  },
+  es: {
+    loading: "Cargando productos...",
+    noProducts: "No se encontraron productos",
+  },
+  de: {
+    loading: "Produkte werden geladen...",
+    noProducts: "Keine Produkte gefunden",
+  },
+};
 
 function EcomCarousel() {
   const { theme } = useLayout();
+  const { locale } = useUser();
+
+  const lang = locale?.split("-")[0] ?? "en";
+
+  function translate(key: string) {
+    return translations[lang]?.[key] ?? translations.en[key];
+  }
 
   const { output, isPending } = useToolInfo<"ecom-carousel">();
   type Product = NonNullable<typeof output>["products"][number];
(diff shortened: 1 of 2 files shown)


examples/ecom-carousel/web/src/widgets/ecom-carousel.tsx
```diff
@@ -1,64 +1,103 @@
...
+    loading: "Produkte werden geladen...",
+    noProducts: "Keine Produkte gefunden",
+  },
+};
+
 function EcomCarousel() {
+  const { theme } = useLayout();
+  const { locale } = useUser();
+
+  const lang = locale?.split("-")[0] ?? "en";
```
qchuchu, line 30: Couldn't we be opinionated and use react-intl ? 

It would be cleaner no WDYT ? It can also show a great example to set the react intl provider with the locale.

paulleseute: We could, but IMO we should keep the examples as vanilla as possible, the goal being showcasing the skybridge hooks without code bits introduced by third-party dependencies. Some users will not have prior knowledge of `react-intl`, which, for them, would add noise.

examples/ecom-carousel/web/src/index.css
```diff
@@ -5,12 +5,21 @@
   margin: 1rem;
```
qchuchu, line 5: /* Light tokens (default) */
.container {
  display: flex;
  flex-direction: column;
  margin: 1rem;
  padding: 1rem;
  font-family: sans-serif;
  border-radius: 8px;

  --bg-elevated: #fff;
  --bg-muted: #f9f9f9;
  --text-primary: #333;
  --text-secondary: #555;
  --shadow-1: 0 2px 8px rgba(0, 0, 0, 0.1);
  --shadow-1-hover: 0 4px 12px rgba(0, 0, 0, 0.15);
  color-scheme: light;
}

/* Dark overrides */
.container[data-theme="dark"] {
  --bg-elevated: #2a2a2a;
  --bg-muted: #333;
  --text-primary: #eee;
  --text-secondary: #bbb;
  --shadow-1: 0 2px 8px rgba(0, 0, 0, 0.3);
  --shadow-1-hover: 0 4px 12px rgba(0, 0, 0, 0.4);
  color-scheme: dark;
}

Maybe we could do something like this and add variables for the colors ?

Don't need to put everything though but I like this 

And then you only need to set the data-theme attribute in the container from the layout

examples/ecom-carousel/web/src/widgets/ecom-carousel.tsx
```diff
@@ -1,64 +1,103 @@
...
       </div>
     );
   }
 
   const activeProduct = selected ?? output.products[0];
 
   return (
-    <div className="container">
-      <div className="carousel">
+    <div className={cn("container")}>
```
qchuchu, line 63: And then you wouldn't need the cn

qchuchu (changes requested):
A few comments - open to discussion as it's not "mandatory", but I like to tell that our examples show best practices of how to use these data ;)

examples/ecom-carousel/web/src/index.css
paulleseute: yes, I'm doing a css refactoring on the next PR which is pretty much what you just did.

examples/ecom-carousel/web/src/widgets/ecom-carousel.tsx
paulleseute: yes, it's also going away in the next pr.

PR #255, opened 2026-01-08
feat: set default values for CSP, domain and prefers border declarations

qchuchu:
<h2>Greptile Overview</h2>

### Greptile Summary

This PR successfully refactors widget resource metadata construction to generate defaults dynamically at runtime instead of at registration time. The implementation introduces a `buildContentMeta` function that generates CSP, domain, and `prefersBorder` defaults based on the server URL when resources are requested.

**Key improvements:**
- Default CSP domains automatically include the server URL and WebSocket URL
- Default domain is set to the server URL  
- Metadata generation adapts to the deployment environment at runtime
- Proper merge priority: defaults < `ui.*` fields < direct `openai/*` keys

**Implementation:**
- Uses `es-toolkit`'s `toMerged` for deep merging with correct priority order
- New test case validates the three-tier priority system
- Both apps-sdk and ext-apps formats are handled appropriately

### Confidence Score: 5/5

- This PR is safe to merge with no issues found
- The implementation correctly uses `toMerged` with proper priority order (defaults < ui.* < direct openai/*), comprehensive tests validate all edge cases including the priority system, and runtime defaults are properly generated. The refactoring improves the codebase by making metadata generation environment-aware without introducing any bugs.

### DIFF
--- a/packages/core/src/server/server.ts
+++ b/packages/core/src/server/server.ts
@@ -22,6 +22,7 @@ import type {
   ServerRequest,
   ToolAnnotations,
 } from "@modelcontextprotocol/sdk/types.js";
+import { toMerged } from "es-toolkit";
 import { templateHelper } from "./templateHelper.js";
 
 export type ToolDef<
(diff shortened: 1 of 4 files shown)


qchuchu:
@greptile review again

packages/core/src/server/server.ts
```diff
@@ -54,30 +57,35 @@ type OpenaiResourceMeta = {
...
+} & Resource["_meta"];
+
 export type WidgetHostType = "apps-sdk" | "mcp-app";
 
 type WidgetResourceConfig = {
   hostType: WidgetHostType;
   uri: string;
   mimeType: string;
+  /** Function to build _meta for this resource given the serverUrl */
+  buildMeta: (serverUrl: string, wsServerUrl: string) => ResourceMeta;
```
fredericbarthelet:
```suggestion
  buildMeta: (urls: {serverUrl: string, wsServerUrl: string}) => ResourceMeta;
```

packages/core/src/server/server.ts
```diff
@@ -176,37 +184,55 @@ export class McpServer<
...
+            wsServerUrl,
+          ],
+        },
+        "openai/widgetDomain": userMeta?.ui?.domain ?? serverUrl,
+        "openai/widgetDescription": toolConfig.description,
+        ...(userMeta?.ui?.prefersBorder !== undefined && {
+          "openai/widgetPrefersBorder": userMeta.ui.prefersBorder,
+        }),
+        ...userMeta,
+      }),
```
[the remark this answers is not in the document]
qchuchu: I want to currently give the possibility for our users to override the values as most users are building ChatGPTApps.

I don't want to introduce a silent breaking change for current users who already set the metas on their own.

packages/core/src/server/server.ts
```diff
@@ -176,37 +184,55 @@ export class McpServer<
...
       mimeType: "text/html+skybridge",
+      buildMeta: (serverUrl, wsServerUrl) => ({
+        "openai/widgetCSP": {
+          resource_domains: userMeta?.ui?.csp?.resourceDomains ?? [serverUrl],
+          connect_domains: userMeta?.ui?.csp?.connectDomains ?? [
+            serverUrl,
+            wsServerUrl,
+          ],
+        },
+        "openai/widgetDomain": userMeta?.ui?.domain ?? serverUrl,
```
fredericbarthelet: openai/widgetDomain is optional - openai builds one per widget by default. There is no need to provide a default value here
-> https://developers.openai.com/apps-sdk/reference#component-resource-_meta-fields

qchuchu: 👍 Then we could do the same as the prefersBorder

fredericbarthelet: however the following is shown on a custom app in the store

Image

So you can keep it

qchuchu: There is also the frame domain + redirect Domains that are optional that does not exist in the mcp apps spec but that we should support too. 

Maybe we can make our own supercharged type on this one, and only apply them in the openai case.

packages/core/src/server/server.ts
```diff
@@ -176,37 +184,55 @@ export class McpServer<
...
+        ...userMeta,
+      }),
     };
 
     const extAppsResourceConfig: WidgetResourceConfig = {
       hostType: "mcp-app",
       uri: `ui://widgets/ext-apps/${name}.html`,
       mimeType: "text/html;profile=mcp-app",
+      buildMeta: (serverUrl, wsServerUrl) => ({
+        ...userMeta,
```
fredericbarthelet: can you also put userMeta last to allow for manual override here ?

qchuchu: Actually if you take a look closely we do allow manual override on the ui resource below, but maybe we could outsource to a lib the merge objects with the default value, it'll be easier to read.

PR #270, opened 2026-01-09
feat: add all skybridge features to the everything app

paulleseute:
Follows https://github.com/alpic-ai/skybridge/pull/243. The everything app showcases the Skybridge framework capabilities. It's presented in a tabbed interface packed in a single widget. There's one tab per feature. Each tab is a mini playground to play with the API, with a link to the docs.

Code is minimal and vanilla by design, so users looking at it can focus on the actual Skybridge API and not stuff introduced by third party dependencies.

image

### DIFF
--- a/examples/everything/web/src/widgets/tabs/data-llm-tab.tsx
+++ b/examples/everything/web/src/widgets/tabs/data-llm-tab.tsx
@@ -7,15 +7,17 @@ export function DataLlmTab() {
   const [page, setPage] = useState<Page>("Home");
 
   return (
-    <div className="tab-content" data-llm={`User is viewing the ${page} page`}>
+    <div className="tab-content" data-llm={`User is viewing the $pagepage`}>
       <p className="description">
         The <code>data-llm</code> attribute syncs UI state with the model. As
         you navigate below, the LLM knows which page you're on.
       </p>
 
       <div className="field">
-        <span className="field-label">current context</span>
-        <code>User is viewing the {page} page</code>
+        <span className="field-label">compiled widgetState</span>
+        <pre>{`<DataLLM content="User is viewing the ${page} page}">
+  <div>{/* The page content */}</div>
+</DataLLM>`}</pre>
       </div>
 
       <div className="button-row">
@@ -23,7 +25,7 @@ export function DataLlmTab() {
           <button
             key={p}
             type="button"
-            className={`btn ${page === p ? "" : "btn-outline"}`}
+            className={`btn $page === p ? "" : "btn-outline"`}
             onClick={() => setPage(p)}
           >
             {p}
--- a/examples/everything/web/src/widgets/tabs/use-display-mode-tab.tsx
+++ b/examples/everything/web/src/widgets/tabs/use-display-mode-tab.tsx
@@ -37,7 +37,7 @@ export function UseDisplayModeTab() {
             className={`btn ${displayMode === mode ? "" : "btn-outline"}`}
             onClick={() => setDisplayMode(mode)}
           >
-            {mode} ({col} col)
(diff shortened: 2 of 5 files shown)


harijoe:
Could you add `allow-forms` to `skybridge/packages/devtools/src/components/layout/tool-panel/widget/widget.tsx` to make the useCallTool tab work? It looks like iframes prevents form submission by default

examples/everything/web/src/widgets/tabs/data-llm-tab.tsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <div className="tab-content" data-llm={`User is viewing the ${page} page`}>
+      <p className="description">
+        The <code>data-llm</code> attribute syncs UI state with the model. As
+        you navigate below, the LLM knows which page you're on.
+      </p>
+
+      <div className="field">
+        <span className="field-label">current context</span>
+        <code>User is viewing the {page} page</code>
```
harijoe: I think this is a bit misleading: To showcase what data-llm does I think we should show the compiled result that's being push to the widgetState instead.
Here the user can think that `User is viewing the {page} page` is a result of using data-llm but it's not true.

examples/everything/web/src/widgets/tabs/use-display-mode-tab.tsx
```diff
@@ -0,0 +1,46 @@
+import { type DisplayMode, useDisplayMode } from "skybridge/web";
+
+const ColByMode = new Map<DisplayMode, number>()
+  .set("inline", 1)
+  .set("pip", 2)
+  .set("fullscreen", 3);
+
+export function UseDisplayModeTab() {
```
harijoe, line 8: I think we could add some details on the different display mode available and when to use each one of them. I believe one of the everything app goal is to be kinda educational.

examples/everything/web/src/widgets/tabs/use-display-mode-tab.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+      <div className="button-row">
+        {Array.from(ColByMode.entries()).map(([mode, col]) => (
+          <button
+            key={mode.toString()}
+            type="button"
+            className={`btn ${displayMode === mode ? "" : "btn-outline"}`}
+            onClick={() => setDisplayMode(mode)}
+          >
+            {mode} ({col} col)
```
harijoe: Using "col" instead of columns doesn't look very clean. I think we should remove the number of columns as the main goal is to showcase the different display modes

examples/everything/web/src/widgets/tabs/use-files-tab.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const file = event.target.files?.[0];
+    if (!file) {
+      return;
+    }
+    setIsUploading(true);
+    try {
+      const { fileId } = await upload(file);
+      setFileId(fileId);
+    } catch (error) {
+      console.error("Upload failed:", error);
```
harijoe: I think we should also display the error to the use : I tried to upload a file that's not supported by chatgpt and the UI is the same as if the file upload would hang…

examples/everything/web/src/widgets/widget.tsx
```diff
@@ -26,8 +78,19 @@ function Widget() {
...
       </nav>
 
-      {tab === "Home" && <HomeTab />}
-      {tab === "useToolInfo" && <ToolInfoTab />}
+      <Component />
+
+      <div style={{ marginTop: "1.75rem", textAlign: "right" }}>
+        <button
+          type="button"
+          className="btn btn-muted btn-small"
```
harijoe: This button looks like it's disabled even though it's not.

PR #271, opened 2026-01-09
[CLI] Implement `skybridge build`

harijoe:
Kapture 2026-01-09 at 12 29 04

This PR needs to be published before changing the default build step in template and examples to `skybridge build`

### DIFF
--- a/packages/create-skybridge/template/server/src/index.ts
+++ b/packages/create-skybridge/template/server/src/index.ts
@@ -1,4 +1,4 @@
-import express, { type Express, type Router } from "express";
+import express, { type Express } from "express";
 import { widgetsDevServer } from "skybridge/server";
 import type { ViteDevServer } from "vite";
 import { mcp } from "./middleware.js";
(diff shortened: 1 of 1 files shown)


packages/create-skybridge/template/_gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules/
 dist/
 .env*
-.DS_store
\ No newline at end of file
+.DS_store
+*.tsbuildinfo
```
harijoe, line 5: This is to prepare `"build": "tsc -b && skybridge build"`

packages/core/src/cli/run-command.ts
fredericbarthelet, line 1: Why do you have to write this file? I feel like this is the scope of oclif

packages/core/src/cli/use-version.ts
fredericbarthelet, line 1: feels weird to remain with this file name if it doesn't export a hook anymore

PR #284, opened 2026-01-12
[CLI] skybridge start command

harijoe:
## Summary

This PR includes : 
- A new `skybridge start` command responsible to serve skybridge built files
- A change in the express middleware configuration to serve assets when started

<h2>Greptile Overview</h2>

### Greptile Summary

This PR adds a new `skybridge start` command to serve production builds and updates the template to use it. The command spawns a node process to run the built `dist/index.js` file, while the server configuration adds static asset serving for production environments.

**Key changes:**
- New command file `packages/core/src/commands/start.tsx` that spawns the production server
- Template's `start` script now uses `skybridge start` instead of directly running `node dist/index.js`
- Production server configuration added to serve static assets from `/assets` directory with CORS support
- Added `cors` dependency to the template

**Critical issue found:** The command class is incorrectly named `Dev` instead of `Start`, which will prevent the `skybridge start` command from working properly in oclif.

### Confidence Score: 1/5

- Not safe to merge - contains a blocking bug that will prevent the main feature from working
- The class name error in start.tsx is a critical bug that will prevent the `skybridge start` command from being registered correctly in oclif. The file is named `start.tsx` but exports a class named `Dev`, which means oclif will try to register this as the `dev` command instead of `start`, causing conflicts and breaking the intended functionality.
- packages/core/src/commands/start.tsx requires immediate attention to fix the class name

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,10 +88,13 @@ jobs:
           node-version-file: ".nvmrc"
           cache: "pnpm"
 
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
       - name: Install template dependencies with workspace:*
         run: |
           cd packages/create-skybridge/template
-          pnpm install skybridge@workspace:* @skybridge/devtools@workspace:* --no-frozen-lockfile
+          pnpm add skybridge@workspace:* @skybridge/devtools@workspace:*
 
       - name: Build core and devtools
         run: |
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -461,7 +461,7 @@ importers:
         version: 1.25.2(hono@4.11.3)(zod@4.3.5)
       '@rjsf/core':
         specifier: ^6.1.2
-        version: 6.1.2(@rjsf/utils@6.1.2(react@19.2.0))(react@19.2.3)
+        version: 6.1.2(@rjsf/utils@6.1.2(react@19.2.3))(react@19.2.3)
       '@rjsf/shadcn':
         specifier: ^6.1.2
         version: 6.1.2(@rjsf/core@6.1.2(@rjsf/utils@6.1.2(react@19.2.3))(react@19.2.3))(@rjsf/utils@6.1.2(react@19.2.3))(@types/react-dom@19.2.3(@types/react@19.2.7))(@types/react@19.2.7)(react-dom@19.2.3(react@19.2.3))(react@19.2.3)(tailwindcss@4.1.18)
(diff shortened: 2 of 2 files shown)


packages/core/src/commands/start.tsx
```diff
@@ -0,0 +1,63 @@
...
+      console.error("  skybridge build");
+      console.error("");
+      process.exit(1);
+    }
+
+    const App = () => {
+      const version = useVersion();
+
+      useEffect(() => {
+        const nodeProcess = spawn("node", ["dist/index.js"], {
```
fredericbarthelet: why not use your new `runCommand` util ?

packages/core/src/commands/start.tsx
```diff
@@ -0,0 +1,63 @@
...
+      console.error("❌ Error: dist/index.js not found");
+      console.error("");
+      console.error("Please build your project first:");
+      console.error("  skybridge build");
+      console.error("");
+      process.exit(1);
+    }
+
+    const App = () => {
+      const version = useVersion();
```
fredericbarthelet: use `this.config.version`

packages/core/src/commands/start.tsx
```diff
@@ -0,0 +1,63 @@
...
+      }, []);
+
+      return (
+        <Box flexDirection="column" padding={1} marginLeft={1}>
+          <Box marginBottom={1}>
+            <Text color="cyan" bold>
+              ⛰{"  "}Welcome to Skybridge
+            </Text>
+            <Text color="cyan"> v{version}</Text>
+          </Box>
```
fredericbarthelet: Maybe you can factor in a dedicated `Header` component ? It looks like it's there on all 3 commands

.github/workflows/ci.yml
```diff
@@ -91,11 +88,24 @@ jobs:
           node-version-file: ".nvmrc"
           cache: "pnpm"
 
-      - name: Install dependencies
-        run: pnpm install
+      - name: Replace template dependencies with workspace:*
```
fredericbarthelet: Can't you just run pnpm install skybridge@workspace:* ?

packages/core/src/cli/run-command.ts
```diff
@@ -1,12 +1,18 @@
-import { spawn } from "node:child_process";
+import { type SpawnOptions, spawn } from "node:child_process";
 
-export function runCommand(command: string, args: string[]): Promise<void> {
+export function runCommand(
+  command: string,
+  options: SpawnOptions = {
+    stdio: ["ignore", "pipe", "pipe"],
```
fredericbarthelet: Why do you need to change this depending on the command ? Is it when command is nested in react use effect or not ?

fredericbarthelet (approved):
A few comments but good to go otherwise :)

PR #285, opened 2026-01-12
[CLI] Start tunneling service in skybridge dev command

harijoe:
image

## Summary

[localtunnel](https://theboroer.github.io/localtunnel-www/) is now used to expose the local server and thus allow the user to test his app to chatgpt without having to start another command such as ngrok. It supports HMR.

<h3>Greptile Summary</h3>

This PR adds support for tunneling development servers by introducing a `--use-forwarded-host` flag to the `skybridge dev` command, enabling developers to use services like ngrok or localtunnel to expose local servers without manual configuration.

## Changes Summary

- **Command line parsing (`bin/run.js`)**: Fixed flag handling by changing from `push()` to `splice(2, 0, "dev")` to properly insert the `dev` command before flags, allowing `skybridge --flag` to work correctly
- **Dev command (`src/commands/dev.tsx`)**: Added `--use-forwarded-host` boolean flag that sets the `SKYBRIDGE_USE_FORWARDED_HOST` environment variable when enabled
- **Server URL construction (`src/server/server.ts`)**: Modified to use forwarded host headers when either in production or when the flag is enabled

## Note on PR Description
The PR description claims "localtunnel is now used to expose the local server," but the implementation only provides infrastructure support for any tunneling solution. No localtunnel-specific code is present—this is framework-level support that users can combine with their choice of tunneling tool.

## Identified Issues
- Missing fallback for malformed URLs when forwarded headers are unavailable

- This PR has reasonable changes for enabling tunneled development, but contains a potential issue with missing URL fallback that could cause runtime failures in misconfigured setups.
- The core functionality works correctly—the flag handling fix is sound and the forwarded host logic properly supports tunneling scenarios. However, the code doesn't gracefully handle the case where the forwarded headers are missing but the flag is enabled, which could produce invalid URLs. This is not a show-stopper since the original code has the same pattern, but it represents a defensive coding gap that should be addressed. Score is 3 because the issue is real but only affects edge cases of misconfiguration.
- packages/core/src/server/server.ts - needs defensive fallback for missing forwarded headers


packages/core/src/commands/dev.tsx
```diff
@@ -1,16 +1,31 @@
...
+      description:
+        "Uses the forwarded host header to construct widget URLs instead of localhost, useful when accessing the dev server through a tunnel (e.g., ngrok)",
+    }),
+  };
 
   public async run(): Promise<void> {
+    const { flags } = await this.parse(Dev);
+
+    const env = {
+      ...process.env,
```
fredericbarthelet: this is now passing all process env in nodemon, are you sure this is intended behavior?

packages/core/src/server/server.ts
```diff
@@ -391,10 +391,13 @@ export class McpServer<
...
+        const useForwardedHost =
+          process.env.SKYBRIDGE_USE_FORWARDED_HOST === "true";
 
-        const serverUrl = isProduction
-          ? `https://${extra?.requestInfo?.headers?.["x-forwarded-host"] ?? extra?.requestInfo?.headers?.host}`
-          : "http://localhost:3000";
+        const serverUrl =
+          isProduction || useForwardedHost
+            ? `https://${extra?.requestInfo?.headers?.["x-forwarded-host"] ?? extra?.requestInfo?.headers?.host}`
+            : "http://localhost:3000";
```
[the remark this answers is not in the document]
fredericbarthelet: Can you really end up in a situation where a request both lacks host and x-forwarded-host headers?

packages/core/bin/run.js
```diff
@@ -2,8 +2,11 @@
 
 import { execute } from "@oclif/core";
 
-if (process.argv.length === 2) {
-  process.argv.push("dev");
+if (
+  process.argv.length === 2 ||
+  (process.argv.length >= 3 && process.argv[2]?.startsWith("-"))
+) {
+  process.argv.splice(2, 0, "dev");
```
fredericbarthelet: as discussed, please revert to original `run.js` oclif implementation. Default command should be handled with oclif configuration `default` - https://oclif.io/docs/config

fredericbarthelet (approved):

PR #293, opened 2026-01-13
feat: add hosts to registerWidget resourceConfig

paulleseute:
So users can restrict the host types to a given subset. Useful for widgets using APIs not supported by all runtimes.

### DIFF
--- a/packages/core/src/test/widget.test.ts
+++ b/packages/core/src/test/widget.test.ts
@@ -424,4 +424,45 @@ describe("McpServer.registerWidget", () => {
     // Description should be from defaults (toolConfig.description)
     expect(meta["openai/widgetDescription"]).toBe("Test tool");
   });
+
+  it("should register tool with ui.resourceUri metadata only", async () => {
+    const mockToolCallback = vi.fn();
+    server.registerWidget(
+      "my-widget",
+      { description: "Test widget", hosts: ["mcp-app"] },
+      { description: "Test tool" },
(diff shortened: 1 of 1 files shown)


packages/core/src/server/server.ts
qchuchu, line 291: You also need to not expose the corresponding resource here :)

paulleseute: Indeed: [7ce5185](https://github.com/alpic-ai/skybridge/pull/293/commits/7ce51856fe64fc50f9c7ec7e53a92a7dfe4f0c4a)

packages/core/src/server/server.ts
```diff
@@ -231,88 +236,92 @@ export class McpServer<
...
-
-    const toolMeta: ToolMeta = {
-      ...toolConfig._meta,
-      "openai/outputTemplate": appsSdkResourceConfig.uri,
-      ui: {
-        resourceUri: extAppsResourceConfig.uri,
-      },
-    };
+      toolMeta.ui = { resourceUri: widgetConfig.uri };
+    }
```
qchuchu, line 324: Could you maybe add a test on this behavior in the widget.test.ts ? :)

paulleseute: [d0fdafb](https://github.com/alpic-ai/skybridge/pull/293/commits/d0fdafb569fac799ebaabddfd01b4ac10dc9690a)

PR #298, opened 2026-01-13
feat: restrict example app hosts to apps-sdk only

paulleseute:
Because they are using APIs not supported by the MCP apps UI variant, which are loaded by default by MCP Jam tools.

<h2>Greptile Overview</h2>

### Greptile Summary

Updated three example apps (capitals, ecom-carousel, everything) to restrict them to `apps-sdk` host only by adding `hosts: ["apps-sdk"]` to their widget configuration and bumping the minimum `skybridge` version from 0.20.0 to 0.22.0.

- The `hosts` configuration was introduced in skybridge v0.22.0 (commit 3fce400)
- These example apps use OpenAI apps-sdk specific APIs (`useRequestModal`, `useFiles`, `useSendFollowUpMessage`, `useLayout`, etc.) that are not available in the MCP apps UI variant
- The create-skybridge template was correctly left unrestricted since it only uses basic APIs compatible with both host types

### Confidence Score: 5/5

- The changes are straightforward and well-targeted: (1) Only example apps using apps-sdk specific APIs are restricted, (2) The version bump to 0.22.0 correctly corresponds to the release that introduced the hosts feature, (3) The create-skybridge template was intentionally and correctly left unrestricted since it uses only basic APIs, (4) All changes are consistent across the affected examples, (5) The pnpm-lock.yaml updates are automatic and expected


examples/capitals/server/src/server.ts
```diff
@@ -39,6 +39,7 @@ const server = new McpServer(
         },
       },
     },
+    hosts: ["apps-sdk"],
```
qchuchu, line 42: This one it should also work on mcp apps. But I'll dig into it

qchuchu (approved):

PR #316, opened 2026-01-15
feat: add use useRequestModal polyfill

paulleseute:
useRequestModal only worked on apps-sdk because it called window.openai.requestModal() directly. MCP App doesn't have native modal support, so we implemented our own.

How it works:
  - McpAppAdaptor.modalStore manages the modal state and provides subscribe, getSnapshot, open, close - all stable references so useSyncExternalStore doesn't re-subscribe on every render
  - useRequestModal() subscribes to the store and returns { isOpen, params, open }
  - ModalProvider wraps the widget, subscribes to the same store, and renders the actual modal UI when isOpen is true
  - clicking backdrop or pressing escape closes the modal
  - children stay mounted across open/close so react state is preserved

I also reworked the everything app to make use of the RequestModalOptions. I noted that the "title" and "anchor" options do not seem to have any effect on the OpenAI runtime, so I did not implement anything on the MCP apps variant as well. I'll double check and get rid of those params.

Screenshot 2026-01-15 at 16 50 08

### DIFF
--- a/packages/core/src/web/mount-widget.ts
+++ b/packages/core/src/web/mount-widget.ts
@@ -2,7 +2,6 @@
 
 import { createElement, StrictMode } from "react";
 import { createRoot, type Root } from "react-dom/client";
-import { ModalProvider } from "./components/modal-provider.js";
 import { installOpenAILoggingProxy } from "./proxy.js";
 
 let rootInstance: Root | null = null;
@@ -22,10 +21,13 @@ export const mountWidget = (component: React.ReactNode) => {
   }
 
   const hostType = window.skybridge?.hostType;
-  const app =
-    hostType === "mcp-app"
-      ? createElement(ModalProvider, null, component)
-      : component;
 
-  rootInstance.render(createElement(StrictMode, null, app));
+  (async () => {
+    let app = component;
+    if (hostType === "mcp-app") {
+      const { ModalProvider } = await import("./components/modal-provider.js");
+      app = createElement(ModalProvider, null, component);
+    }
+    rootInstance.render(createElement(StrictMode, null, app));
+  })();
 };


valentinbeggi:
If you don't mind, can you also update documentation:

useRequestModal is tagged "chatgpt only" at several place, you can update

packages/core/src/web/bridges/mcp-app/adaptor.ts
```diff
@@ -35,6 +37,31 @@ export class McpAppAdaptor implements Adaptor {
...
+    subscribe: (onChange: () => void) => {
+      this.modalStore.listeners.add(onChange);
+      return () => this.modalStore.listeners.delete(onChange);
+    },
+    getSnapshot: () => this._modalState,
+    open: (options: RequestModalOptions) => {
+      this._modalState = { isOpen: true, options };
+      for (const listener of this.modalStore.listeners) {
+        listener();
+      }
```
valentinbeggi: .forEach, to stay consistent with the rest of the file

packages/core/src/web/bridges/mcp-app/adaptor.ts
```diff
@@ -35,6 +37,31 @@ export class McpAppAdaptor implements Adaptor {
...
+    },
+    close: () => {
+      this._modalState = { isOpen: false, options: {} };
+      for (const listener of this.modalStore.listeners) {
+        listener();
```
valentinbeggi: same ?

packages/core/src/web/components/modal-provider.tsx
```diff
@@ -0,0 +1,63 @@
...
+  border-radius: 12px;
+  position: fixed;
+  inset: 0;
+  margin: auto;
+  width: fit-content;
+  height: fit-content;
+  background: white;
+  z-index: 9999;
+}
+`;
```
valentinbeggi, line 21: to what extent could these styles be overridable by user
maybe passing a config to mountWidget ? 

not for this PR but for further improvements I guess

packages/core/src/web/mount-widget.ts
```diff
@@ -20,5 +21,11 @@ export const mountWidget = (component: React.ReactNode) => {
     installOpenAILoggingProxy();
   }
 
-  rootInstance.render(createElement(StrictMode, null, component));
+  const hostType = window.skybridge?.hostType;
+  const app =
+    hostType === "mcp-app"
+      ? createElement(ModalProvider, null, component)
+      : component;
```
valentinbeggi: don't we want to make a conditional import of ModalProvider here ?

packages/core/src/web/bridges/mcp-app/adaptor.ts
```diff
@@ -216,6 +243,19 @@ export class McpAppAdaptor implements Adaptor {
...
+  public useRequestModal() {
+    const state = useSyncExternalStore(
+      this.modalStore.subscribe,
+      this.modalStore.getSnapshot,
+    );
+
+    return {
+      isOpen: state.isOpen,
+      params: state.options.params,
+      open: this.modalStore.open,
```
valentinbeggi: since we control mcp apps modal but not chatgpt's, can we maybe add an optional close param so that users can at least call close on mcp apps ?

valentinbeggi: Or maybe if changing display mode actually triggers a modal close, we could write a close method on chatgpt's adaptor

  const close = () => {
    window.openai.setDisplayMode?.("inline");
  };

valentinbeggi: but that's more opinionated haha

packages/core/src/web/bridges/apps-sdk/adaptor.ts
```diff
@@ -66,4 +68,19 @@ export class AppsSdkAdaptor implements Adaptor {
 
     return window.openai.setWidgetState(newState);
   };
+
+  public useRequestModal() {
```
valentinbeggi: i find it weird to find react code in the adaptor ?

This mixes React hook code into adaptor classes. The codebase already has useHostContext for consuming host state, I think the modal state should follow the same pattern ? wdyt ?

I think we could add modal to HostContext, implement getHostContextStore("modal") in each adaptor, and expose modalActions for the open/close methods.

goal is to have in the end, liek the other hooks:
```ts
export function useRequestModal() {
  const { isOpen, params } = useHostContext("modal");
  const { open, close } = getAdaptor().modalActions;

  return { isOpen, params, open, close };
}
```

This keeps React code in hooks, adaptors as plain state/method containers, and stays consistent with the existing patterns.

wdyt?

packages/core/src/web/components/modal-provider.tsx
valentinbeggi: maybe create an issue if we decide to do so

paulleseute: They can do whatever they want inside the modal container, but not styling the actual modal container or backdrop, same as for the openai variant.

packages/core/src/web/bridges/mcp-app/adaptor.ts
paulleseute: Honestly I'd not add programmatic method to close the modal since OpenAI did not provide any. The goal here is to have a somewhat feature parity with OpenAI, and mimick what they do until the MCP ext spec eventually comes up with something the actually deals with modal.

packages/core/src/web/bridges/apps-sdk/adaptor.ts
paulleseute: Good point:
- The "view" is now exposed by the host context - this is basically what window.openai exposes.
- For MCP Apps, the view lives in a store.
- Adaptor exposes an openModal method
- The MCP Adaptor exposes a close method, which is used to mimick the ChatGPT App behavior in the modal provider

packages/core/src/web/mount-widget.ts
paulleseute: [c0de06d](https://github.com/alpic-ai/skybridge/pull/316/commits/c0de06d941a3f494e1375191a35b58005550ad89)

PR #317, opened 2026-01-15
feat: stub useFiles

paulleseute:
So it does not break mcp apps

### DIFF
--- a/packages/core/src/web/bridges/apps-sdk/adaptor.ts
+++ b/packages/core/src/web/bridges/apps-sdk/adaptor.ts
@@ -67,10 +67,11 @@ export class AppsSdkAdaptor implements Adaptor {
     return window.openai.setWidgetState(newState);
   };
 
-  public useFiles() {
-    return {
-      upload: window.openai.uploadFile,
-      getDownloadUrl: window.openai.getFileDownloadUrl,
-    };
-  }
+  public uploadFile = (file: File) => {
+    return window.openai.uploadFile(file);
+  };
+
+  public getFileDownloadUrl = (file: { fileId: string }) => {
+    return window.openai.getFileDownloadUrl(file);
+  };
 }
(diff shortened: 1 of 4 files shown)


packages/core/src/web/bridges/apps-sdk/adaptor.ts
```diff
@@ -66,4 +66,11 @@ export class AppsSdkAdaptor implements Adaptor {
...
     return window.openai.setWidgetState(newState);
   };
+
+  public useFiles() {
+    return {
+      upload: window.openai.uploadFile,
+      getDownloadUrl: window.openai.getFileDownloadUrl,
+    };
+  }
 }
```
valentinbeggi: same as previous PR, I find it cleaner not to introduce any React specific naming inside,

this can be named fileActions maybe ?

packages/core/src/web/bridges/mcp-app/adaptor.ts
```diff
@@ -216,6 +216,17 @@ export class McpAppAdaptor implements Adaptor {
...
+  public useFiles() {
+    return {
+      upload: () => {
+        throw new Error("File upload is not supported in MCP App.");
+      },
+      getDownloadUrl: () => {
+        throw new Error("File download is not supported in MCP App.");
+      },
+    };
+  }
```
valentinbeggi: maybe we can add jsdoc so that on hover, users can see that it is not implemented in mcp apps ?

packages/core/src/web/bridges/apps-sdk/adaptor.ts
paulleseute: right - since adaptor methods are verb i split those useFiles into uploadFile and getFileDownloadUrl

valentinbeggi (approved):

PR #322, opened 2026-01-16
Add telemetry on skybridge CLI

fredericbarthelet:
Introduces a new `~/.skybridge/config.json` file with machineId and telemetry settings
Adds 3 new commands: `skybridge telemetry status`, `skybridge telemetry enable`, `skybridge telemetry disable`
In addition to disabling telemetry in the skybridge config file, the env variables `DO_NOT_TRACK` and `SKYBRIDGE_TELEMETRY_DISABLED` set to 1 or true disable the tracking.
Each `skybridge` command is send back with telemetry (outcome, error message, node version, os....)

<h3>Greptile Summary</h3>

Adds comprehensive telemetry tracking to the Skybridge CLI with PostHog integration. Creates new `~/.skybridge/config.json` for storing machine ID and telemetry preferences, adds three management commands (`status`, `enable`, `disable`), and implements a "finally" hook that captures command execution events including outcome, errors, Node version, and OS details.

- New telemetry commands provide clear opt-out mechanisms via CLI or environment variables (`SKYBRIDGE_TELEMETRY_DISABLED`, `DO_NOT_TRACK`)
- Telemetry is enabled by default but respects multiple opt-out methods
- CI environments have telemetry enabled by default (unless explicitly disabled via env vars)
- PostHog client singleton has a bug where it's not reset after shutdown, which could affect programmatic usage or tests

- Safe to merge with minor fixes needed for edge cases
- The implementation is solid with good error handling and privacy controls, but has a client lifecycle bug that could cause silent failures in non-standard usage patterns (tests, programmatic usage). The bug is caught by try-catch so it won't crash the CLI, but telemetry won't work correctly for subsequent commands in the same process.
- Pay close attention to `packages/core/src/hooks/telemetry.ts` for the PostHog client lifecycle fix

### DIFF
--- a/docs/telemetry.mdx
+++ b/docs/telemetry.mdx
@@ -134,7 +134,3 @@ This will print the telemetry event to stderr instead of sending it, useful for 
 - Data is stored on PostHog's US servers
 - We retain aggregated data for product analytics purposes
 - Individual events are anonymized and cannot be linked to specific users
-
-<Card title="Telemetry CLI Reference" icon="terminal" href="/api-reference/cli-telemetry">
-  View the complete CLI reference for telemetry commands
-</Card>


packages/core/src/cli/telemetry.ts
```diff
@@ -0,0 +1,176 @@
...
+function getPostHogClient(): PostHog {
+  if (!posthogClient) {
+    posthogClient = new PostHog(POSTHOG_API_KEY, {
+      host: POSTHOG_HOST,
+      flushAt: 1,
+      flushInterval: 0,
+    });
+  }
+
+  return posthogClient;
```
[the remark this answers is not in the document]
fredericbarthelet: removed the shutdown part

packages/core/src/commands/telemetry/disable.tsx
```diff
@@ -0,0 +1,35 @@
+import { Command } from "@oclif/core";
+import { Box, render, Text } from "ink";
+import { setEnabled } from "../../hooks/telemetry.js";
+
+export default class TelemetryDisable extends Command {
+  static override description = "Disable Skybridge telemetry";
```
harijoe: It was not clear for me at first if `skybridge telemetry disable` was project-wide or machine-wide (I wrongly assumed it was project-wise). Maybe add this info to the command description?

packages/core/src/commands/telemetry/disable.tsx
```diff
@@ -0,0 +1,35 @@
+import { Command } from "@oclif/core";
+import { Box, render, Text } from "ink";
+import { setEnabled } from "../../hooks/telemetry.js";
```
harijoe: The CLI folder goal was to contain all helper files related to the CLI, I suggest to move the telemetry.js inside.
Also telemetry.js is not providing a hook at the moment, just a regular function so I find it misleading to put it in a "hooks" folder. Usually only files starting as useSomething are expected in a hooks folder.

packages/core/src/commands/telemetry/disable.tsx
```diff
@@ -0,0 +1,35 @@
...
+          <Text color="yellow" bold>
+            disabled
+          </Text>
+        </Box>
+        <Box marginTop={1}>
+          <Text color="gray">Config saved to ~/.skybridge/config.json</Text>
+        </Box>
+        <Box marginTop={1}>
+          <Text color="gray">
+            You can also set SKYBRIDGE_TELEMETRY_DISABLED=1 to disable telemetry
```
harijoe: This is not very useful : giving the user another way to disable telemetry when he just disabled it system-wide. I'd suggest instead showing message telling the user how to enable it again and maybe reassure him that enabling telemetry does not imply security threats.

packages/core/src/commands/telemetry/disable.tsx
harijoe: The root of the confusion is the name "hook" used by react and oclif to name two very different things

harijoe (approved):

packages/core/src/commands/telemetry/disable.tsx
fredericbarthelet: Indeed, it's an oclif hook here.
I moved the file to the `cli` folder to avoid confusion

Issue #301, opened 2026-01-13
Example app mixed OAuth


fredericbarthelet:
Here's the example recently added to OpenAI Example apps: https://github.com/openai/openai-apps-sdk-examples/tree/main/authenticated_server_python 

it has mixed auth

PR #325, opened 2026-01-16
doc/mixed auth example

valentinbeggi:
Mixed auth example

Note that it doesn't work very well in chatGPT. chatGPT should show UI for login in case the user is unauthenticated, except that they require that the tools return a `securitySchemes` key, which is not yet in the MCP spec nor in TS sdk

https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1488

#301

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,9 @@
 {
   "editor.defaultFormatter": "biomejs.biome",
+  "editor.formatOnSave": true,
+  "editor.codeActionsOnSave": {
+    "source.organizeImports.biome": "explicit",
+    "source.fixAll.biome": "explicit"
+  },
   "biome.lsp.bin": "${workspaceFolder}/node_modules/.bin/biome"
 }
--- a/examples/mixed-auth/package.json
+++ b/examples/mixed-auth/package.json
@@ -35,6 +35,7 @@
     "@types/react-dom": "^19.2.3",
     "@vitejs/plugin-react": "^5.1.2",
     "nodemon": "^3.1.11",
+    "shx":"^0.4.0",
     "tsx": "^4.21.0",
     "typescript": "^5.9.3"
   },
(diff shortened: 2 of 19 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1,9 @@
 {
   "editor.defaultFormatter": "biomejs.biome",
+  "editor.formatOnSave": true,
+  "editor.codeActionsOnSave": {
+    "source.organizeImports.biome": "explicit",
+    "source.fixAll.biome": "explicit"
```
fredericbarthelet, line 6: why do you need to manually set those 2 ?

examples/mixed-auth/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "type": "module",
+  "scripts": {
+    "dev": "skybridge",
+    "build": "skybridge build",
+    "start": "skybridge start",
+    "inspector": "mcp-inspector http://localhost:3000/mcp",
+    "server:build": "tsc -p tsconfig.server.json",
+    "server:start": "node dist/index.js",
+    "web:build": "tsc -b web && vite build -c web/vite.config.ts",
+    "web:preview": "vite preview -c web/vite.config.ts"
```
fredericbarthelet, line 15: those are legacy, you can remove

qchuchu:
@valentinbeggi @neil-ac something is blocked here ? :)

paulleseute:
@valentinbeggi should we revive this ? mixed auth is supported by openai

neil-ac:
Apparently it works as well in Claude (haven't tested myself): https://claude.com/docs/connectors/building/lazy-authentication

-> +1 on this! I will need this for the V2s of some MCP App projects. V1 is often focused on unauthenticated product discovery, and V2 for authenticated user flows, could be cool to handle both with a single App

PR #348, opened 2026-01-20
feat: add creating-chatgpt-app skill

paulleseute:
This PR adds the first block of the ChatGPT app creation skill, which is having the LLM capable of brainstorming with users and refine their ideas so they make sense to run within ChatGPT.

This can be tested with a LLM before/after loading the actual skills. Alternatively, you can run the evaluation set with Claude.

### DIFF
--- a/skills/chatgpt-app-builder/references/brainstorm.md
+++ b/skills/chatgpt-app-builder/references/brainstorm.md
@@ -1,85 +1,56 @@
 # Brainstorming Workflow
 
-Help users validate and refine their ChatGPT app idea before implementation. Be brief - only go deep when there are issues to address.
-Brainstorming is about validating the idea, not planning implementation. Do not define tools, ask for API schemas, or discuss technical details. Keep it high level.
+Validate whether a ChatGPT app idea is a good fit. Stay high level—focus on the problem, user, and core action. Do not
+discuss implementation details (tools, APIs, data schemas, field mappings).
+
+If you have enough context to validate, move forward. Don't ask questions you can infer or that only matter during     
+implementation.
 
 ## 1. Explore the idea
 
-Start conversationally. Understand:
+Understand:
 - What problem does this solve?
 - Who is the target user?
 - What actions/data does it involve?
 - What exists today (website, API, product)?
 
-Ask clarifying questions naturally. Surface issues as they arise. Don't over-explain when things look good.
+When you have a clear idea of what the app does, move to the next step.
 
 ## 2. Validate against UX principles
 
-Creating a great ChatGPT app is about delivering a focused, conversational experience that feels native to ChatGPT.
-The goal is to design experiences that feel consistent and useful while extending what can be done in ChatGPT conversations in ways that add real value.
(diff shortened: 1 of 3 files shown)


skills/creating-chatgpt-app/SKILL.md
```diff
@@ -0,0 +1,17 @@
+---
+name: creating-chatgpt-app
```
fredericbarthelet: is this intentional to go with participe present instead of simple verb? `create-chatgpt-app` vs `creating-chatgpt-app` ?

paulleseute: This is what Anthropic [recommends](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices#naming-conventions)

rodincave: Weirdly enough that's not what they do in their examples repo: https://github.com/anthropics/skills/tree/main/skills
nor openai https://github.com/openai/skills/tree/main/skills
I'd go with build-chatgpt-apps for marketing/discovery purposes (create being a little restrictrive on the scope and intent)

paulleseute: What about "chatgpt-app-builder" ? cf anthropics [mcp-builder](https://github.com/anthropics/skills/tree/main/skills/mcp-builder)

paulleseute: There a lot of "-builder" and "-creator" on skillsmp

fredericbarthelet: Like the `chatgpt-app-builder` idea

skills/chatgpt-app-builder/references/brainstorm.md
```diff
@@ -0,0 +1,85 @@
+# Brainstorming Workflow
+
+Help users validate and refine their ChatGPT app idea before implementation. Be brief - only go deep when there are issues to address.
```
fredericbarthelet: > Be brief - only go deep when there are issues to address.

Less is more in context engineering. I don't know in advance the final size the skill will have. I'd stick to the strict essential in a first approach and expand only when necessary

paulleseute: I added that bit because Claude was too verbose (border-line sycophantic) on why a given aspects of an idea are matching the guidelines. What I want is the LLM to provide actual guidance on stuff to fix, not discourse on what already works.

skills/creating-chatgpt-app/SKILL.md
paulleseute: There's already one on [skillsmp](https://skillsmp.com/skills/aiskillstore-marketplace-skills-bayramannakov-chatgpt-app-builder-skill-md) (which is actually not bad), nothing prevents us to have the same name but we might want to have something different anyway.

rodincave: chatgpt-app-builder / creator sound good yes, the other skill seems to have only 12 stars so might not be an issue: https://github.com/modelcontextprotocol/ext-apps/tree/main/plugins/mcp-apps/skills

paulleseute: OK - whichever you like, I don't have any strong opinion on this.

PR #351, opened 2026-01-21
feat: add repo option to create command

paulleseute:
So user can scaffold projects from templates. I'll update the example apps READMEs once this is released.

### DIFF
--- a/packages/create-skybridge/src/index.ts
+++ b/packages/create-skybridge/src/index.ts
@@ -126,19 +126,13 @@ export async function init(args: string[] = process.argv.slice(2)) {
   const root = path.join(process.cwd(), targetDir);
 
   // 3. Download from repo or copy template
-  if (argRepo) {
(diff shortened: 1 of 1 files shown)


packages/create-skybridge/src/index.ts
```diff
@@ -114,28 +125,53 @@ export async function init(args: string[] = process.argv.slice(2)) {
...
+      fs.renameSync(
+        path.join(root, "_gitignore"),
+        path.join(root, ".gitignore"),
+      );
+      prompts.log.success(`Project created in ${root}`);
+    } catch (error) {
+      prompts.log.error("Failed to copy repository");
+      console.error(error);
+      process.exit(1);
+    }
```
fredericbarthelet: try/catch can be put around the if/else with a generic enough error message to avoid duplication

paulleseute: @fredericbarthelet PTAL

PR #357, opened 2026-01-21
Improves significantly `skybridge dev` output

harijoe:
image

<h3>Greptile Summary</h3>

Introduces two new React hooks (`useNodemon` and `useTypeScriptCheck`) to improve the dev command experience by displaying real-time TypeScript errors and server logs in the terminal UI. The hooks spawn background processes (nodemon for server watching and tsc for type checking) and capture their output to display in the Ink-based CLI interface. Template package dependencies switched to workspace references for local development.

- This PR is safe to merge with minor considerations for edge case behavior
- The implementation is well-structured with proper cleanup handlers and ref management. The refactoring addresses previous concerns about memory leaks by properly removing listeners before adding new ones. However, deducting one point due to the `env` dependency in useEffect which could theoretically cause re-initialization if env reference changes, though in practice this should be handled correctly by the singleton nature of nodemon.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -458,7 +458,7 @@ importers:
         version: 0.18.0(@types/node@25.0.3)(@types/react-dom@19.2.3(@types/react@19.2.8))(@types/react@19.2.8)(hono@4.11.3)(typescript@5.9.3)
       '@skybridge/devtools':
         specifier: '>=0.22.0 <1.0.0'
-        version: 0.25.0(@types/node@25.0.3)(@types/react-dom@19.2.3(@types/react@19.2.8))(@types/react@19.2.8)(hono@4.11.3)(immer@9.0.21)(jiti@2.6.1)(lightningcss@1.30.2)(react-router-dom@7.11.0(react-dom@19.2.3(react@19.2.3))(react@19.2.3))(react-router@7.11.0(react-dom@19.2.3(react@19.2.3))(react@19.2.3))(terser@5.44.1)(tsx@4.21.0)(typescript@5.9.3)(use-sync-external-store@1.6.0(react@19.2.3))(yaml@2.8.2)(zod@4.3.5)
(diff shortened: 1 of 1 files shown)


packages/create-skybridge/template/package.json
```diff
@@ -16,13 +16,13 @@
     "express": "^5.2.1",
     "react": "^19.2.3",
     "react-dom": "^19.2.3",
-    "skybridge": ">=0.25.0 <1.0.0",
+    "skybridge": "workspace:*",
```
fredericbarthelet: I don't think this should be committed

packages/create-skybridge/template/package.json
```diff
@@ -16,13 +16,13 @@
...
     "vite": "^7.3.1",
     "zod": "^4.3.5"
   },
   "devDependencies": {
     "@modelcontextprotocol/inspector": "^0.18.0",
-    "@skybridge/devtools": ">=0.22.0 <1.0.0",
+    "@skybridge/devtools": "workspace:*",
```
fredericbarthelet: same here

pnpm-lock.yaml
fredericbarthelet, line 1: this shouldn't be changed

packages/core/src/cli/use-nodemon.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const handleStderrData = (chunk: Buffer) => {
+      const message = chunk.toString().trim();
+      if (message) {
+        setMessages((prev) => [...prev, { text: message, type: "error" }]);
+      }
+    };
+
+    const setupStdoutListener = () => {
+      // @ts-expect-error - nodemon types don't include "stdout" property
```
fredericbarthelet: given the quantity of ts-expect-error statement, I think you should be better off overwritting the types of nodemon to include the missing properties, don't you think?

fredericbarthelet (approved):
A few comments but otherwise good to go

PR #383, opened 2026-01-22
feat: add architecture and bootstrap to chatgpt-app-builder skill

paulleseute:
- Architecture: widgets and tools definitions are inferred from UX flows. Some guidance is provided to differentiate widgets from tools and when/how to use them.
- Bootstrap: a project is scaffolded with the `create skybridge` command. Minimal boilerplate code for the user use-case replaces the magic ball examples.

### DIFF
--- a/skills/chatgpt-app-builder/references/bootstrap.md
+++ b/skills/chatgpt-app-builder/references/bootstrap.md
@@ -1,0 +1,195 @@
+# Bootstrap Workflow
+
+Scaffold a Skybridge project and generate stub code for the architecture. Skybridge is a TypeScript framework for building ChatGPT apps (MCP servers) with type-safe APIs and React widgets.
+
+## Skybridge Project Structure
+
+```
+my-app/
+├── server/src/
+│   └── index.ts          # McpServer with registerWidget/registerTool
+├── web/src/
+│   ├── widgets/          # React components (filename = widget name)
+│   │   └── search-flights.tsx
+│   ├── index.css         # Global CSS
+│   └── helpers.ts        # Type-safe hooks via generateHelpers
+└── package.json
+```
+
+**Naming convention**: Widget filename must match registered name using kebab-case.
+`search_flights` → register as `search-flights` → file `search-flights.tsx`
+
+## Step 1: Check for Existing Project
+
+Search for `package.json` with `skybridge` in dependencies.
+
+- **If found**: Skip to Step 3 (update routes only)
+- **If not found**: Continue to Step 2
+
+## Step 2: Scaffold New Project
+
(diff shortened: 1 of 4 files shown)


skills/chatgpt-app-builder/references/bootstrap.md
```diff
@@ -0,0 +1,195 @@
+# Bootstrap Workflow
+
+Scaffold a Skybridge project and generate stub code for the architecture. Skybridge is a TypeScript framework for building ChatGPT apps (MCP servers) with type-safe APIs and React widgets.
+
+## Skybridge Project Structure
```
fredericbarthelet, line 5: I feel like this should be part of the embedded via external documentation part of the skill
My concern is getting out of sync - there is very little chances to every go out of sync on the architecture part of the skill, but the bootstrap being that detailed on generated project structure feels like it belongs in template README (closer to source) and that skill should reference output README from npm create command

skills/chatgpt-app-builder/references/bootstrap.md
```diff
@@ -0,0 +1,195 @@
...
+For each widget/tool in the architecture, generate **stub code only**.
+
+**Simple placeholders only - no implementation:**
+- Server handlers: return empty/mock data matching the output schema
+- React components: basic UI skeleton showing input/output
+- Do NOT write business logic, API calls, or data fetching
+- Do NOT add error handling, validation, or edge cases
+- Implement the bare minimum boilerplate so the app can run
+
+### Server: `server/src/index.ts`
```
fredericbarthelet, line 53: all following code snippets feel too brittle and also not at the correct step. Can't you just reference the output files from the generation command so that the LLM actually grep from those ?

skills/chatgpt-app-builder/references/bootstrap.md
rodincave: for this we would need to test if the agents actually check the documentation outside the skill every time - i saw the mentions of project structure in most of the skills linked to frameworks (but can't say if it's backed by science or just a trend) [example](https://skillsmp.com/skills/einverne-dotfiles-claude-skills-nextjs-skill-md)

paulleseute: I feel having references documents referencing external stuff make the process more cumbersome. Anthropic docs says to keep the nesting level to 1 (SKILL.md -> references). I feel it's acceptable to duplicate the core concepts within the skill, even if it's more maintenance.

skills/chatgpt-app-builder/references/bootstrap.md
paulleseute: The snipped are "show don't tell" examples to refrain LLM from outputting verbose boilerplate (or even worse, starting to hallucinate an actual implementation) when the goal here is to have a working skeleton fit for the user app needs. I've been testing without at first and what I observed is that coding agent really do like to code.

PR #388, opened 2026-01-23
feat: add run-locally to chatgpt-app-builder skill

paulleseute:
So agents can install, run and expose local servers. Also reworked the boostrap app to further restrict the LLM start implementing - which wouldn't be an issue per se, but it produces a lot of React boilerplate without using Skybridge.

### DIFF
--- a/skills/chatgpt-app-builder/references/run-locally.md
+++ b/skills/chatgpt-app-builder/references/run-locally.md
@@ -10,7 +10,7 @@ Install dependencies and start the dev server in the background:
 
 For Deno projects, use `deno task dev` instead.
 
-When started, output the local MCP server and playground URL.
+When started, output the local MCP server and devtools URL.
 
 ## 2. Connect to ChatGPT
 


skills/chatgpt-app-builder/references/run-locally.md
```diff
@@ -0,0 +1,41 @@
...
+{pm} install && {pm} run dev
+```
+
+For Deno projects, use `deno task dev` instead.
+
+When started, output the local MCP server and playground URL.
+
+## 2. Connect to ChatGPT
+
+Run ngrok in the background to expose the app:
```
fredericbarthelet, line 17: what happens if ngrok is not installed?

skills/chatgpt-app-builder/references/run-locally.md
```diff
@@ -0,0 +1,41 @@
...
+
+Install dependencies and start the dev server in the background:
+
+```bash
+{pm} install && {pm} run dev
+```
+
+For Deno projects, use `deno task dev` instead.
+
+When started, output the local MCP server and playground URL.
```
fredericbarthelet: playground URL is only available post deployement. You mean the devtool URL ?

skills/chatgpt-app-builder/references/run-locally.md
paulleseute: command will fail and the agent will offer to install it (or provide guidance to do so) - ngrok is well documented and agent are familiar with it.

PR #392, opened 2026-01-26
Integrate index.ts and middleware.ts into skybridge commands

harijoe:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR updates Skybridge’s core server layer to expose an Express-based HTTP server (`packages/core/src/server/express.ts`) and wires it into `McpServer` via new `run()` and `use()` APIs (`packages/core/src/server/server.ts`). The create-skybridge template is adjusted accordingly to rely on `McpServer.run()` (removing its bespoke Express entrypoint/middleware) and updates the web helper type import to match the new server entry file.

Dependency metadata is updated to include `@skybridge/devtools` as a peer dep for core, and the template pins `@skybridge/devtools`/`skybridge` versions differently to fit the workspace template setup.

- This PR is mergeable but has a few behavioral edge cases that can cause runtime issues in real deployments.
- Core changes introduce a new Express-backed `run()`middleware mechanism and an MCP HTTP transport handler; the implementation looks plausible, but there are a couple of concrete lifecycle/type-resolution pitfalls (per-request `connect` semantics, template type import strategy) that can break users depending on concurrency and TS config. Fixing these should reduce surprise and production issues.
- packages/core/src/server/express.ts, packages/core/src/server/server.ts, packages/create-skybridge/template/web/src/helpers.ts

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,16 +61,13 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Install template dependencies with workspace package versions
+        run: pnpm add "skybridge@workspace:0.0.0" "@skybridge/devtools@workspace:0.0.0" --filter apps-sdk-template
+
       - name: Build core and devtools
         run: |
           pnpm --filter skybridge build
           pnpm --filter @skybridge/devtools build
-
-      - name: Install template dependencies with workspace package versions
-        run: pnpm add "skybridge@workspace:0.0.0" "@skybridge/devtools@workspace:0.0.0" --filter apps-sdk-template
-
-      - name: Install (link workspace packages)
-        run: pnpm install
 
       - name: Build template
         run: pnpm --filter apps-sdk-template run build


packages/core/src/commands/build.tsx
```diff
@@ -1,11 +1,47 @@
...
+        throw new Error(
+          `server.ts not found at ${serverTsPath}. Please ensure your server.ts file exists in server/src/`,
+        );
+      }
+
+      const indexTsPath = join(serverSrcDir, "index.ts");
+      if (existsSync(indexTsPath)) {
+        throw new Error(
+          `There should be no index.ts file in server/src. Please remove it before building.`,
+        );
```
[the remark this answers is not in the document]
harijoe: Not now, will do it in a separate PR

packages/core/src/cli/use-execute-steps.ts
```diff
@@ -38,8 +38,12 @@ export const useExecuteSteps = (steps: CommandStep[]) => {
       setImmediate(() => {
         process.exit(1);
       });
+    } finally {
+      if (cleanup) {
+        cleanup();
+      }
```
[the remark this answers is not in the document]
harijoe: This is not true. The step 3 is part of the commands run by useExecuteSteps and cleanup is used after all commands have been run (or failed)

packages/core/src/server/index.ts
```diff
@@ -6,6 +6,7 @@ export type {
   ToolOutput,
   ToolResponseMetadata,
 } from "./inferUtilityTypes.js";
+export { createMcpMiddleware } from "./mcpMiddleware.js";
```
fredericbarthelet: I wouldn't export things from here that are not to be used by the project implementation, WDYT ?

packages/core/src/commands/build.tsx
```diff
@@ -21,15 +60,27 @@ export const commandSteps: CommandStep[] = [
   },
 ];
 
+const cleanup = () => {
+  if (indexTsWasCreated) {
+    const projectRoot = process.cwd();
+    const serverIndex = resolve(projectRoot, "server/src/index.ts");
+    if (existsSync(serverIndex)) {
+      rmSync(serverIndex);
```
fredericbarthelet: why do you want to cleanup afterwards ? We should generate server entry in the `dist` output, just like xmcp for per-transport entries (http.js or stdio.js) -> https://xmcp.dev/docs/getting-started/installation#troubleshooting

packages/core/src/server/index.ts
harijoe: The idea was to still allow a user to do his own implementation but I agree we can remove it 👍

packages/core/src/commands/build.tsx
harijoe: It looks a bit more challenging but I will give it a try

packages/core/package.json
```diff
@@ -41,6 +41,7 @@
   "license": "ISC",
   "peerDependencies": {
     "@modelcontextprotocol/sdk": ">=1.0.0",
+    "@skybridge/devtools": ">=0.26.1",
```
fredericbarthelet, line 44: add it also to the dev dependency so that you can use it in tests and keep the ci.yml unchanged

.github/workflows/ci.yml
fredericbarthelet, line 1: should be unchanged, see my comment on package.json to ensure it stays unchanged

packages/core/src/server/server.ts
```diff
@@ -208,10 +218,65 @@ type ToolHandler<
...
+    const express = this.express;
+    return new Promise((resolve, reject) => {
+      const server = express.listen(3000, () => {
+        resolve();
+      });
+
+      server.on("error", (error: Error) => {
+        console.error("Failed to start server:", error);
+        reject(error);
+      });
```
fredericbarthelet, line 276: shouldn't you declare your error handler before starting the listener?

fredericbarthelet (approved):
A few small comments, but otherwise good to go :)
Could you also update the template readme that mentions project structure (affected by this PR)

.github/workflows/ci.yml
harijoe: Still had to fix the local module resolution

PR #425, opened 2026-02-02
feat: add skill install to create cmd

paulleseute:
Which is having an optional step running `npx skills add` under the hood.

### DIFF
--- a/packages/create-skybridge/template/AGENTS.md
+++ b/packages/create-skybridge/template/AGENTS.md
@@ -1,0 +1,1 @@
+Before writing code, first explore the project structure, then invoke the chatgpt-app-builder skill for documentation.


packages/create-skybridge/src/index.ts
```diff
@@ -173,7 +173,35 @@ export async function init(args: string[] = process.argv.slice(2)) {
...
+      message: "Install the coding agents skills? (recommended)",
+      initialValue: true,
+    });
+    if (prompts.isCancel(skillsResult)) {
+      return cancel();
+    }
+    if (skillsResult) {
+      run(
+        [
+          "npx",
```
fredericbarthelet: use `pkgManager` instead

paulleseute: @fredericbarthelet PTAL

fredericbarthelet (approved):

paulleseute:
> A more metaphysical question: will this add both chagpt-apps and mcp-apps skill?
> 
> Also, what do we think of:
> 
> * https://vercel.com/blog/agents-md-outperforms-skills-in-our-agent-evals

Regarding which skill it'll be adding: I'm not sure yet, I guess only the ChatGPT variant since the MCP Apps one will be a subset of it. Regarding the template AGENTS.md, I need to think about how we can package the skill into the template without bloating the project, which I assume is not trivial. For now I'm creating the markdown file with the "explicit instruction" thing, which should improve the trigger rate.

PR #429, opened 2026-02-03
feat: add bump script

paulleseute:
Bumps the Skybridge version of the template and all example apps.

### DIFF
--- a/scripts/bump.js
+++ b/scripts/bump.js
@@ -7,7 +7,7 @@
  *   node scripts/bump.js 0.30.0   # Uses specific version
  */
 import { execSync } from "child_process";
-import { existsSync, readFileSync, writeFileSync } from "fs";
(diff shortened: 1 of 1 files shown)


scripts/bump.js
```diff
@@ -0,0 +1,61 @@
...
+console.log(`Setting skybridge version to: ${version}`);
+
+const versionRange = `>=${version} <1.0.0`;
+
+const targets = [
+  "packages/create-skybridge/template/package.json",
+  "examples/capitals/package.json",
+  "examples/everything/package.json",
+  "examples/ecom-carousel/package.json",
+  "examples/productivity/package.json",
```
valentinbeggi: maybe can list dir under examples so we don't have to explicitely reference them here ?

paulleseute: good idea

PR #435, opened 2026-02-04
feat: use openai widget state shape

paulleseute:
We've been treating the entire widget state as model content, but openai's actual api uses a structured format (modelContent, privateContent, imageIds). This change wraps Skybridge widget state in modelContent at the adaptor boundary so we don't introduce breaking changes. This paves the way for pushing file imagesIds into the widget (and possibly private content).

### DIFF
--- a/packages/core/src/web/bridges/apps-sdk/types.ts
+++ b/packages/core/src/web/bridges/apps-sdk/types.ts
@@ -9,7 +9,7 @@ import type {
 type DisplayMode = "pip" | "inline" | "fullscreen" | "modal";
 
 type WidgetState = {
-  modelContent: Record<string, unknown> | null;
+  modelContent: Record<string, unknown>;
 };
 
 export const TOOL_RESPONSE_EVENT_TYPE = "openai:tool_response";
@@ -33,6 +33,7 @@ export type AppsSdkContext<
   ToolInput extends UnknownObject = Record<never, unknown>,
   ToolOutput extends UnknownObject = UnknownObject,
   ToolResponseMetadata extends UnknownObject = UnknownObject,
+  WS extends WidgetState = WidgetState,
 > = {
   theme: Theme;
   userAgent: UserAgent;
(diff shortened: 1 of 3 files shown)


packages/core/src/web/bridges/apps-sdk/types.ts
```diff
@@ -8,7 +8,9 @@ import type {
 
 type DisplayMode = "pip" | "inline" | "fullscreen" | "modal";
 
-type WidgetState = UnknownObject;
+type WidgetState = {
+  modelContent: Record<string, unknown> | null;
```
fredericbarthelet: The `| null` looks breaking to me, especially because it makes previously returned value from widgetState potentially nullable, no?

```suggestion
  modelContent: UnknownObject
```

packages/core/src/web/bridges/apps-sdk/types.ts
```diff
@@ -50,61 +51,58 @@ export type AppsSdkContext<
...
-       */
-      mode: DisplayMode;
-    }>;
-
-    /**
-     * Sets the widget state.
-     * This state is persisted across widget renders.
-     */
-    setWidgetState: (state: WidgetState) => Promise<void>;
+export type AppsSdkMethods = {
```
fredericbarthelet: why removing the generic here? The type is exported, so this will be breaking for people directly using it don't you think?

packages/core/src/web/data-llm.test.tsx
```diff
@@ -53,7 +53,7 @@ describe("DataLLM", () => {
       // Keep the mock available for React cleanup, but reset it
       if (typeof window !== "undefined" && window.openai) {
         window.openai.setWidgetState = vi.fn();
-        window.openai.widgetState = {};
+        window.openai.widgetState = { modelContent: null };
```
fredericbarthelet:
```suggestion
        window.openai.widgetState = { modelContent: {} };
```

fredericbarthelet (approved):
LGTM. But I have 2 main concerns that you might want to address before merging :
- breaking change on the Apps SDK exported bridge type. Even if we don't use the generic type for widget state in skybridge code base itself, some using the lib might be using it. Do we really want to break this simple generic type?
- I feel like there is also a breaking change on the return value from `useWidgetState` that previously couldn't return null value for existing state, right?

Issue #444, opened 2026-02-07
devtools disappear if the default `/mcp` route is changed


ppshobi:
there is no way to tell the devtools that the `/mcp` endpoint is elsewhere . lets say I add some `uuid-1/mcp` to be the mcp endpoint, the devtools still look for `/mcp` endpoint.

ppshobi:
@harijoe my mcp app is living at a different endpoint that the default localhost. right now I am using mcpjam to make it work

PR #449, opened 2026-02-10
feat: use an env variable `VITE_MCP_SERVER_URL` to make it possible that devtools can point at a different url

ppshobi:
see https://github.com/alpic-ai/skybridge/issues/444

This pull request adds a support that the devtools can point at a different mcp server url than the hardcoded `localhost:3000/mcp`. while defaulting to the same

<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR changes the devtools MCP client initialization to allow overriding the default MCP endpoint (previously hardcoded to `http://localhost:3000/mcp`) via an environment variable (`MCP_SERVER_URL`).

The change is implemented in `packages/devtools/src/lib/mcp/index.ts`, which is imported by multiple React UI components and therefore runs in the Vite browser bundle at module load time.

- Not safe to merge as-is due to a likely browser runtime error in devtools.
- The PR reads `process.env.MCP_SERVER_URL` in code that is bundled for the browser via Vite. Without explicit Vite `define` configuration or using `import.meta.env`, `process` is typically undefined in the browser, which can crash the devtools UI at startup. Fixing the env access pattern should make the change safe.
- packages/devtools/src/lib/mcp/index.ts

### DIFF
--- a/packages/devtools/src/lib/mcp/index.ts
+++ b/packages/devtools/src/lib/mcp/index.ts
@@ -6,8 +6,8 @@ import { useSelectedToolName } from "../nuqs.js";
 import { McpClient } from "./client.js";
 
 const client = new McpClient();
-
-client.connect("http://localhost:3000/mcp").then(() => {
+const mcpServerUrl = process.env.MCP_SERVER_URL || 'http://localhost:3000/mcp';
+client.connect(mcpServerUrl).then(() => {
   console.info("Connected to MCP server");
 });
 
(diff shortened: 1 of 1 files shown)


harijoe:
@connorads Indeed we now ship a tunnel with skybridge that allows exactly that 👌

@ppshobi Thanks for your contribution, I'm closing this for now but took note of the feature suggested — using the devtools on a different server than the one started with the `skybridge dev` commands — and will talk with the team if this is something we want to push further.

PR #451, opened 2026-02-10
Add murder in the valley in skybridge examples

csonigo:
<h2>Greptile Overview</h2>

<h3>Greptile Summary</h3>

This PR adds a new `examples/investigation-game` workspace containing an Express-based MCP server (`server/`) and a Vite/React widget frontend (`web/`) implementing the “Murder in the Valley” investigation game. The server registers the widget/tool via `skybridge/server` and exposes it at `/mcp`, while the web package contains the interactive multi-screen UI (start/intro/main/victory) plus assets, puzzle logic, and suspect data.

Key merge blockers are around the new example’s wiring/config correctness: the `web/package.json` is currently invalid JSON due to an unterminated string, the MCP middleware connects the `McpServer` on every request (likely leaking/duplicating state), and the widget metadata hardcodes an external `domain` that will break typical local dev flows (ngrok/emulator) unless manually edited.

- Not safe to merge until JSON validity and MCP lifecycle issues are fixed
- The PR introduces a parse-breaking `web/package.json` (tooling will fail immediately) and an MCP middleware pattern that connects the server on each POST without disconnecting, which is likely to cause leaked/duplicated state under repeated requests. There’s also a hardcoded widget domain that will misconfigure local development for an example project.
- examples/investigation-game/web/package.json; examples/investigation-game/server/src/middleware.ts; examples/investigation-game/server/src/server.ts

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -25,7 +25,7 @@ importers:
     devDependencies:
       mintlify:
         specifier: ^4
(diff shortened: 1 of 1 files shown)


paulleseute (commented):
I think we need to migrate this to be on par with the template, meaning a single/manager agnostic package. I can take of it if you want.

PR #484, opened 2026-02-17
feat: add oauth support on devtools

qchuchu:
- Currently an Oauth server cannot use the capabilities of the devtools
- Adding a behavior to be able to test oauth servers on the devtools

image

<h3>Greptile Summary</h3>

This PR adds OAuth support to the Skybridge devtools, enabling connection to OAuth-protected MCP servers. It introduces a `BrowserOAuthProvider` for handling the OAuth flow via localStorage, a Zustand-based auth store for managing connection state, and updates the UI to conditionally render based on authentication status.

- **`main.tsx`**: Top-level `await init()` blocks React rendering until the connection attempt completes. If the server is slow or unreachable, users see a blank white page instead of the "Connecting..." UI that `AppLayout` already supports. Additionally, when the OAuth callback URL is present but the `code` parameter is missing (e.g., user denied consent), the app silently stays in "idle" state with no error feedback.
- **`index.ts`**: The `logout()` function calls `client.close()` without `await`, which can lead to race conditions during teardown. The `useSuspenseTools` function name is now misleading since it was changed from `useSuspenseQuery` to `useQuery`.
- **`browser-oauth-provider.ts`**, **`auth-store.ts`**, **`client.ts`**, **`status-badge.tsx`**: Clean implementations with no issues found.

- The core OAuth flow is well-structured, but the startup blocking and missing error handling in the callback path need attention before merging.
- Score of 3 reflects that the overall architecture is sound, but there are two functional issues in main.tsx (blocking render and silent failure on missing OAuth code) and one in index.ts (unawaited async close in logout) that could impact user experience and reliability.
- `packages/devtools/src/main.tsx` and `packages/devtools/src/lib/mcp/index.ts` need the most attention.

<sub>Last reviewed commit: e117a2e</sub>

### DIFF
--- a/packages/devtools/src/lib/mcp/index.ts
+++ b/packages/devtools/src/lib/mcp/index.ts
@@ -4,7 +4,7 @@ import {
   UnauthorizedError,
 } from "@modelcontextprotocol/sdk/client/auth.js";
 import type { Tool } from "@modelcontextprotocol/sdk/types.js";
-import { useMutation, useQuery, useSuspenseQuery } from "@tanstack/react-query";
+import { useMutation, useSuspenseQuery } from "@tanstack/react-query";
 import type { AppsSdkContext, CallToolArgs } from "skybridge/web";
 import { useAuthStore } from "@/lib/auth-store.js";
 import { useStore } from "@/lib/store.js";
@@ -68,10 +68,10 @@ export async function finishOAuthCallback(code: string): Promise<void> {
   await connectToServer();
 }
 
-export function logout(): void {
+export async function logout(): Promise<void> {
   currentAuthProvider?.invalidateCredentials("all");
   currentAuthProvider = null;
-  client.close();
+  await client.close();
   useAuthStore.getState().reset();
   queryClient.invalidateQueries({ queryKey: ["list-tools"] });
 }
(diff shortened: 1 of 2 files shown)


packages/devtools/src/lib/mcp/index.ts
```diff
@@ -1,15 +1,80 @@
...
 import { useMutation, useSuspenseQuery } from "@tanstack/react-query";
 import type { AppsSdkContext, CallToolArgs } from "skybridge/web";
+import { useAuthStore } from "@/lib/auth-store.js";
 import { useStore } from "@/lib/store.js";
 import { useSelectedToolName } from "../nuqs.js";
+import { queryClient } from "../query-client.js";
+import { BrowserOAuthProvider } from "./browser-oauth-provider.js";
 import { McpClient } from "./client.js";
 
+const DEFAULT_SERVER_URL = "http://localhost:3000/mcp";
```
harijoe, line 16: We may have a problem when port selection will be dynamic thanks to @thleonard work? Maybe we can just assume that the server is running on `${window.location.origin}/mcp`?

qchuchu: Yes this should be changed but it was already the case - we were connecting by default to http://localhost:3000.

But we'll need to change once yesss

qchuchu: And also it breaks the local development if doing that no ?

PR #487, opened 2026-02-17
fix: use request URL for content domain hash

paulleseute:
so mcp servers served on other endpoints that /mcp work with claude

### DIFF
--- a/packages/core/src/server/express.ts
+++ b/packages/core/src/server/express.ts
@@ -72,6 +72,9 @@ const mcpMiddleware = (server: McpServer): express.RequestHandler => {
       });
 
       await server.connect(transport);
+      // Express strips the mount path from req.url (e.g. "/mcp" becomes "/").
+      // Restore it so the SDK builds the correct requestInfo.url.
+      req.url = req.originalUrl;
       await transport.handleRequest(req, res, req.body);
     } catch (error) {
       console.error("Error handling MCP request:", error);
--- a/packages/core/src/server/server.ts
+++ b/packages/core/src/server/server.ts
@@ -497,7 +497,10 @@ export class McpServer<
           domain: isClaude
             ? `${crypto
                 .createHash("sha256")
-                .update(`https://${hostFromHeaders}/mcp`)
+                .update(
+                  extra.requestInfo?.url?.href ||
+                    `https://${hostFromHeaders}/mcp`,
+                )
                 .digest("hex")
                 .slice(0, 32)}.claudemcpcontent.com`
             : serverUrl,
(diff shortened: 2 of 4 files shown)


packages/core/src/server/server.ts
```diff
@@ -488,7 +497,10 @@ export class McpServer<
           domain: isClaude
             ? `${crypto
                 .createHash("sha256")
-                .update(`https://${hostFromHeaders}/mcp`)
+                .update(
+                  extra.requestInfo?.url?.href ||
+                    `https://${hostFromHeaders}/mcp`,
```
qchuchu, line 502: Do we still need this part then ? As we ensured the url should be set

qchuchu: the hostFromHeaders/mcp

paulleseute: i guess so, but typing says it's optional so I guess it does not hurt to keep it

packages/core/src/server/express.ts
```diff
@@ -72,6 +72,9 @@ const mcpMiddleware = (server: McpServer): express.RequestHandler => {
       });
 
       await server.connect(transport);
+      // Express strips the mount path from req.url (e.g. "/mcp" becomes "/").
+      // Restore it so the SDK builds the correct requestInfo.url.
+      req.url = req.originalUrl;
```
fredericbarthelet, line 77: does this mean that we should instead always use `req.originalUrl` in https://github.com/modelcontextprotocol/typescript-sdk/blob/342ea394ca6e660e294162efdeafc411284bcc0d/src/server/webStandardStreamableHttp.ts#L611

I agree with this fix in the meantime, but if you can open the correcting issue (targeting v1.x branch), in order to fix this, that'd be great.
Can you also add a `@debt` marker here to make it clear it should be removed once the PR is merged?

packages/core/src/server/server.ts
fredericbarthelet: you can use `serverUrl` that is more generic as a fallback instead I think, no?

packages/core/src/server/express.ts
fredericbarthelet: @valentinbeggi you might want to check this as well

fredericbarthelet (approved):
Good to go as his, even if some work is required to always have the original url form the SDK side I think.

@valentinbeggi opened a PR on the core hosting platform to pass query param down to the server instance on Alpic hosting platform (https://github.com/alpic-ai/alpic/pull/840). There might be an additional required work to pass path as well

packages/core/src/server/express.ts
paulleseute: the originalUrl stuff is specific to express isn't it ? the sdk deals with vanilla request object

Issue #417, opened 2026-01-30
[FEAT] Support `-p / --port` option for `skybridge dev` and `skybridge start`


fpasquet:
Hello,

It would be very useful if `skybridge` supported a `-p` / `--port` (or `--port`) argument for both the `dev` and `start` commands, similar to what is available in Next.js.

Example usage:

```bash
skybridge dev --port 4000
skybridge start --port 4000
```

or

```bash
skybridge dev -p 4000
skybridge start -p 4000
```

### Motivation

In many real-world environments (monorepos, CI pipelines, containerized setups, or when running multiple services locally), the default port may already be in use or needs to be explicitly defined.

Having a CLI option to configure the port would:

* Improve developer experience
* Enable better integration with monorepos and process managers
* Align `skybridge` with common tooling conventions (e.g. Next.js, Vite, NestJS)
* Avoid relying solely on environment variables for basic runtime configuration

### Expected behavior

If provided, the `--port` or `-p` argument should override the default port and be used consistently in both `dev` and `start` modes.

## Bonus

* Detect if the port is already in use => Fallback to an other port instead :

```
✔ [WARNING] Something is already running on port 3000. Probably:
  `/Users/paul/.nvm/versions/node/v24.11.1/bin/node --require /Users/paul/dev/skybridge/node_modules/.pnpm/tsx@4.21.0/node_modules/tsx/dist/preflight.cjs --import file:///Users/paul/dev/skybridge/node_modules/.pnpm/tsx@4.21.0/node_modules/tsx/dist/loader.mjs server/src/index.ts` (pid 55027) in "/Users/paul/dev/skybridge/examples/capitals"

Would you like to run the app on another port instead? … yes
[SUCCESS] Docusaurus website is running at: http://localhost:3001/
```

* Also works with an environment variable (SKY_PORT)
* flag > SKY_PORT > fallback

qchuchu:
Hi @fpasquet ! As said on the PR you opened, @thleonard is currently working on providing a solution. It should be released today or on monday !

PR #493, opened 2026-02-18
Theo/sky 197 feat support p port option for skybridge dev and skybridge

thleonard:
PR to add the feature request #417

```
> pnpm run dev
Port 3000 is in use, falling back to port 51352
Server running at: http://localhost:51352/mcp
```

```
> pnpm run dev -p 3001
Server running at: http://localhost:3001/mcp
```

```
> export SKYBRIDGE_PORT=3001 && pnpm run dev
Server running at: http://localhost:3001/mcp
```

```
> export SKYBRIDGE_PORT=3001 && pnpm run dev -p 3002
Server running at: http://localhost:3002/mcp
```

<h3>Greptile Summary</h3>

This PR adds configurable port support for `skybridge dev` and `skybridge start` commands through three mechanisms: a `-p` CLI flag, the `SKYBRIDGE_PORT` environment variable, and automatic fallback to a random free port when the default port (3000) is occupied. It also makes the Vite HMR WebSocket port dynamic via auto-detection and updates the devtools client to use `window.location.origin` instead of a hardcoded `localhost:3000`.

- Adds new `detect-port.ts` utility using `net.createServer` to check port availability and find alternatives
- Both `dev.tsx` and `start.ts` now resolve the port with precedence: CLI flag > `SKYBRIDGE_PORT` env > auto-detect from 3000
- The resolved port is passed to child processes via `SKYBRIDGE_PORT` env variable, which `server.ts` reads at listen time
- HMR port is auto-detected in `widgetsDevServer.ts` and shared via `SKYBRIDGE_HMR_PORT` env for CSP headers in `server.ts`
- Devtools MCP client now connects to `window.location.origin/mcp` instead of hardcoded `localhost:3000/mcp`
- Removes unused `"declaration": true` from example tsconfig files (unrelated cleanup)

- This PR is safe to merge with a minor edge case around non-numeric SKYBRIDGE_PORT values worth addressing.
- The core logic is sound and well-structured with clear port precedence (flag > env > auto-detect). The detect-port utility is clean. The NaN propagation from invalid SKYBRIDGE_PORT env values is a real but low-likelihood edge case. All hardcoded port references have been properly replaced.
- packages/core/src/commands/dev.tsx and packages/core/src/commands/start.ts need attention for the NaN propagation edge case when SKYBRIDGE_PORT contains non-numeric values.

<sub>Last reviewed commit: e068fc7</sub>

### DIFF
--- a/docs/api-reference/cli.mdx
+++ b/docs/api-reference/cli.mdx
@@ -28,7 +28,16 @@ skybridge dev
 
 | Flag | Description |
 |------|-------------|
+| `-p, --port <number>` | Port to run the server on. Defaults to `3000`, or the next available port if `3000` is in use. |
 | `--use-forwarded-host` | Uses the forwarded host header to construct widget URLs instead of localhost. Useful when accessing the dev server through a tunnel (e.g., ngrok). |
+
+You can also set the port via the `PORT` environment variable:
+
+```bash
+PORT=8080 skybridge dev
+```
+
+The flag takes precedence over the environment variable.
 
 The `--use-forwarded-host`  allows your widgets to work on any device connected to the internet, not just on your local machine. Useful when developing on multiple devices or when you need to test your app from different devices.
 


packages/core/src/cli/detect-port.ts
```diff
@@ -0,0 +1,75 @@
+import net from "node:net";
+
+const DEFAULT_PORT = 3000;
+
+export async function resolvePort(flagPort?: number) {
+  if (flagPort) {
+    return { port: flagPort, fallback: false };
+  }
+
+  const rawEnv = process.env.SKYBRIDGE_PORT;
```
qchuchu: I think we should use PORT rather than SKYBRIDGE PORT. Easier for devs.

packages/core/src/cli/detect-port.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (rawEnv) {
+    const parsed = Number(rawEnv);
+    if (Number.isInteger(parsed) && parsed > 0) {
```
qchuchu, line 13: But then we don't do the fallback on user provided port ?

packages/core/src/cli/detect-port.ts
```diff
@@ -0,0 +1,75 @@
+import net from "node:net";
+
+const DEFAULT_PORT = 3000;
+
+export async function resolvePort(flagPort?: number) {
+  if (flagPort) {
```
qchuchu: if flagPort is == 0 (weird I know) but then you don't pass by this block. 

if (flagPort !== undefined).

packages/core/src/cli/detect-port.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Returns the given port if available, otherwise lets the OS
+ * pick a free port via `listen(0)`.
+ *
+ * @param host - Bind address for the check. Pass `"localhost"` for
+ *   services that bind to 127.0.0.1 (e.g. Vite HMR). Omit for
+ *   services that bind to all interfaces (e.g. the HTTP server).
+ */
+export async function detectAvailablePort(
```
qchuchu, line 35: I think we can use a package for this no ? 

https://www.npmjs.com/package/get-port seems to be the most downloaded

qchuchu: Otherwise this is nextjs code 

server.on('error', (err: NodeJS.ErrnoException) => {
  if (
    allowRetry &&
    port &&
    isDev &&
    err.code === 'EADDRINUSE' &&
    portRetryCount < 10
  ) {
    port += 1
    portRetryCount += 1
    server.listen(port, hostname)
  } else {
    Log.error(`Failed to start server`)
    process.exit(1)
  }
})

packages/devtools/src/lib/mcp/index.ts
```diff
@@ -13,12 +13,11 @@ import { queryClient } from "../query-client.js";
 import { BrowserOAuthProvider } from "./browser-oauth-provider.js";
 import { McpClient } from "./client.js";
 
-const DEFAULT_SERVER_URL = "http://localhost:3000/mcp";
-
 const client = new McpClient();
 let currentAuthProvider: BrowserOAuthProvider | null = null;
 
 export async function connectToServer(): Promise<void> {
+  const serverUrl = `${window.location.origin}/mcp`;
```
qchuchu: You can put this in a function (not variable bc window might be set)

to then reuse it in finishOauthCallback

packages/core/src/server/widgetsDevServer.ts
```diff
@@ -45,6 +47,9 @@ export const widgetsDevServer = async (): Promise<Router> => {
     ...devConfig
   } = configResult?.config || {};
 
+  const hmrPort = await detectAvailablePort(DEFAULT_HMR_PORT, "localhost");
+  process.env.__SKYBRIDGE_HMR_PORT = String(hmrPort);
```
qchuchu, line 51: Then I think we need to also modify the csp inside the server if I'm not wrong to make it work with HMR in ChatGPT

packages/core/src/server/server.ts
```diff
@@ -478,7 +482,8 @@ export class McpServer<
 
         const connectDomains = [serverUrl];
         if (!isProduction) {
-          const VITE_HMR_WEBSOCKET_DEFAULT_URL = "ws://localhost:24678";
+          const hmrPort = process.env.__SKYBRIDGE_HMR_PORT ?? DEFAULT_HMR_PORT;
+          const VITE_HMR_WEBSOCKET_DEFAULT_URL = `ws://localhost:${hmrPort}`;
```
qchuchu, line 485: Ah my bad it's here

packages/core/src/server/server.ts
```diff
@@ -258,7 +259,9 @@ export class McpServer<
         console.error("Failed to start server:", error);
         reject(error);
       });
-      server.listen(3000, () => {
+      const rawPort = parseInt(process.env.__SKYBRIDGE_PORT ?? "3000", 10);
+      const port = Number.isNaN(rawPort) ? 3000 : rawPort;
```
qchuchu: This double check is only for typing right ? Bc normally this env variable should be already set

packages/core/src/cli/detect-port.ts
thleonard: I feel like PORT might easily conflict with other processes. What do you think ?

packages/core/src/cli/detect-port.ts
thleonard: Do you mean that i'm not checking the user provided port through flag like i'm doing for the env var ?

thleonard: The validation is done by @oclif/core. I've added a min value for the integer to 1

packages/core/src/cli/detect-port.ts
thleonard: I've seen that with Fred, he likes better to use listen(0) which auto discovers a free port than incrementing

packages/core/src/cli/detect-port.ts
thleonard: Fixed with flag validation >1

packages/core/src/server/server.ts
thleonard: I need to test again. It was to fix the issue where SKYBRIDGE_PORT=abc
But this may be catched by `if (Number.isInteger(parsed) && parsed > 0)` that i added after

packages/core/src/cli/detect-port.ts
qchuchu: Yes but not here. Then you should maybe put it here no ? if (flagPort > 1) ? WDYT ?

packages/core/src/cli/detect-port.ts
qchuchu: when you do PORT=3000 skybridge dev actually you set the env variable only for this specific process.

https://nextjs.org/docs/pages/guides/upgrading/version-11#port-is-now-supported-for-next-dev-and-next-start 

I think it's kind of market standard.

packages/core/src/cli/detect-port.ts
qchuchu: Sorry I wasn't maybe clear : you don't fallback on another port when the port was provided by a user ? 

In next.js case indeed, we increment from the provided port or the default port (3000). So that it always work.

qchuchu: But my bad - I double checked an actually Next does not fallback on a provided port - so it should raise an ADDRINUSE.

packages/core/src/commands/start.ts
```diff
@@ -1,14 +1,26 @@
...
 import { runCommand } from "../cli/run-command.js";
 
 export default class Start extends Command {
   static override description = "Start production server";
   static override examples = ["skybridge start"];
-  static override flags = {};
+  static override flags = {
+    port: Flags.integer({
+      char: "p",
+      description: "Port to run the server on",
```
qchuchu, line 13: You didn't add the min > 1 here. 

But I'd be more confident if the inner function does flagPort > 1 too

packages/core/src/server/server.ts
qchuchu: Normally __SKYBRIDGE_PORT is kind of vetted (i.e. generated by us - after detecting port) so it should always be a number

packages/core/src/cli/detect-port.ts
qchuchu: Ok - don't have a strong opinion on this

packages/core/src/cli/detect-port.ts
thleonard: Ok adding that

packages/core/src/commands/start.ts
thleonard: Good catch ! Thanks

packages/core/src/server/server.ts
thleonard: Yes that's right actually

qchuchu (approved):

PR #502, opened 2026-02-20
fix: claude domain

valentinbeggi:
<h3>Greptile Summary</h3>

Enforces Claude-specific domain security requirement by preventing user overrides to the `domain` field in MCP Apps widget metadata. After merging user-provided metadata with defaults, the code now explicitly sets `merged.ui.domain` back to the system-generated value (`{hash}.claudemcpcontent.com` for Claude). This ensures Claude's CSP requirements are always met regardless of user configuration.

- Safe to merge with no risk
- The change is a focused security fix that correctly enforces Claude's domain requirements. The logic is sound - it stores the merge result in a variable, then explicitly overwrites the domain property to prevent user overrides. The defensive `if (merged.ui)` check prevents potential null reference errors, though the defaults should always ensure `ui` is defined.

<sub>Last reviewed commit: 7f7cb23</sub>

### DIFF
--- a/packages/core/src/server/server.ts
+++ b/packages/core/src/server/server.ts
@@ -122,12 +122,15 @@ type WidgetResourceConfig<T extends ResourceMeta = ResourceMeta> = {
   uri: string;
   mimeType: string;
   /** Function to build _meta for this resource content given computed defaults */
-  buildContentMeta: (defaults: {
-    resourceDomains: string[];
-    connectDomains: string[];
-    domain: string;
-    baseUriDomains: string[];
-  }) => T;
+  buildContentMeta: (
+    defaults: {
+      resourceDomains: string[];
+      connectDomains: string[];
+      domain: string;
+      baseUriDomains: string[];
+    },
+    overrides: { domain?: string },
+  ) => T;
 };
 
 type McpServerOriginalResourceConfig = Omit<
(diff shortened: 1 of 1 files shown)


packages/core/src/server/server.ts
```diff
@@ -363,7 +363,14 @@ export class McpServer<
             },
           };
 
-          return mergeWithUnion(defaults, { ui: userMeta?.ui });
+          const merged = mergeWithUnion(defaults, { ui: userMeta?.ui });
+
+          // Claude requires domain to be {hash}.claudemcpcontent.com, user overrides are never valid
+          if (merged.ui) {
+            merged.ui.domain = domain;
+          }
```
fredericbarthelet: this is hard, it overrides the value for all client, not Claude only. User override can be completly valid, especially if it doesn't want to be available on Claude 😅

packages/core/src/server/server.ts
```diff
@@ -363,7 +369,9 @@ export class McpServer<
             },
           };
 
-          return mergeWithUnion(defaults, { ui: userMeta?.ui });
+          return mergeWithUnion(defaults, {
```
fredericbarthelet, line 375: don't forget to also implement on the apps-sdk path the inclusion of overrides (even if might not be used right now)

PR #505, opened 2026-02-23
feat: add time's up to example

paulleseute:
To be deployed on Alpic after it's merged.


examples/times-up/README.md
```diff
@@ -0,0 +1,117 @@
...
+## Live Demo
+
+[Try it in Alpic's Playground](https://times-up.skybridge.tech/try) to launch the live widget experience, or use the MCP URL with your client of choice: `https://times-up.skybridge.tech/mcp`.
+
+## Getting Started
+
+### Prerequisites
+
+- Node.js 24+
+- HTTP tunnel such as [ngrok](https://ngrok.com/download) if you want to test with remote MCP hosts like ChatGPT or Claude.ai.
```
qchuchu, line 25: This is actually not true for Claude. I'd rather point out to the docs for testing app : https://docs.skybridge.tech/quickstart/test-your-app

examples/times-up/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+Modify files in `server/` and refresh the connection with your testing MCP Client to see the changes.
+
+### Testing your App
+
+You can test your App locally by using our DevTools UI on `http://localhost:3000` while running the dev command.
+
+To test your app with other MCP Clients like ChatGPT, Claude or VSCode, see [Testing Your App](https://docs.skybridge.tech/quickstart/test-your-app).
+
+## Deploy to Production
```
qchuchu, line 101: FYI @harijoe we'll need to also adapt this part (on the create PR github actions that @paulleseute created) so that people can use the alpic cli ;)

docs/showcase/times-up.mdx
```diff
@@ -0,0 +1,40 @@
+---
+title: Time's Up
+description: Word-guessing party game for MCP Apps
+---
+
+A word-guessing party game where the user sees a secret word and gives hints to the AI, which tries to guess it. Features hidden metadata, PiP display mode, multilingual support, and widget-to-tool communication.
```
qchuchu, line 6: I think maybe not "hidden metadata" - but leverage _meta attribute that is hidden from the model context : https://developers.openai.com/apps-sdk/reference#tool-results

https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx#2-tool-result-via-uinotificationstool-result-notification

examples/times-up/README.md
paulleseute: yes, this is actually what we are doing below in the readme. i'm gonna fix it for all examples in another pr

PR #518, opened 2026-02-25
Poc/claude harness

thleonard:
image

image


packages/devtools/src/server/claude-session-server.ts
```diff
@@ -0,0 +1,198 @@
...
+      {
+        encoding: "utf-8",
+        timeout: 10000,
+        cwd,
+        env: buildEnv({}),
+      },
+    );
+  } catch {
+    // ignore
+  }
```
harijoe, line 148: **[High] `claude mcp add` mutates global Claude config permanently and never cleans up.**

Every `pnpm dev --claude` run appends entries to the user's global Claude config. Running it 10 times = 10 duplicate entries. More importantly, there's no corresponding `claude mcp remove` on exit, so entries persist after the project is deleted and pollute the user's global toolset indefinitely.

Two fixes needed:

1. **Use project scope** — pass `--scope project` to `claude mcp add`. This writes to a `.claude/` directory relative to `cwd` instead of the user's global config, which is the right mental model for a per-project dev tool.

2. **Clean up on exit** — expose a `cleanup()` method on `ClaudeSessionServer` that runs `claude mcp remove <name>` for each registered server. Call it from `dev-middleware.ts` inside a `process.on('SIGINT')` / `process.on('SIGTERM')` handler.

As a safety net, check `claude mcp list` before registering to skip entries that already exist with the same URL — makes restarts safe even if cleanup didn't run after a crash.

packages/core/src/server/claude-session-server.ts
```diff
@@ -0,0 +1,198 @@
...
+    send(ws, { type: "exit", code: exitCode });
+  });
+
+  return proc;
+}
+
+export function createClaudeSessionServer(
+  options: ClaudeSessionServerOptions = {},
+): ClaudeSessionServer {
+  const port = options.port ?? 3001;
```
harijoe: **[Medium] Port 3001 is hardcoded — will crash with `EADDRINUSE` if anything else uses it.**

`new WebSocketServer({ port: 3001 })` throws immediately if the port is taken. The fallback in `devtoolsStaticServer.ts` (`?? 3001`) means the browser will silently try to connect to the wrong port rather than surfacing the real error.

Use dynamic port allocation instead — bind to port 0 and let the OS assign a free one:

```ts
async function findFreePort(preferred: number): Promise<number> {
  return new Promise((resolve) => {
    const server = net.createServer();
    server.listen(preferred, () => {
      const { port } = server.address() as net.AddressInfo;
      server.close(() => resolve(port));
    });
    server.on("error", () => {
      const fallback = net.createServer();
      fallback.listen(0, () => {
        const { port } = fallback.address() as net.AddressInfo;
        fallback.close(() => resolve(port));
      });
    });
  });
}
```

Make `createClaudeSessionServer` async, call `findFreePort(3001)` before constructing the `WebSocketServer`, and log the resolved port clearly so developers know where to look.

packages/core/package.json
```diff
@@ -52,6 +52,8 @@
   },
   "dependencies": {
     "@babel/core": "^7.29.0",
+    "node-pty": "1.2.0-beta.12",
```
harijoe: **[Medium] `node-pty` should not be a regular dependency of `packages/core`.**

`node-pty` is a native addon — it runs `node-gyp` on install and requires Python + a C++ toolchain. Putting it in `dependencies` of the core framework package means every consumer of `skybridge` compiles it, even if they never use `--claude`. This breaks in CI environments without build tools, in containerised builds, and in constrained environments like Stackblitz.

The PTY logic belongs in `packages/devtools`, not `packages/core`:
- Move `claude-session-server.ts` to `packages/devtools/src/server/`
- Move `node-pty` (and the `ws` dep that's only used by the session server) to `packages/devtools/dependencies`
- Mark it `optionalDependencies` there so install failures don't block devtools in environments that don't need the PTY feature
- In `dev-middleware.ts`, import via `@skybridge/devtools` which is already a peer dep of core

packages/devtools/src/server/claude-session-server.ts
```diff
@@ -0,0 +1,198 @@
...
+      } catch {
+        // ignore
+      }
+      proc = null;
+    });
+  });
+
+  return {
+    port,
+    close() {
```
harijoe, line 258: **[Medium] `close()` does not kill the spawned PTY process, and it's never called anyway.**

Two problems here:

1. `wss.close()` shuts down the WebSocket server but the `proc` variable is scoped inside the `connection` callback closure — `close()` cannot reach it. Orphaned `claude` processes accumulate across dev restarts, each holding a PTY fd open.

2. `close()` is never called anywhere. `dev-middleware.ts` has no shutdown hook, and `express.ts` registers no `SIGINT`/`SIGTERM` handler.

Suggested fix:
- Lift `proc` to be a property on the server object so `close()` can call `proc.kill()`
- Have `setupDevMiddleware` return a `shutdown(): Promise<void>` that sequences: send a close message to WS clients → SIGTERM the PTY with a 2s SIGKILL fallback → close SSE connections on `ScreenshotBridge` → call `wss.close()`
- Register `process.on('SIGINT', shutdown)` / `process.on('SIGTERM', shutdown)` in `express.ts` when the harness is active

packages/core/src/server/devtools-mcp-server.ts
```diff
@@ -0,0 +1,60 @@
...
+  Router,
+} from "express";
+import type { ScreenshotBridge } from "./screenshot-bridge.js";
+
+export function createDevtoolsMcpRouter(bridge: ScreenshotBridge): Router {
+  const router = Router();
+
+  router.post("/", async (req: Request, res: Response, _next: NextFunction) => {
+    try {
+      const server = new McpServer({
```
harijoe: **[Low–Medium] Two issues: new `McpServer` per request, and devtools MCP is on the same port as the user's app.**

**Per-request instantiation:** A new `McpServer` is created on every HTTP POST. For a single-tool dev server this is harmless today, but it's inconsistent with how the main app MCP server works (one instance, reused per request). Move instantiation outside the handler.

**Shared port with the user's app:** `/mcp-devtools` is mounted on the same Express app as `/mcp`. This means:
- Any middleware the user adds to their app (e.g. auth) applies inconsistently — depending on mount order, it may or may not cover `/mcp-devtools`
- If the user adds auth to `/mcp`, they'd reasonably expect it to protect all MCP surfaces, but `/mcp-devtools` is mounted in `dev-middleware.ts` before their custom middleware runs (see `express.ts` line 31–35)

Recommended fix: spin up a dedicated `http.createServer()` on a dynamically allocated port for the devtools MCP surface (reuse the `findFreePort` utility from the port hardcoding fix). This fully decouples it from the user's app. The `devtoolsMcpUrl` is already a string passed around, so the change is localised to `dev-middleware.ts`.

harijoe (commented):
Overall the architecture is solid and the screenshot bridge idea is genuinely clever. Five issues below worth addressing before merging.

packages/devtools/src/server/claude-session-server.ts
thleonard: The claude process is already killed when the browser is closed as it is a child of the main process

PR #531, opened 2026-03-05
feat(core): add cache-busting version param to apps-sdk widget output…

### LINKED ISSUE #530: Allow versioning of view Resource URIs
Hi,

I recently came across a caching issue. I release a new version of the app, go to [chatgpt.com](<http://chatgpt.com>) to hit the refresh button, and it still uses the old template.

According to the OpenAI [docs](<https://developers.openai.com/apps-sdk/build/mcp-server/#step-1--register-a-component-template>), a suggested best practice is to treat the view resource URI as a cache key and version the URI on every update. The example from the docs is:

```
// Old
contents: [{ uri: "ui://view/kanban-board.html" /* ... */ }];

// New
contents: [{ uri: "ui://view/kanban-board-v2.html" /* ... */ }];
```

But something like this would also work:

```
contents: [{ uri: "ui://view/kanban-board.html?v=123" /* ... */ }];
```

I’ve seen that in `packages/core/src/server/server.ts` we have:<br>[https://github.com/alpic-ai/skybridge/blob/1eacc9db5cd3b0897a9161bcb518e7c87c16aa7c/packages/core/src/server/server.ts#L300](<https://github.com/alpic-ai/skybridge/blob/1eacc9db5cd3b0897a9161bcb518e7c87c16aa7c/packages/core/src/server/server.ts#L300>)

So right now, this cannot be achieved.

One simple solution would be to give users the option to enable versioning. This could be done by exposing an optional parameter to specify the version when registering the widget. What do you think?

I would be happy to submit a pr to add this feature

slackerzz:
this will close #530 

as suggest by @qchuchu I added a cache-busting parameter to the outputTemplate uri

<h3>Greptile Summary</h3>

This PR adds a `?v=${Date.now()}` cache-busting query parameter to the `apps-sdk` widget URI in `registerWidget`, so that each server startup produces a new `openai/outputTemplate` value, preventing stale widget content from being served to clients that cache the URI.

**Changes:**
- `server.ts`: The `widgetConfig.uri` for the `apps-sdk` host now appends `?v=${Date.now()}` at widget-registration time. Because the same URI value is assigned to both the MCP resource registration and the `openai/outputTemplate` tool metadata, the two stay in sync.
- `widget.test.ts`: The existing `outputTemplate` assertion is updated to expect the versioned URI. A `vi.spyOn(Date, "now").mockReturnValue(now)` is introduced to make the timestamp deterministic in the test.

**Issue found:**
- The `vi.spyOn(Date, "now")` mock in the new test is **not restored** by the existing `afterEach(() => vi.clearAllMocks())`. `clearAllMocks` resets call history only — it does not restore the original `Date.now` implementation. Any tests added after this one in the same describe block will receive the frozen timestamp instead of the real current time. The `afterEach` should be updated to call `vi.restoreAllMocks()` as well.

- Safe to merge after fixing the test spy cleanup; the production code change itself is correct.
- The core change in `server.ts` is minimal and correct — a single-line addition of a timestamp query parameter that is consistently applied to both the resource URI and the tool metadata. The only real concern is in the test: the `vi.spyOn(Date, "now")` mock is not restored by `vi.clearAllMocks()`, which could silently break future tests. This is a logic-level test infrastructure issue, not a production bug, hence a score of 4 rather than 5.
- `packages/core/src/test/widget.test.ts` — the `Date.now` spy introduced in the new test needs proper teardown via `vi.restoreAllMocks()`.

<sub>Last reviewed commit: d81ccb7</sub>


slackerzz:
Hi @qchuchu , have you managed to test this workaround?

harijoe:
Hi @slackerzz, thanks for your contribution,

We discussed this internally, and we’d like to suggest a little improvement before merging this: instead of using the date, could you compute the hash of the html template? It would avoid bursting the cache on every list tool call.

PR #551, opened 2026-03-11
feat: add infrastructure

paulleseute:
So we can deal with the skybridge.tech domain directly within the repo. I used a single construct for all records because splitting them seemed a bit premature since there are only a few. Infra is deployed on release on a dedicated AWS account.

### DIFF
--- a/infrastructure/lib/constructs/SkybridgeRecords.ts
+++ b/infrastructure/lib/constructs/SkybridgeRecords.ts
@@ -1,10 +1,5 @@
-import {
-  ARecord,
-  CnameRecord,
-  PublicHostedZone,
-  RecordTarget,
-  TxtRecord,
-} from "aws-cdk-lib/aws-route53";
+import { CnameRecord, PublicHostedZone } from "aws-cdk-lib/aws-route53";
+import { HttpsRedirect } from "aws-cdk-lib/aws-route53-patterns";
 import { Construct } from "constructs";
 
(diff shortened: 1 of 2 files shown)


infrastructure/lib/constructs/SkybridgeRecords.ts
```diff
@@ -0,0 +1,73 @@
...
+      domainName: "cname.mintlify-dns.com",
+    });
+    new CnameRecord(this, "DocsStaging", {
+      zone: hostedZone,
+      recordName: "docs-staging",
+      domainName: "cname.mintlify-dns.com",
+    });
+
+    // Vercel (www)
+    new CnameRecord(this, "Www", {
```
qchuchu: Do we still need them ? I thought it was the old docusaurus doc

infrastructure/lib/constructs/SkybridgeRecords.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export class SkybridgeRecords extends Construct {
+  constructor(scope: Construct, id: string, { domain }: SkybridgeRecordsProps) {
+    super(scope, id);
+
+    const hostedZone = new PublicHostedZone(this, "HostedZone", {
+      zoneName: domain,
+    });
+
+    new ARecord(this, "Apex", {
```
qchuchu: Do you know what this one is for ?

PR #566, opened 2026-03-16
Use directly Ext-Apps App as dependency

qchuchu:
<h3>Greptile Summary</h3>

This PR promotes `@modelcontextprotocol/ext-apps` from a dev dependency to a runtime dependency (bumping `^1.2.0` → `^1.2.2`) and replaces the hand-rolled JSON-RPC protocol layer in `McpAppBridge` with the `App` class provided by the SDK. The result is a significant reduction in code — the bespoke message routing, pending-request map, timeout management, and `ResizeObserver`-based size-notification logic are all deleted in favour of equivalent SDK behaviour.

Key changes:
- `McpAppBridge` now wraps `App` and delegates connection, teardown, and all outbound requests to the SDK. Event handlers (`ontoolinput`, `ontoolresult`, `ontoolcancelled`, `onhostcontextchanged`, `onteardown`) are mapped directly onto the existing context-update mechanism.
- `McpAppBridge.resetInstance()` is now `async` (because `cleanup()` awaits `app.close()`), and all test `afterEach` hooks have been correctly updated to `async` / `await` accordingly.
- Test mocks gained `source: window.parent` on every `MessageEvent`, which is required by the SDK v1.2.2 to validate message origin.
- The `containerDimensions` shape in layout tests lost the `height` field, reflecting a type change in the updated SDK.
- Two tests that exercised internal implementation details (request timeout, manual size-changed notification) were removed, since those behaviours are now the SDK's responsibility.

- This PR is safe to merge; it is a well-contained refactor with no logic regressions and correctly updated tests.
- The refactoring is clean and the test suite is correctly updated throughout. Two minor concerns prevent a perfect score: (1) `cleanup()` calling `app.close()` from within the `onteardown` handler is re-entrant and relies on the SDK being tolerant of this; and (2) `use-mcp-app-context.test.ts` is the only test file that does not call `resetInstance()` in `afterEach`, creating a slight asymmetry that could cause issues if the test suite grows.
- `packages/core/src/web/bridges/mcp-app/bridge.ts` — verify that `App.close()` is safe to call from within an `onteardown` callback per the SDK docs.

<details><summary><h3>Comments Outside Diff (5)</h3></summary>

1. `packages/core/src/web/hooks/use-open-external.test.ts`, line 62-65 ([link](https://github.com/alpic-ai/skybridge/blob/e08ace96171d38b197d35cd86b56ba66e9ad823b/packages/core/src/web/hooks/use-open-external.test.ts#L62-L65)) 

   **Unawaited async `resetInstance()`**

   `McpAppBridge.resetInstance()` is now `async` (returns `Promise<void>`) since this PR changed the method signature in `bridge.ts:123`. Without `await`, the cleanup (including `this.app.close()`) may not complete before the next test starts, leading to potential test pollution or flaky tests. The `use-mcp-app-context.test.ts` file was correctly updated to use `await`, but this call site was missed.

   The same issue also exists in unchanged files (`data-llm.test.tsx:171`, `create-store.test.ts:129`, `use-user.test.ts:74`, `use-tool-info.test.ts:100`) that should be fixed as well.

2. `packages/core/src/web/hooks/use-layout.test.ts`, line 77-81 ([link](https://github.com/alpic-ai/skybridge/blob/e08ace96171d38b197d35cd86b56ba66e9ad823b/packages/core/src/web/hooks/use-layout.test.ts#L77-L81)) 

   **Unawaited async `resetInstance()`**

   Same issue as in `use-open-external.test.ts` — `McpAppBridge.resetInstance()` is now async but isn't being awaited here. The `afterEach` callback needs to be `async` and `await` the call.

3. `packages/core/src/web/bridges/mcp-app/bridge.ts`, line 42-45 ([link](https://github.com/alpic-ai/skybridge/blob/3a4af4d1256f909d4f2114bd3c9f36cd3a9e1614/packages/core/src/web/bridges/mcp-app/bridge.ts#L42-L45)) 

   **`cleanup()` not awaited in `onteardown`**

   `cleanup()` is declared `async` and its meaningful async work — `await this.app.close()` — runs after `onteardown` has already returned `{}` and the SDK has sent the teardown response back to the host. While the synchronous parts (`listeners.clear()` and neutralizing `sendSizeChanged`) correctly fire before the return, the transport close races with any subsequent SDK teardown housekeeping.

   Consider making `onteardown` async so the SDK can wait for the full cleanup before finalising the teardown (if the ext-apps SDK supports an async callback here):

   

   If the SDK requires a synchronous callback, at least add a comment explaining that `app.close()` fires after the teardown response is already sent, so that future readers understand the intentional fire-and-forget pattern.

4. `packages/core/src/web/bridges/mcp-app/use-mcp-app-context.test.ts`, line 17-20 ([link](https://github.com/alpic-ai/skybridge/blob/213f6e6f15d5c2647df490fdcbdb63ef2392afd7/packages/core/src/web/bridges/mcp-app/use-mcp-app-context.test.ts#L17-L20)) 

   **Missing `resetInstance` in `afterEach`**

   `afterEach` does not call `await McpAppBridge.resetInstance()`, unlike every other test file changed in this PR (e.g. `use-layout.test.ts`, `use-tool-info.test.ts`, `use-user.test.ts`, `create-store.test.ts`). The bridge instance created during the test is only torn down by the *next* test's `beforeEach`. While this is harmless for a single-test suite, it is asymmetric and could silently bite if new tests are added.

5. `packages/core/src/web/bridges/mcp-app/bridge.ts`, line 40-43 ([link](https://github.com/alpic-ai/skybridge/blob/213f6e6f15d5c2647df490fdcbdb63ef2392afd7/packages/core/src/web/bridges/mcp-app/bridge.ts#L40-L43)) 

   **`app.close()` called from within `onteardown`**

   `cleanup()` calls `this.app.close()`. When `onteardown` is triggered by the host, the SDK is already in the middle of processing the teardown request. Calling `close()` re-entrantly may attempt to send a disconnect notification to a host that has already initiated the teardown, or trigger internal SDK state transitions that are not designed to happen during an active teardown handler.

   The `.catch(() => {})` in `cleanup()` ensures errors are silently swallowed, which limits the practical blast radius, and the pre-emptive `sendSizeChanged` neutralisation is a good defensive measure. That said, it may be worth checking whether the ext-apps `App.close()` is documented as safe to call from within an `onteardown` callback to confirm this is intentional.

</details>

`````markdown
Path: packages/core/src/web/bridges/mcp-app/use-mcp-app-context.test.ts
Line: 17-20

Comment:
**Missing `resetInstance` in `afterEach`**

`afterEach` does not call `await McpAppBridge.resetInstance()`, unlike every other test file changed in this PR (e.g. `use-layout.test.ts`, `use-tool-info.test.ts`, `use-user.test.ts`, `create-store.test.ts`). The bridge instance created during the test is only torn down by the *next* test's `beforeEach`. While this is harmless for a single-test suite, it is asymmetric and could silently bite if new tests are added.

```suggestion
  afterEach(async () => {
    await McpAppBridge.resetInstance();
    vi.unstubAllGlobals();
    vi.clearAllMocks();
  });
```

---

Path: packages/core/src/web/bridges/mcp-app/bridge.ts
Line: 40-43

Comment:
**`app.close()` called from within `onteardown`**

`cleanup()` calls `this.app.close()`. When `onteardown` is triggered by the host, the SDK is already in the middle of processing the teardown request. Calling `close()` re-entrantly may attempt to send a disconnect notification to a host that has already initiated the teardown, or trigger internal SDK state transitions that are not designed to happen during an active teardown handler.

The `.catch(() => {})` in `cleanup()` ensures errors are silently swallowed, which limits the practical blast radius, and the pre-emptive `sendSizeChanged` neutralisation is a good defensive measure. That said, it may be worth checking whether the ext-apps `App.close()` is documented as safe to call from within an `onteardown` callback to confirm this is intentional.

`````

</details>

<sub>Last reviewed commit: 213f6e6</sub>

### DIFF
--- a/packages/core/src/web/bridges/mcp-app/bridge.ts
+++ b/packages/core/src/web/bridges/mcp-app/bridge.ts
@@ -37,11 +37,6 @@ export class McpAppBridge implements Bridge<McpAppContext> {
       this.updateContext(params);
     };
 
-    this.app.onteardown = async () => {
-      await this.cleanup();
-      return {};
-    };
-
     this.connectPromise = this.connect();
   }
 
@@ -110,17 +105,13 @@ export class McpAppBridge implements Bridge<McpAppContext> {
     return this.context[key];
   }
 
-  public cleanup = async () => {
+  public cleanup = () => {
     this.listeners.clear();
-    // Neutralize sendSizeChanged so any pending requestAnimationFrame callbacks
-    // from the SDK's auto-resize become no-ops after close.
-    this.app.sendSizeChanged = () => Promise.resolve();
-    await this.app.close().catch(() => {});
   };
 
-  public static async resetInstance(): Promise<void> {
+  public static resetInstance(): void {
     if (McpAppBridge.instance) {
-      await McpAppBridge.instance.cleanup();
+      McpAppBridge.instance.cleanup();
       McpAppBridge.instance = null;
     }
   }
--- a/packages/core/src/web/bridges/mcp-app/use-mcp-app-context.test.ts
+++ b/packages/core/src/web/bridges/mcp-app/use-mcp-app-context.test.ts
@@ -11,7 +11,7 @@ describe("useMcpAppContext", () => {
   beforeEach(async () => {
     vi.stubGlobal("skybridge", { hostType: "mcp-app" });
     vi.stubGlobal("ResizeObserver", MockResizeObserver);
-    await McpAppBridge.resetInstance();
+    McpAppBridge.resetInstance();
   });
 
   afterEach(() => {
--- a/packages/core/src/web/create-store.test.ts
+++ b/packages/core/src/web/create-store.test.ts
@@ -1,3 +1,4 @@
+import { cleanup } from "@testing-library/react";
 import {
   afterEach,
   beforeEach,
@@ -126,7 +127,8 @@ describe("createStore", () => {
     });
 
     afterEach(async () => {
-      await McpAppBridge.resetInstance();
+      cleanup();
+      McpAppBridge.resetInstance();
       McpAppAdaptor.resetInstance();
     });
 
--- a/packages/core/src/web/data-llm.test.tsx
+++ b/packages/core/src/web/data-llm.test.tsx
@@ -168,6 +168,7 @@ describe("DataLLM", () => {
     });
 
     afterEach(() => {
+      cleanup();
       McpAppBridge.resetInstance();
       McpAppAdaptor.resetInstance();
     });
--- a/packages/core/src/web/hooks/use-layout.test.ts
+++ b/packages/core/src/web/hooks/use-layout.test.ts
@@ -76,7 +76,7 @@ describe("useLayout", () => {
     afterEach(async () => {
       vi.unstubAllGlobals();
       vi.resetAllMocks();
-      await McpAppBridge.resetInstance();
+      McpAppBridge.resetInstance();
       McpAppAdaptor.resetInstance();
     });
 
--- a/packages/core/src/web/hooks/use-open-external.test.ts
+++ b/packages/core/src/web/hooks/use-open-external.test.ts
@@ -61,7 +61,7 @@ describe("useOpenExternal", () => {
     afterEach(async () => {
       vi.unstubAllGlobals();
       vi.resetAllMocks();
-      await McpAppBridge.resetInstance();
+      McpAppBridge.resetInstance();
     });
 
     it("should return a function that sends ui/open-link request to the MCP host", async () => {
--- a/packages/core/src/web/hooks/use-tool-info.test.ts
+++ b/packages/core/src/web/hooks/use-tool-info.test.ts
@@ -97,7 +97,7 @@ describe("useToolInfo", () => {
     afterEach(async () => {
       vi.unstubAllGlobals();
       vi.resetAllMocks();
-      await McpAppBridge.resetInstance();
+      McpAppBridge.resetInstance();
       McpAppAdaptor.resetInstance();
     });
 
--- a/packages/core/src/web/hooks/use-user.test.ts
+++ b/packages/core/src/web/hooks/use-user.test.ts
@@ -71,7 +71,7 @@ describe("useUser", () => {
     afterEach(async () => {
       vi.unstubAllGlobals();
       vi.resetAllMocks();
-      await McpAppBridge.resetInstance();
+      McpAppBridge.resetInstance();
       McpAppAdaptor.resetInstance();
     });
 
--- a/packages/core/vitest.config.ts
+++ b/packages/core/vitest.config.ts
@@ -1,9 +1,20 @@
 import { defineConfig } from "vitest/config";
 
+// Noise from @modelcontextprotocol/ext-apps PostMessageTransport & Protocol
+// that leaks into test output during setup/teardown.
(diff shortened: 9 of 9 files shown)

### CODE REVIEW

File: packages/core/src/web/bridges/mcp-app/bridge.ts [hunk 1]
```diff
@@ -1,115 +1,88 @@
...
   };
   private listeners = new Map<McpAppContextKey, Set<() => void>>();
-  private pendingRequests = new Map<string | number, PendingRequest<unknown>>();
-  private nextId = 1;
-  private initialized: boolean;
-  private appInitializationOptions: McpUiInitializeRequest["params"];
-  private requestTimeout: number;
-  private cleanupSizeObserver: (() => void) | null = null;
+  private app: App;
+  private transport: PostMessageTransport;
```
valentinbeggi: not sure it needs to be an instance field

packages/core/src/web/bridges/mcp-app/bridge.ts
```diff
@@ -1,115 +1,88 @@
...
-    requestTimeout: number = 10_000,
-  ) {
-    this.requestTimeout = requestTimeout;
-    this.initialized = false;
-    this.appInitializationOptions = {
-      appInfo: options.appInfo,
-      appCapabilities: {},
-      protocolVersion: LATEST_PROTOCOL_VERSION,
+    this.app.onteardown = () => {
+      this.cleanup();
```
valentinbeggi: isn't this async ?

packages/core/package.json
```diff
@@ -52,20 +52,20 @@
   },
   "dependencies": {
     "@babel/core": "^7.29.0",
-    "@oclif/core": "^4.9.0",
+    "@modelcontextprotocol/ext-apps": "^1.2.2",
```
fredericbarthelet, line 55: should be peer

packages/core/package.json
fredericbarthelet, line 46: we can remove this no?

packages/core/package.json
qchuchu: ? why ?

qchuchu: We are indeed using the code underlying and wrapping the API

packages/core/src/web/bridges/mcp-app/use-mcp-app-context.test.ts
```diff
@@ -11,7 +11,7 @@ describe("useMcpAppContext", () => {
   beforeEach(async () => {
     vi.stubGlobal("skybridge", { hostType: "mcp-app" });
     vi.stubGlobal("ResizeObserver", MockResizeObserver);
-    McpAppBridge.resetInstance();
+    await McpAppBridge.resetInstance();
```
fredericbarthelet: why did it became async all of sudden?

packages/core/package.json
qchuchu: -> Seen together, for the moment let's keep it as deps

packages/core/package.json
qchuchu: No bc we need it in the backend

packages/core/src/web/bridges/mcp-app/use-mcp-app-context.test.ts
qchuchu: Removed - I don't close anymore

Issue #535, opened 2026-03-06
Support Cloudflare hosting for skybridge apps


fredericbarthelet:
## Background

The initial plan to support Cloudflare hosting included switching to Hono (hence the original name of the issue).
Following a conversation with @yusukebe showcasing an [example implementation of ext-apps with Hono hosted on Cloudflare](https://github.com/yusukebe/mcp-app-with-hono), it was decided to try and migrate the underlying router of Skybridge from Express to Hono.

## Express and Hono differences

| | Express | Hono |
|---|---|---|
| **Core interface** | `req`/`res` (Node.js `IncomingMessage`/`ServerResponse`) | Web Standard `fetch` (`Request` -> `Response`) |
| **Node.js** | Native | Via `@hono/node-server` adapter |
| **Cloudflare Workers** | Via `httpServerHandler` from `cloudflare:node` | Native (`export default app`) |
| **Bun** | Via Node.js compat layer | Native (`export default app`) |
| **Deno** | Via Node.js compat layer | Native (`Deno.serve(app.fetch)`) |
| **Architecture** | Node.js-first, needs bridges for non-Node runtimes | Fetch-first, needs adapter only for Node.js |

Express is built around Node.js streams and its `http` module. Runtimes like Cloudflare Workers, Bun, and Deno support the Web Standard Fetch API natively, so Express requires compatibility layers to run on them. Hono takes the opposite approach: it is built on the Fetch API, which makes it a natural fit for Workers/Bun/Deno, while using an adapter (`@hono/node-server`) for Node.js.

## Experimentations

- Deploying Skybridge demo *8 ball* app from `npm create skybridge` command as-is to Cloudflare with `wrangler deploy`:
  - ❌ Failed because the codebase has no default export, which means it is assumed to be a Service Worker format Worker.
  - Additionally, dev-only dynamic imports (`@skybridge/devtools`, `widgetsDevServer` which pulls in Vite/Rollup/lightningcss) get bundled by wrangler even though they never run in production, causing native addon resolution errors.

- Deploying with an entry point using the [Express compat layer](https://developers.cloudflare.com/workers/tutorials/deploy-an-express-app/) (`httpServerHandler` from `cloudflare:node`) for Cloudflare Workers:
  - Requires `McpServer.run()` to return the `httpServerHandler` when running in a Workers environment.
  - Requires the template's `server/src/index.ts` to `export default await server.run()`.
[...]

PR #568, opened 2026-03-16
Cloudflare hosting phase #1 - Express compat layer

fredericbarthelet:
Addresses #535

This is phase 1 of Cloudflare hosting support.

One big rework not initially planned was to get read of all readFileSync API calls that can't be used in a Cloudflare worker environment. We were relying on those to:
- read handlebars template at runtime -> switch to a [pre-compilation phase](https://handlebarsjs.com/guide/installation/precompilation.html)
- read vite manifest for entrypoint location at runtime -> switch from json file read to js module export file

I created a separate issue to move away from native Node.js API in production build : https://github.com/alpic-ai/skybridge/issues/571

I also had to make 2 additional small changes compared to original plan:
- remove the string literal from the dev only import statements to prevent esbuild from importing those (Vite deps like lightningcss are erroring wrangler build step)
- remove the widgetsDevServer exports from the @skybridge/server barrel file (for the same reason). This is still here for retro-compatibility reason, but can be broken if need be

Run `pnpm create skybridge@cloudflare` and then `npm run cloudflare:deploy` in the generated directory to deploy on Cloudflare

<h3>Greptile Summary</h3>

This PR begins Cloudflare Workers hosting support (phase 1) by adding an Express compatibility layer. The `McpServer.run()` method now starts the HTTP server and then conditionally imports `httpServerHandler` from `cloudflare:node` to return a Workers-compatible fetch handler. In non-Workers environments, it returns `undefined`. Dynamic imports in `express.ts` are indirected through variables to prevent Cloudflare's bundler from statically resolving dev-only dependencies.

- `server.run()` return type changed from `Promise<void>` to `Promise<{ fetch: (...args: unknown[]) => unknown } | undefined>` to expose the Cloudflare Workers fetch interface
- Dev-only `import()` calls in `express.ts` use variable indirection to avoid bundler static analysis
- Starter template now uses `export default await server.run()` for Workers module compatibility
- `widgetsDevServer` removed from public exports in `index.ts` — **this is a breaking change** as documentation (`docs/devtools/index.mdx`) still references importing it from `skybridge/server`

- This PR is generally safe to merge; the Cloudflare integration is well-implemented with a graceful fallback, but contains a breaking export removal that needs documentation alignment.
- The Cloudflare Workers integration follows the documented pattern correctly and degrades gracefully in non-Workers environments. The dynamic import indirection is a proven bundler technique. Score is 4 instead of 5 due to the removal of the `widgetsDevServer` public export without updating the documentation that references it, which will break users following the "Custom Integration" guide.
- `packages/core/src/server/index.ts` — removed `widgetsDevServer` export still referenced in documentation

1. `packages/core/src/server/index.ts`, line 10 ([link](https://github.com/alpic-ai/skybridge/blob/25e58e21343444869f4de85fe07e443c99b1d8c7/packages/core/src/server/index.ts#L10)) 

   **Breaking public export removal**

   The `widgetsDevServer` export was removed here, but `docs/devtools/index.mdx` (line 81) still documents it as part of the "Custom Integration" guide:

   ```typescript
   import { devtoolsStaticServer, widgetsDevServer } from "skybridge/server";
   ```

   Users following that documentation will get an import error after this change. The docs should be updated to remove or replace this reference — or the export should be kept if it's still intended for external use.

</details>

`````markdown
Path: packages/core/src/server/index.ts
Line: 10

Comment:
**Breaking public export removal**

The `widgetsDevServer` export was removed here, but `docs/devtools/index.mdx` (line 81) still documents it as part of the "Custom Integration" guide:

```typescript
import { devtoolsStaticServer, widgetsDevServer } from "skybridge/server";
```

Users following that documentation will get an import error after this change. The docs should be updated to remove or replace this reference — or the export should be kept if it's still intended for external use.

`````

</details>

<sub>Last reviewed commit: 25e58e2</sub>


harijoe:
Has been implemented in #731

PR #583, opened 2026-03-18
feat: enable remote hmr

paulleseute:
Vite HMR ran on a dedicated port (24678), separate from the Express server. This meant HMR only worked on the machine running the server: testing on a on any other device (mobile) had no hot reload.

Instead of Vite creating its own WebSocket server on a separate port, we share the http.Server instance that Express already uses. Vite attaches its HMR upgrade handler to the same server via `hmr: { server: httpServer }`. The http.Server is created in McpServer.run() before the Express app, passed through createServer() → widgetsDevServer(), and used for both HTTP requests and WebSocket upgrades on the same port. Since HMR now goes through the same origin as the server, it works through the tunnel that's already required for testing on ChatGPT/Claude. The CSP connect_domains now includes both the https:// and wss:// variants of the server URL, so the HMR WebSocket is allowed by ChatGPT/Claude CSP.

Additionally, the --use-forwarded-host flag has been removed. The server now auto-detects tunnel usage by checking for the X-Forwarded-Host header on each request, and derives the protocol from X-Forwarded-Proto. This means skybridge dev just works locally and through a tunnel with no extra flags.

Can be tested with ngrok or alpic tunnel (staging).

### DIFF
--- a/packages/core/src/server/server.ts
+++ b/packages/core/src/server/server.ts
@@ -628,7 +628,9 @@ export class McpServer<
         } else if (origin) {
           serverUrl = origin;
         } else if (host) {
-          const proto = ["127.0.0.1:", "localhost:"].some((p) => host.startsWith(p))
+          const proto = ["127.0.0.1:", "localhost:"].some((p) =>
+            host.startsWith(p),
+          )
             ? "http"
             : "https";
           serverUrl = `${proto}://${host}`;
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -44,7 +44,7 @@
   "license": "ISC",
   "peerDependencies": {
     "@modelcontextprotocol/sdk": ">=1.0.0",
-    "@skybridge/devtools": ">=0.26.1",
+    "@skybridge/devtools": ">=0.35.8 <1.0.0",
     "nodemon": ">=3.0.0",
     "react": ">=18.0.0",
     "react-dom": ">=18.0.0",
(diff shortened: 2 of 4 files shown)


packages/core/src/server/server.ts
```diff
@@ -396,22 +395,24 @@ export class McpServer<
 
   async run(): Promise<void> {
     this.applyMcpMiddleware();
+    const httpServer = http.createServer();
+
     if (!this.express) {
       this.express = await createServer({
```
valentinbeggi: rename createServer -> createApp ? it takes 2 server as argument
we could also rename the server param mcpServer maybe

packages/core/src/server/server.ts
```diff
@@ -640,11 +642,8 @@ export class McpServer<
 
         const connectDomains = [serverUrl];
         if (!isProduction) {
-          const hmrPort = process.env.__SKYBRIDGE_HMR_PORT ?? DEFAULT_HMR_PORT;
-          const VITE_HMR_WEBSOCKET_DEFAULT_URL = `ws://localhost:${hmrPort}`;
-          connectDomains.push(VITE_HMR_WEBSOCKET_DEFAULT_URL);
+          connectDomains.push(serverUrl.replace(/^http/, "ws"));
```
valentinbeggi: maybe use URL constructor here instead ?

paulleseute: good call

Issue #514, opened 2026-02-24
Update Apps SDK native types for requestModal


fredericbarthelet:
`window.openai.requestModal({ params, template })` supports `template` values as URI of other resources describing templates on the same app. The current types of native `requestModal` in bridge do not allow this optional parameter.
Somebody relying on the native implementation to perform rendering of another template will face type mismatch.

PR #586, opened 2026-03-21
fix(core): support requestModal template option

Jean-crack:
## Summary

Add support for the optional `template` field in `RequestModalOptions` so Skybridge's native bridge types match the Apps SDK API.

## Changes

- add `template?: string` to `RequestModalOptions`
- add a test covering `requestModal({ params, template })`

## Testing

- ran targeted unit test for `use-request-modal`

Fixes alpic-ai/skybridge#514

<h3>Greptile Summary</h3>

This PR adds the optional `template?: string` field to `RequestModalOptions` in `types.ts` and adds a unit test covering the `requestModal({ params, template })` call path. The change aligns Skybridge's native bridge types with the Apps SDK API contract.

The implementation is clean and minimal:
- The `useRequestModal` hook passes the options object directly to `adaptor.openModal(opts)` — no hook changes are needed.
- The `apps-sdk` adaptor forwards the entire options object to `window.openai.requestModal(options)`, so `template` will be correctly sent to the host.
- The `mcp-app` adaptor's `openModal` only copies `options.params` into `_viewState` and ignores `template`. Since `HostContext.view` doesn't define a `template` field, there is currently no way to surface it back to the widget via `useRequestModal`. This is a minor gap worth tracking as a follow-up if MCP App ever needs to simulate template-based modal rendering, but it does not affect the primary Apps SDK integration path targeted by this fix.

- This PR is safe to merge — it's a small, additive type change with a targeted test and no breaking changes.
- The change is a one-line type addition to an existing optional-field struct. The hook and both adaptors accept the full options object, so no downstream plumbing changes are required. The new test exercises the exact scenario introduced. The only minor gap (MCP App adaptor ignoring `template` in `_viewState`) is a non-blocking follow-up and does not affect the primary Apps SDK path.
- No files require special attention.

<sub>Last reviewed commit: ["fix(core): support r..."](https://github.com/alpic-ai/skybridge/commit/5ec0568cbf3e572fd5aab449888097eecb923d8c)</sub>


packages/core/src/web/hooks/use-request-modal.test.ts
```diff
@@ -68,4 +68,18 @@ describe("useRequestModal", () => {
     expect(requestModalMock).toHaveBeenCalledTimes(1);
     expect(requestModalMock).toHaveBeenCalledWith(options);
   });
+
+  it("should support template URIs in requestModal options", () => {
```
fredericbarthelet: This test is redundant, you should add the template parameter in previous test `should call window.openai.requestModal with the options when open is called`

PR #602, opened 2026-03-24
feat: make create-skybridge skill step agent friendly

paulleseute:
by properly supporting non-interactive mode

### DIFF
--- a/packages/create-skybridge/src/index.ts
+++ b/packages/create-skybridge/src/index.ts
@@ -193,7 +193,7 @@ export async function init(args: string[] = process.argv.slice(2)) {
         "add",
         "alpic-ai/skybridge",
         "-s",
-        "skybridge",
+        "chatgpt-app-builder",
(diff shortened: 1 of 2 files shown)


packages/create-skybridge/src/index.ts
```diff
@@ -182,21 +183,24 @@ export async function init(args: string[] = process.argv.slice(2)) {
...
+  }
+
+  if (skill) {
+    run(
+      [
+        ...getPkgExecCmd(pkgManager, "skills"),
+        "add",
+        "alpic-ai/skybridge",
+        "-s",
+        "skybridge",
```
qchuchu: why chatgpt-app-builder became skybridge ?

paulleseute: because i believe this was set before we had aliases. i guess it's more consistent to use the skybridge skills on the skybridge template.

PR #606, opened 2026-03-25
feat: support OpenAI file library APIs

### LINKED ISSUE #605: feat: support new OpenAI file library APIs (selectFiles, uploadFile library option)
## Context

OpenAI has added new file-related APIs to the Apps SDK:

### New API
- **`window.openai.selectFiles()`** — opens ChatGPT's file library picker (feature-detect before using) and returns app-authorized files as `[{ fileId, fileName, mimeType }]`.

### Modified APIs
- **`window.openai.uploadFile(file, { library?: boolean })`** — now accepts an options object; pass `{ library: true }` to also save the upload into the user's ChatGPT file library when that library is available.
- **`window.openai.getFileDownloadUrl({ fileId })`** — now works for files uploaded by the widget, files selected via `selectFiles()`, or files provided via tool/file params (previously only uploads or file params).

### State management update
- `imageIds` may include file IDs selected via `selectFiles()` in addition to uploaded file IDs or file params.

## What needs to change in Skybridge

1. **Types** — Update `AppsSdkOpenAI` interface to add `selectFiles()` and the new `uploadFile` options parameter
2. **Adaptor** — Update `AppsSdkAdaptor` to support `selectFiles()` and forward the `library` option on `uploadFile`
3. **Hook** — Extend `useFiles()` to expose `selectFiles` alongside `upload` and `getDownloadUrl`
4. **Adaptor interface** — Update the shared `Adaptor` interface with the new method + updated signature
5. **MCP adaptor** — Add `selectFiles` stub that throws (not supported in MCP Apps)
6. **Docs** — Update `docs/api-reference/use-files.mdx`
7. **Skills** — Update chatgpt-app-builder skill if it references file APIs

## Source

Diff from `alpic-ai/openai-apps-sdk-docs`

qchuchu:
## Summary

- Add `selectFiles()` to `useFiles()` hook — opens ChatGPT's file library picker (with feature detection for hosts that don't support it yet)
- Add `{ library?: boolean }` option to `upload()` — saves uploads to the user's ChatGPT file library
- Extend `FileMetadata` type with optional `fileName` and `mimeType` fields
- Update `getFileDownloadUrl` to document expanded scope (uploads, selected files, tool/file params)
- Extract `trackFileIds` helper in `AppsSdkAdaptor` to DRY widget state updates

Closes #605

## Test plan

- [x] Unit tests pass (`pnpm test`)
- [x] Build passes (`pnpm build`)
- [ ] Manual test in everything example app on ChatGPT (note: `selectFiles` is documented but not yet available in the host runtime)

<h3>Greptile Summary</h3>

This PR extends `useFiles()` with two new capabilities: a `selectFiles()` method that opens ChatGPT's file library picker (with proper feature detection for older host versions) and a `{ library?: boolean }` option on `upload()` to persist files to the user's library. `FileMetadata` is also enriched with optional `fileName` and `mimeType` fields.

Key changes:
- `AppsSdkAdaptor` extracts the repeated widgetState `imageIds` update into a private `trackFileIds` helper used by both `uploadFile` and the new `selectFiles` — a clean DRY improvement
- `McpAppAdaptor` adds a correctly-throwing `selectFiles` stub, consistent with the existing pattern for other unsupported file methods
- New unit tests cover the `library` upload option and `selectFiles` success path; the switch from `vi.resetAllMocks()` to `vi.clearAllMocks()` leaves a dynamically-added `OpenaiMock.selectFiles` property alive across subsequent tests, which could interfere with testing the unsupported-host code path in the future
- Documentation and example app are updated thoroughly

- PR is safe to merge; the one minor concern is a test isolation issue that doesn't affect production behavior.
- The implementation is clean, feature detection is correctly applied, the `trackFileIds` refactor is a net improvement, and the MCP stub follows the established pattern. The only flag is a P2 test hygiene issue: `vi.clearAllMocks()` doesn't remove the `selectFiles` property added in one test, leaving it visible to subsequent tests and making the unsupported-host path untestable without a cleanup step.
- packages/core/src/web/hooks/use-files.test.ts — mock cleanup for the dynamically-added `selectFiles` property

`````markdown
Path: packages/core/src/web/hooks/use-files.test.ts
Line: 45-56

Comment:
**Test mock leaks to sibling tests**

`OpenaiMock.selectFiles` is assigned inside this test body, but `vi.clearAllMocks()` (changed from `vi.resetAllMocks()`) only clears call history — it does not remove properties added to `OpenaiMock`. Because `OpenaiMock` is shared across all tests in the suite, any test that runs *after* this one will observe `window.openai.selectFiles` as defined, even if the test doesn't expect it to be.

In practice this means the "unsupported host" code path in `AppsSdkAdaptor.selectFiles` (the `!window.openai.selectFiles` guard) can never be exercised by a future test without an explicit cleanup step. Consider either:

1. Deleting the property in `afterEach` (or resetting it to `undefined`), or
2. Moving the assignment into `beforeEach` alongside the other mock setup and using `vi.resetAllMocks()` (which would clear the mock implementation) combined with a per-test override.

```suggestion
  it("should select files from ChatGPT", async () => {
    const selectedFiles = [
      { fileId: "file_1", fileName: "doc.pdf", mimeType: "application/pdf" },
    ];
    OpenaiMock.selectFiles = vi.fn().mockResolvedValue(selectedFiles);

    const { result } = renderHook(() => useFiles());

    const files = await result.current.selectFiles();
    expect(OpenaiMock.selectFiles).toHaveBeenCalled();
    expect(files).toEqual(selectedFiles);

    delete OpenaiMock.selectFiles;
  });
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat: support OpenAI file library APIs (..."](https://github.com/alpic-ai/skybridge/commit/916ff75a21217bc169a1dca003abcbd91cb8fe7b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=26282275)</sub>

### DIFF
--- a/packages/core/src/web/bridges/apps-sdk/adaptor.ts
+++ b/packages/core/src/web/bridges/apps-sdk/adaptor.ts
@@ -1,12 +1,14 @@
 import type {
   Adaptor,
   CallToolResponse,
+  FileMetadata,
   HostContext,
   HostContextStore,
   OpenExternalOptions,
   RequestDisplayMode,
   RequestModalOptions,
   SetWidgetStateAction,
+  UploadFileOptions,
 } from "../types.js";
 import { AppsSdkBridge } from "./bridge.js";
 import type { AppsSdkWidgetState } from "./types.js";
@@ -83,23 +85,39 @@ export class AppsSdkAdaptor implements Adaptor {
     });
   };
 
-  public uploadFile = (file: File) => {
-    return window.openai.uploadFile(file).then(async (metadata) => {
-      const state: AppsSdkWidgetState = window.openai.widgetState
-        ? { ...window.openai.widgetState }
-        : { modelContent: {}, privateContent: {} };
-      if (!state.imageIds) {
-        state.imageIds = [];
-      }
-      state.imageIds.push(metadata.fileId);
-      await window.openai.setWidgetState(state);
-      return metadata;
-    });
+  public uploadFile = async (file: File, options?: UploadFileOptions) => {
+    const metadata = await window.openai.uploadFile(file, options);
+    await this.trackFileIds(metadata.fileId);
+    return metadata;
   };
 
   public getFileDownloadUrl = (file: { fileId: string }) => {
     return window.openai.getFileDownloadUrl(file);
   };
+
+  public selectFiles = async (): Promise<FileMetadata[]> => {
+    if (!window.openai.selectFiles) {
+      throw new Error(
+        "selectFiles is not supported by the current host version.",
+      );
+    }
+    const files = await window.openai.selectFiles();
+    if (files.length > 0) {
+      await this.trackFileIds(...files.map((f) => f.fileId));
+    }
+    return files;
+  };
+
+  private async trackFileIds(...fileIds: string[]): Promise<void> {
+    const state: AppsSdkWidgetState = window.openai.widgetState
+      ? { ...window.openai.widgetState }
+      : { modelContent: {}, privateContent: {} };
+    if (!state.imageIds) {
+      state.imageIds = [];
+    }
+    state.imageIds.push(...fileIds);
+    await window.openai.setWidgetState(state);
+  }
 
   public openModal(options: RequestModalOptions) {
     return window.openai.requestModal(options);
--- a/packages/core/src/web/hooks/use-files.test.ts
+++ b/packages/core/src/web/hooks/use-files.test.ts
@@ -53,6 +53,8 @@ describe("useFiles", () => {
     const files = await result.current.selectFiles();
     expect(OpenaiMock.selectFiles).toHaveBeenCalled();
     expect(files).toEqual(selectedFiles);
+
+    delete OpenaiMock.selectFiles;
   });
 
   it("should download a file from ChatGPT", () => {
--- a/packages/core/src/web/hooks/use-files.test.ts
+++ b/packages/core/src/web/hooks/use-files.test.ts
@@ -1,9 +1,10 @@
 import { renderHook } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { AppsSdkAdaptor } from "../bridges/apps-sdk/adaptor.js";
 import { useFiles } from "./use-files.js";
 
 describe("useFiles", () => {
-  const OpenaiMock = {
+  const OpenaiMock: Record<string, unknown> = {
     uploadFile: vi.fn().mockResolvedValue({
       fileId: `sediment://file_abc123`,
     }),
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: packages/core/src/web/bridges/apps-sdk/adaptor.ts [hunk 1]
```diff
@@ -83,24 +85,40 @@ export class AppsSdkAdaptor implements Adaptor {
...
+    if (files.length > 0) {
+      await this.trackFileIds(...files.map((f) => f.fileId));
+    }
+    return files;
+  };
+
+  private async trackFileIds(...fileIds: string[]): Promise<void> {
+    const state: AppsSdkWidgetState = window.openai.widgetState
+      ? { ...window.openai.widgetState }
+      : { modelContent: {}, privateContent: {} };
```
harijoe, line 114: It looks suspicious having to initialize the widget state manually with those keys, especially because we don't use them right after.

packages/core/src/web/bridges/apps-sdk/adaptor.ts
```diff
@@ -83,24 +85,40 @@ export class AppsSdkAdaptor implements Adaptor {
...
+    if (!state.imageIds) {
+      state.imageIds = [];
+    }
+    state.imageIds.push(...fileIds);
```
harijoe, line 118: Why do we track such fileIds in the global state under the key `imageIds`? Isn't it the responsibility of the user to handle those ids?

packages/core/src/web/hooks/use-files.test.ts
```diff
@@ -28,7 +30,31 @@ describe("useFiles", () => {
...
+    OpenaiMock.selectFiles = vi.fn().mockResolvedValue(selectedFiles);
+
+    const { result } = renderHook(() => useFiles());
+
+    const files = await result.current.selectFiles();
+    expect(OpenaiMock.selectFiles).toHaveBeenCalled();
+    expect(files).toEqual(selectedFiles);
+
+    delete OpenaiMock.selectFiles;
   });
```
harijoe, line 58: We could add a test covering the error path when window.openai.selectFiles is undefined, given feature detection is a key part of this feature

harijoe (commented):
Few questions that needs to be addressed before approval

packages/core/src/web/bridges/apps-sdk/adaptor.ts
qchuchu: What do you mean ?

packages/core/src/web/bridges/apps-sdk/adaptor.ts
qchuchu: Bc basically : 

- The structure modelContent / privateContent / imageIds is a context structure proper to OpenAI : https://developers.openai.com/apps-sdk/build/state-management#image-ids-in-widget-state-model-visible-images-chatgpt-extension

Currently, we "help" the users, when using the setWidgetState API, by putting the state inside the modelContent key. 

And we do the upload the "imageIds" for the user, so that the agent has the context of the image.

harijoe (approved):

PR #658, opened 2026-04-08
fix: resolve asset URLs at runtime to support tunnels

paulleseute:
Static assets (images, fonts, etc.) imported in widgets were not available to clients other than the machine running the tunnel. The Vite transform plugin hardcoded `http://localhost:${port}` as the asset origin, so only localhost could reach them.

Asset paths are now rewritten as expressions that read `window.skybridge.serverUrl` at runtime, which already contains the correct origin (tunnel or localhost). We can't use `import.meta.url` because widget iframes render on a sandbox domain (e.g. `web-sandbox.oaiusercontent.com` on ChatGPT), so the module origin doesn't point back to the tunnel.


harijoe (approved):
LGTM

PR #659, opened 2026-04-09
feat(core): add StatsD monitoring for tool invocations

harijoe:
### Summary

- Add StatsD monitoring middleware that emits a `Requests_1` counter over UDP for every `tools/call`
- Respects existing telemetry opt-out (`SKYBRIDGE_TELEMETRY_DISABLED`, `DO_NOT_TRACK`, `skybridge telemetry disable`)
- Provision a `t4g.nano` EC2 instance with CloudWatch Agent as the StatsD receiver at `metrics.skybridge.tech:8125`
- Assign an Elastic IP so the endpoint address is stable across instance replacements

### Architecture Notes

- **Metric as internal middleware** — injected as the outermost `McpMiddlewareEntry` in `applyMcpMiddleware()`, no public API surface added
- **Lazy UDP socket** — created on first send, `unref()`'d so it never blocks process exit
- **`STATSD_HOST` placeholder** — left empty until the first `cdk deploy` outputs the Elastic IP; no packets sent while blank

<h3>Greptile Summary</h3>

This PR adds end-to-end StatsD monitoring for MCP tool invocations: a new `monitoring.ts` middleware that emits counters over UDP, and a companion CDK construct that provisions the receiving EC2 instance (`metrics.skybridge.tech:8125`) forwarding to CloudWatch. All three concerns from the prior review round (placeholder hostname, shared cold-start flag, registration-time telemetry check) have been resolved.

Safe to merge — all prior P1 concerns are resolved; remaining findings are P2 style/operational suggestions.

The three blocking issues flagged in the previous review round have each been addressed. The only remaining notes are P2: a missing SSM managed policy on the EC2 role (makes the instance hard to debug but doesn't affect runtime behavior of the SDK) and a module-level UDP socket that is never explicitly closed (mitigated by socket.unref()). Neither blocks merge.

infrastructure/lib/constructs/Monitoring.ts — consider adding AmazonSSMManagedInstanceCore to the IAM role for operational access.

`````markdown
Path: infrastructure/lib/constructs/Monitoring.ts
Line: 36-41

Comment:
**Missing SSM managed policy for instance access**

The role only has `CloudWatchAgentServerPolicy`. There's no SSH ingress rule and no `AmazonSSMManagedInstanceCore`, which means there's no way to connect to this instance for debugging (no SSH, no SSM Session Manager). If the CloudWatch Agent fails to start or the user data script errors out, the instance is a black box.

```suggestion
    const role = new Role(this, "Role", {
      assumedBy: new ServicePrincipal("ec2.amazonaws.com"),
      managedPolicies: [
        ManagedPolicy.fromAwsManagedPolicyName("CloudWatchAgentServerPolicy"),
        ManagedPolicy.fromAwsManagedPolicyName("AmazonSSMManagedInstanceCore"),
      ],
    });
```

---

Path: packages/core/src/server/monitoring.ts
Line: 9-17

Comment:
**Module-level socket outlives disabled monitoring**

`socket` is created lazily at module scope and is never closed. If monitoring is later disabled mid-process (opt-out set after first call), the socket stays alive for the process lifetime. Since `socket.unref()` is called, this won't prevent Node.js from exiting — it's a minor resource concern rather than a correctness bug. If a future change needs explicit teardown (e.g., test cleanup), returning a dispose function from `createMonitoringEntry` would be the path.

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(core): address Greptile review comme..."](https://github.com/alpic-ai/skybridge/commit/cff0ec6d161385c93971eb9ebf2d838ab29c5f71) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27839916)</sub>

### CODE REVIEW

**Comment by harijoe** [comment]
@greptile

File: infrastructure/lib/constructs/Monitoring.ts [hunk 1]
```diff
@@ -0,0 +1,92 @@
...
+ * Receives UDP StatsD metrics from Skybridge apps on port 8125 and forwards
+ * them to CloudWatch under the "Skybridge" namespace.
+ *
+ * Accessible at metrics.skybridge.tech:8125 (UDP).
+ */
+export class Monitoring extends Construct {
+  constructor(scope: Construct, id: string, { hostedZone }: MonitoringProps) {
+    super(scope, id);
+
+    const vpc = Vpc.fromLookup(this, "DefaultVpc", { isDefault: true });
```
fredericbarthelet: You should create your own VPC, there is no inherent cost in setting up another one, and its considered bad practice to use the default one because the resource is not owned by the cloudformation stack

```suggestion
    const vpc = new Vpc(this, "Vpc");
```

fredericbarthelet: You should create your own VPC, there is no inherent cost in setting up another one, and its considered bad practice to use the default one because the resource is not owned by the cloudformation stack

```suggestion
    const vpc = new Vpc(this, "Vpc");
```

packages/core/src/server/monitoring.ts
```diff
@@ -0,0 +1,76 @@
...
+  if (!socket) {
+    socket = createSocket("udp4");
+    socket.unref();
+  }
+  return socket;
+}
+
+function sendMetrics(metrics: string[]): void {
+  const payload = Buffer.from(metrics.join("\n"));
+  getSocket().send(payload, STATSD_PORT, STATSD_HOST, () => {
```
fredericbarthelet: Let me check with Matthieu if there were a specific reason for using an IP instead of a domain as target on the Bref's implementation.
I'm suspecting there might be a performance issue in resolving domain

infrastructure/lib/constructs/Monitoring.ts
```diff
@@ -0,0 +1,92 @@
...
+      securityGroup,
+      role,
+      userData,
+      userDataCausesReplacement: true,
+    });
+
+    new ARecord(this, "MetricsRecord", {
+      zone: hostedZone,
+      recordName: "metrics",
+      target: RecordTarget.fromIpAddresses(instance.instancePublicIp),
```
fredericbarthelet: If we use a domain, that's fine.
If we need to switch to IP target, we'll need to replace record with an elastic IP to ensure it stays the same when we restart/update the EC2

infrastructure/lib/constructs/SkybridgeRecords.ts
```diff
@@ -7,16 +7,18 @@ type SkybridgeRecordsProps = {
 };
 
 export class SkybridgeRecords extends Construct {
+  public readonly hostedZone: PublicHostedZone;
```
fredericbarthelet: you're reversing dependency here, it feels a bit weird.

I suggest you either :
- move the hosted zone out of this construct and hoist in in root stack so that it can be used by this construct and the new Monitoring construct (this will require updating the new hosted zone NS records on namecheap to avoid downtime)
- add the EC2 ip as dependency of SkybridgeRecords and create the A record in this construct

packages/core/src/server/monitoring.ts
fredericbarthelet, line 1: metric.ts to stay consistent ?

infrastructure/lib/constructs/Monitoring.ts
```diff
@@ -0,0 +1,92 @@
...
+};
+
+/**
+ * A t4g.nano EC2 instance running the CloudWatch Agent with StatsD receiver.
+ * Receives UDP StatsD metrics from Skybridge apps on port 8125 and forwards
+ * them to CloudWatch under the "Skybridge" namespace.
+ *
+ * Accessible at metrics.skybridge.tech:8125 (UDP).
+ */
+export class Monitoring extends Construct {
```
fredericbarthelet: Given the name you're giving to the domain, I'd suggest using `Metrics` too here no?

infrastructure/lib/constructs/Metrics.ts
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    const securityGroup = new SecurityGroup(this, "SecurityGroup", {
+      vpc,
+      description: "Skybridge StatsD monitoring ingress",
+      allowAllOutbound: true,
+    });
+    securityGroup.addIngressRule(
+      Peer.anyIpv4(),
+      Port.udp(8125),
```
fredericbarthelet, line 53: Out of curiosity, why this port?

packages/core/src/server/monitoring.ts
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * Metrics (StatsD counter format, no native sampling field):
+ *   Invocations_1:1|c       — every tool call
+ *   Errors_1:1|c            — tool returned isError:true or threw
+ *   ColdInvocations_1:1|c   — first call since process start
+ *
+ * The _1 suffix encodes the sampling denominator (1-in-1 = 100%).
+ * When sampling is introduced later, change the guard and rename to _100.
+ */
+export function createMonitoringEntry(): McpMiddlewareEntry | null {
```
fredericbarthelet:
```suggestion
export function createMiddlewareEntry(): McpMiddlewareEntry | null {
```

packages/core/src/server/monitoring.ts
```diff
@@ -0,0 +1,76 @@
...
+  });
+}
+
+/**
+ * Returns an internal MCP middleware entry that emits StatsD counters over UDP
+ * for every tool invocation. Enabled by default; respects the existing telemetry
+ * opt-out (SKYBRIDGE_TELEMETRY_DISABLED, DO_NOT_TRACK, or `skybridge telemetry disable`).
+ *
+ * Metrics (StatsD counter format, no native sampling field):
+ *   Invocations_1:1|c       — every tool call
```
fredericbarthelet: Can you measure performance impact of this call using the Nodejs `performance` API? Depending on actual execution duration, we might want to revise this sampling rate

packages/core/src/server/monitoring.ts
```diff
@@ -0,0 +1,76 @@
...
+ *   Errors_1:1|c            — tool returned isError:true or threw
+ *   ColdInvocations_1:1|c   — first call since process start
```
fredericbarthelet: this was making sense for Bref's metrics where Lambda's cold invocations were monitored, but I don't understand the intention here

packages/core/src/server/monitoring.ts
```diff
@@ -0,0 +1,76 @@
...
+        metrics.push("Errors_1:1|c");
+      }
+      if (isCold) {
+        metrics.push("ColdInvocations_1:1|c");
+      }
+      sendMetrics(metrics);
+    }
+  };
+
+  return { filter: "tools/call", handler };
```
fredericbarthelet: we should mesure all requests, why only tools/call?

packages/core/src/server/monitoring.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * Returns an internal MCP middleware entry that emits StatsD counters over UDP
+ * for every tool invocation. Enabled by default; respects the existing telemetry
+ * opt-out (SKYBRIDGE_TELEMETRY_DISABLED, DO_NOT_TRACK, or `skybridge telemetry disable`).
+ *
+ * Metrics (StatsD counter format, no native sampling field):
+ *   Invocations_1:1|c       — every tool call
+ *   Errors_1:1|c            — tool returned isError:true or threw
```
fredericbarthelet: those are user error, and not directly related to a usage metric. I don't see the point in collecting it.

packages/core/src/server/monitoring.ts
```diff
@@ -0,0 +1,76 @@
...
+    // fire-and-forget: errors are intentionally silenced
+  });
+}
+
+/**
+ * Returns an internal MCP middleware entry that emits StatsD counters over UDP
+ * for every tool invocation. Enabled by default; respects the existing telemetry
+ * opt-out (SKYBRIDGE_TELEMETRY_DISABLED, DO_NOT_TRACK, or `skybridge telemetry disable`).
+ *
+ * Metrics (StatsD counter format, no native sampling field):
```
fredericbarthelet: we should use Skybridge current version as a dimension for the metric, in order to get information on rate of update. In order to not have too many dimensions (that quickly ends up pricey on AWS), we might only want to report minor, or even only major

packages/core/src/server/monitoring.ts
fredericbarthelet:
```suggestion
 *   Requests_1:1|c       — every request
```
Invocations is very lambda specific naming

fredericbarthelet (changes requested):
A few comments, but otherwise all good :)
Could you run the performance test to ensure the impact is minimal in order to adjust the sampling if need be?

I'll let you know when I have the answer from Matthieu on IP vs domain

packages/core/src/server/monitoring.ts
harijoe: I was also wondering if we wanted a static ip for this behind the domain.
Maybe using an ip directly is a way to avoid dns resolution and limit the "overhead"?

infrastructure/lib/constructs/Metrics.ts
harijoe: I learnt that it was a [commonly used statd default port](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-custom-metrics-statsd.html)

packages/core/src/server/monitoring.ts
harijoe: to me that's not obvious, if we want to track actual usage, tools/call seem like the most reliable proxy
it's also a way to send less message

packages/core/src/server/monitoring.ts
harijoe: it's also open to debate : if the goal is to monitor actual usage, the version is not relevant
do we want to monitor more than overall usage?

packages/core/src/server/monitoring.ts
harijoe: Decision's made to use an elastic ip

fredericbarthelet (approved):
LGTM! Thanks @harijoe

PR #665, opened 2026-04-13
feat: Add stripe to ecom example app

harijoe:
### Preview

image
image

### Summary

- Stripe Checkout (redirect flow) added to the ecom-carousel example.
- Widget polls for payment completion, shows animated success card, and notifies the LLM.
- Fix 3 broken product image URLs.

### Architecture Notes

- **Inline pricing** — `price_data` on session creation; no Stripe product catalog setup needed.
- **Polling over webhooks** — avoids public URL requirement; keeps example self-contained.

---

<h3>Greptile Summary</h3>

Adds a Stripe Checkout redirect flow to the ecom-carousel example: a new `create-checkout` and `check-checkout-status` server tool pair, a `useCheckoutPolling` hook, a `PaymentSuccess` component, and extracted i18n. The code is clean and well-decomposed. The previously flagged issues (missing try/catch, unguarded empty lineItems, stale timeout closure, missing `.env.example`) have all been addressed in this PR.

One remaining P2 concern: `useRequestModal` has no `close()` API, so after payment succeeds and the user clicks \"Back to products,\" `resetCheckout` resets phase to `\"idle\"` while `isOpen` stays `true` — landing the user on an empty `CheckoutSummary` instead of the product carousel.

Safe to merge with the modal-navigation UX issue acknowledged; LLM notification (flagged in prior review round) is still unimplemented.

All P0/P1 issues from the previous review round are resolved in this diff. The one new finding (modal not closing after reset) is P2 for an example app and doesn't break the payment path itself. The previously flagged missing LLM notification remains open but was already surfaced; that unresolved documented feature is the reason for 4 rather than 5.

examples/ecom-carousel/web/src/widgets/browse-catalog.tsx — modal phase/isOpen ordering

`````markdown
Path: examples/ecom-carousel/web/src/widgets/browse-catalog.tsx
Line: 83-95

Comment:
**"Back to products" doesn't navigate away from the modal**

`useRequestModal` exposes only `open` — there is no `close()`. After payment completes and the user clicks "Back to products," `resetCheckout()` resets `phase` to `"idle"` but `isOpen` stays `true`. On the next render, the `if (isOpen)` branch fires and the user lands on `CheckoutSummary` showing an empty cart, not the product carousel. The user must manually dismiss the modal to return to browsing.

Since `useRequestModal` has no programmatic close, the cleanest fix is to treat the `isOpen` modal as the permanent checkout container and not rely on `phase === "complete/expired"` to escape it. For example, render `PaymentSuccess` *inside* the `if (isOpen)` block instead of before it:

```tsx
if (isOpen) {
  if (phase === "complete") {
    const paidItems = output.products.filter((p) => cart.ids.includes(p.id));
    return (
      <div className={`${theme} checkout`}>
        <PaymentSuccess items={paidItems} />
        <button type="button" className="checkout-button" onClick={resetCheckout}>
          {t("backToProducts")}
        </button>
      </div>
    );
  }
  if (phase === "expired") { /* ... */ }
  // default: CheckoutSummary
}
```

This way "Back to products" resets back to `CheckoutSummary` in a predictable state (still inside the modal), and the product carousel is visible once the user closes the modal naturally.

`````

</details>

<sub>Reviews (4): Last reviewed commit: ["fix: address remaining review comments"](https://github.com/alpic-ai/skybridge/commit/d695687dfd381e3c9ad6f96aa598361d3c9c2e5d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28182499)</sub>

### CODE REVIEW

**Comment by harijoe** [comment]
@greptile

**Comment by harijoe** [comment]
@greptileai

File: examples/ecom-carousel/server/src/env.ts [hunk 1]
```diff
@@ -0,0 +1,12 @@
+import "dotenv/config";
+
+import { createEnv } from "@t3-oss/env-core";
+import { z } from "zod";
+
+export const env = createEnv({
+  server: {
+    STRIPE_SECRET_KEY: z.string().min(1),
```
paulleseute, line 8: do we need 2 extra deps for a single var ?

examples/ecom-carousel/web/src/hooks/use-checkout-polling.ts
```diff
@@ -0,0 +1,81 @@
...
+          const status = result.structuredContent?.status;
+          if (status === "complete") {
+            stopPolling();
+            setCheckoutState({ phase: "complete" });
+          } else if (status === "expired") {
+            stopPolling();
+            setCheckoutState({ phase: "expired" });
+          }
+        } catch {
+          // Stripe may temporarily error — keep polling
```
paulleseute, line 57: we should have an exit condition we control with a timeout, otherwise this keeps polling forever.

harijoe:
@greptileai

examples/ecom-carousel/server/src/index.ts
```diff
@@ -21,56 +22,167 @@ const server = new McpServer(
...
+            isError: true,
+          };
         }
-        filtered.push(product);
+
+        const session = await stripe.checkout.sessions.create({
+          mode: "payment",
+          line_items: lineItems,
+          success_url: "https://docs.skybridge.tech",
+          cancel_url: "https://docs.skybridge.tech",
```
paulleseute, line 126: can't we redirect to the chat ?

paulleseute (commented):
is managing checkout session in via tool calls a good pattern ? is the llm suppose to use it ? what is the advantage vs an API (which is not subject to chatgpt duration penalty) ?

examples/ecom-carousel/server/src/env.ts
harijoe: It's more about showing an example of what we consider as a good practice?

examples/ecom-carousel/server/src/index.ts
harijoe: I am not sure it's possible, let's talk

examples/ecom-carousel/web/src/hooks/use-checkout-polling.ts
harijoe: We do have a timeout mechanism firing after 5 minutes, is that what you meant?

paulleseute (approved):
I'd mention in the doc that dev should consider a non-mcp endpoint for checkout mgmt since the tools do not really bring value here, and add latency because of the sandboxing.

PR #671, opened 2026-04-14
feat(infra): replace CloudWatch Agent with Vector for StatsD version filtering

harijoe:
### Summary

- Replace CloudWatch Agent with Vector as the StatsD receiver on the metrics EC2 instance.
- Filter incoming metrics against published skybridge versions fetched from npm.
- Metrics with unknown version tags are dropped before reaching CloudWatch, preventing unbounded custom metric cardinality and billing.
- A cron job refreshes the allow-list every 15 minutes; the CSV is seeded at boot before Vector starts.

### Architecture Notes

- **Vector enrichment tables** — the version allow-list is a CSV file loaded as a Vector enrichment table, queried in the filter transform via `get_enrichment_table_records`. Vector auto-reloads the file on change.
- **Instance replacement** — `userDataCausesReplacement: true` means deploying this will replace the EC2 instance (same Elastic IP preserved).

<h3>Greptile Summary</h3>

This PR replaces the CloudWatch Agent with Vector as the StatsD receiver and adds a version-filter that drops metrics whose `version` tag doesn't match a published `major.minor` from npm. The metric client in `metric.ts` already emits tags in `major.minor` format, so the allow-list format is consistent.

All previously flagged blockers (broken heredoc delimiter, wrong VRL function name, wrong encoding field, hardcoded region, first-boot failure, Vector 0.45.0 crash) are addressed in the current implementation. Two minor items remain: the RPM is installed without importing Vector's GPG public key (signature not verified), and the refresh script doesn't clean up its temp file on early exit.

Safe to merge — all prior blocking issues are resolved; remaining findings are P2 improvements only.

Every P0/P1 issue raised in previous review rounds (heredoc delimiter, VRL function name, encoding field, region hardcoding, boot-time failure, Vector crash bug) has been corrected. The two new findings (GPG signature verification and temp-file cleanup) are P2 hardening suggestions that don't affect correctness or availability of the feature.

infrastructure/lib/constructs/Metrics.ts — minor: GPG key import and temp file trap in refresh script.

<details open><summary><h3>Security Review</h3></summary>

- **Missing GPG key import before `rpm -i`** (`infrastructure/lib/constructs/Metrics.ts`, line 107): Vector's signing key is not imported prior to installation, so the package signature is not verified. An attacker who can compromise `yum.vector.dev` or intercept the download (despite HTTPS) can deliver a malicious binary. The previous `curl | bash` risk has been reduced by switching to a pinned RPM URL, but signature enforcement requires an explicit `rpm --import` of Vector's public key first.
</details>

`````markdown
Path: infrastructure/lib/constructs/Metrics.ts
Line: 107

Comment:
**RPM installed without GPG signature verification**

`rpm -i` on a fresh Amazon Linux 2 instance does not verify the package signature when Vector's public key has not been imported, so a compromise of `yum.vector.dev` or a DNS hijack can still deliver a malicious package. Import the key first so rpm enforces signature checking before installation:

```suggestion
      "rpm --import https://keys.openpgp.org/vks/v1/by-fingerprint/XXXXXXXXXXXXXXXXXXXX",
      "rpm -i https://yum.vector.dev/stable/vector-0/aarch64/vector-0.46.1-1.aarch64.rpm",
```

The official Vector GPG key fingerprint is published at https://vector.dev/releases/ — replace the placeholder with the actual fingerprint URL. Alternatively, download the RPM, verify its SHA-256 checksum against Vector's published hash, then install locally.

---

Path: infrastructure/lib/constructs/Metrics.ts
Line: 47-57

Comment:
**Temp file leaks on script failure**

`TMP=$(mktemp)` is not cleaned up if the script exits early (e.g. `curl` or `jq` fail under `set -euo pipefail`). Every failed cron run leaves an orphan file in `/tmp`. Add an `EXIT` trap immediately after `mktemp`:

```suggestion
TMP=$(mktemp)
trap "rm -f $TMP" EXIT

echo "version" > "$TMP"
```

`````

</details>

<sub>Reviews (4): Last reviewed commit: ["fix(infra): bump Vector to 0.46.1 to avo..."](https://github.com/alpic-ai/skybridge/commit/13285c11a85b0efffb8095a11b0f2aed07d8d308) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=28373970)</sub>


harijoe:
@greptileai

harijoe:
@greptileai

harijoe:
@greptileai

fredericbarthelet (approved):
Sounds like a great plan :)
A few improvements:
- I'd move the vector config to a dedicated config file rather than inline in ts file
- I'd move the package version pulling frequency down to one per minute to avoid "large" blank periods where latest is not in allowed list. npmjs is fine being frequently requested 😅
- I'd check that new TMP is largest than allowed-version before swapping in case npmjs is down for some time to avoid inadvertidly denying all versions

infrastructure/lib/constructs/vector/vector.toml
fredericbarthelet, line 1: can't you just use a toml file and import it in Metrics.ts using a raw import?

harijoe: I agree, done

PR #683, opened 2026-04-15
fix: sort dev logs chronologically across sources

paulleseute:
nodemon and tunnel messages were interleaved by source rather than by time, favoring tunnel logs


fredericbarthelet (approved):
Good for me :)

PR #701, opened 2026-04-23
feat: unify registerTool API, tool-widget e2e typing, flat structure (SKY-332)

harijoe:
## Summary

- Unify `registerTool` / `registerWidget` into a single `registerTool(config, handler)` call with an optional `view` field for widget config. view.component key is strongly typed based on automatically discovered widgets.
- Remove `mountWidget` from public API — Vite plugin auto-mounts widgets via virtual entries
- Add content helpers (`text`, `image`, `audio`, `embeddedResource`, `resourceLink`) exported from `skybridge/server`. Also handler `content` now accepts a plain string or single `ContentBlock` (normalized to array)
- Vite plugin: configurable `widgetsDir` (default `src/views`), `.tsx`/`.jsx` only, generates `.skybridge/widgets.d.ts` with `WidgetNameRegistry` for compile-time `view.component` narrowing
- Flatten project layout to a single `src/` tree — `server/` + `web/` split is no longer supported
- One-tool-per-widget enforcement at startup, manifest lookup by entry name

## Commit Layout

Split into four commits for review:

1. **`feat: migrate template to flat layout + unified registerTool API`** 
2. **`test: cover unified registerTool API, content helpers, and widget discovery`** 
3. **`feat(core): unify registerTool API, drop server/+web/ layout (SKY-332)`**
4. **`docs: update to unified registerTool API`**

**Only template has been adapted, examples will be updated in a following PR**

## Breaking Changes — Migration Guide

Skybridge is pre-1.0 and this PR drops support for the old `server/` + `web/` split layout. All projects must migrate to the flat `src/` layout.

## Architecture Notes

- **Base class override** — `McpServerBase.registerTool` has an incompatible 3-arg signature; solved with `Omit` on the base type + interface declaration merging + prototype assignment
- **WidgetName register pattern** — empty `WidgetNameRegistry` interface in core; generated `.skybridge/widgets.d.ts` augments `skybridge/server` to add keys, narrowing `WidgetName` from `string` to the actual union
- **Build output layout** — Vite plugin pins `build.outDir` to `dist/assets`, so `tsc` (server → `dist/`) and Vite (widgets → `dist/assets/`) share one `dist/` without collision; build order is now tsc → vite
- **Manifest lookup** — Vite keys virtual-module entries by resolved id (`skybridge:widget:<name>`), not by the `input` key. Server iterates manifest entries and matches on `entry.name`, so the plugin's virtual-id convention can change without breaking production lookups

---

<h3>Greptile Summary</h3>

This PR unifies the `registerTool`/`registerWidget` API into a single `registerTool(config, handler)` call with an optional `view` field, removes `mountWidget` from the public API (replaced by the Vite plugin's auto-mount virtual entries), adds content helpers, and flattens the project layout to a single `src/` tree. The implementation is well-structured: the `outDir` fix, Windows path normalization in `buildVirtualEntry`, duplicate-view filtering ordering, dev server crash handling, and pre-scan `viewsDir` resolution from the Vite config are all correctly addressed in this revision.

Mostly safe to merge; remaining concerns from prior review rounds are tracked on the PR and no new P0/P1 bugs were found in this pass.

No new P0 or P1 bugs found in the current diff. Most previously-flagged issues have been addressed (outDir, Windows paths, duplicate-name ordering, dev-server crash, pre-scan viewsDir resolution, content normalisation). Two items from prior rounds remain open on the PR (globSync Node 22 compatibility, overload return-type mismatch on the second registerTool signature), which cap the score below 5.

packages/core/src/web/plugin/scan-views.ts (globSync Node 22 compatibility), packages/core/src/server/server.ts (second overload return type)

<details><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `packages/core/src/server/server.ts`, line 430-436 ([link](https://github.com/alpic-ai/skybridge/blob/291ea9ee1d514a620c31cc2a9f3aa643082dde3a/packages/core/src/server/server.ts#L430-L436)) 

   <a href="#">P1</a> **`content` required by type but documented and used as optional**

   The `ToolHandler` constraint `TReturn extends { content: HandlerContent }` makes `content` a required field. However, the published docs (`register-tool.mdx`) explicitly show `content?` as optional, and several documentation examples return only `structuredContent` with no `content` field — patterns that would fail TypeScript compilation with the current constraint.

   Downstream, `wrapHandler` unconditionally calls `normalizeContent(result.content)`. When `content` is absent (as the old `WidgetHandler` API permitted), `normalizeContent(undefined)` falls through every branch and returns `[undefined]` — an invalid `ContentBlock[]` that will cause silent MCP protocol errors.

   The fix is to make `content` optional in both the type and the normalizer:

   ```ts
   // server.ts — type
   type ToolHandler<
     TInput extends ZodRawShapeCompat,
     TReturn extends { content?: HandlerContent } = { content?: HandlerContent },
   > = ...

   // normalizeContent — guard undefined
   export function normalizeContent(content: HandlerContent | undefined): ContentBlock[] {
     if (content === undefined || content === null) return [];
     if (typeof content === "string") {
       return [{ type: "text", text: content }];
     }
     if (Array.isArray(content)) {
       return content;
     }
     return [content];
   }
   ```

   This also aligns with the old `WidgetHandler` where `content?` was optional.

   `````markdown
   Path: packages/core/src/server/server.ts
   Line: 430-436

   Comment:
   **`content` required by type but documented and used as optional**

   The `ToolHandler` constraint `TReturn extends { content: HandlerContent }` makes `content` a required field. However, the published docs (`register-tool.mdx`) explicitly show `content?` as optional, and several documentation examples return only `structuredContent` with no `content` field — patterns that would fail TypeScript compilation with the current constraint.

   Downstream, `wrapHandler` unconditionally calls `normalizeContent(result.content)`. When `content` is absent (as the old `WidgetHandler` API permitted), `normalizeContent(undefined)` falls through every branch and returns `[undefined]` — an invalid `ContentBlock[]` that will cause silent MCP protocol errors.

   The fix is to make `content` optional in both the type and the normalizer:

   ```ts
   // server.ts — type
   type ToolHandler<
     TInput extends ZodRawShapeCompat,
     TReturn extends { content?: HandlerContent } = { content?: HandlerContent },
   > = ...

   // normalizeContent — guard undefined
   export function normalizeContent(content: HandlerContent | undefined): ContentBlock[] {
     if (content === undefined || content === null) return [];
     if (typeof content === "string") {
       return [{ type: "text", text: content }];
     }
     if (Array.isArray(content)) {
       return content;
     }
     return [content];
   }
   ```

   This also aligns with the old `WidgetHandler` where `content?` was optional.

   `````
   </details>

2. `packages/core/src/web/plugin/plugin.ts`, line 183-186 ([link](https://github.com/alpic-ai/skybridge/blob/55c7bb29bd248f86cc450efb9dd02da19eb0d676/packages/core/src/web/plugin/plugin.ts#L183-L186)) 

   <a href="#">P1</a> **Uncaught exception crashes the dev server on duplicate widget names**

   `rescan` calls `discoverWidgetsSync`, which throws a synchronous `Error` when a duplicate widget name is detected. An unhandled synchronous throw inside a Chokidar `"add"` or `"unlink"` event handler propagates up through `EventEmitter.emit()` as an uncaught exception, crashing the Vite dev server process.

   A realistic scenario: a developer already has `views/foo.tsx` and then creates `views/foo/index.tsx` with a default export — this immediately kills the dev server rather than logging a friendly error.

   Wrap the callback body in a try/catch and use `server.config.logger.error` to report the failure instead of crashing:

   ```ts
   const rescan = () => {
     try {
       const widgets = discoverWidgetsSync(resolvedWidgetsDir);
       widgetMap = new Map(widgets.map((w) => [w.name, w]));
       writeWidgetsDts(projectRoot, widgets);
     } catch (err) {
       server.config.logger.error(String(err));
     }
   };
   ```

   `````markdown
   Path: packages/core/src/web/plugin/plugin.ts
   Line: 183-186

   Comment:
   **Uncaught exception crashes the dev server on duplicate widget names**

   `rescan` calls `discoverWidgetsSync`, which throws a synchronous `Error` when a duplicate widget name is detected. An unhandled synchronous throw inside a Chokidar `"add"` or `"unlink"` event handler propagates up through `EventEmitter.emit()` as an uncaught exception, crashing the Vite dev server process.

   A realistic scenario: a developer already has `views/foo.tsx` and then creates `views/foo/index.tsx` with a default export — this immediately kills the dev server rather than logging a friendly error.

   Wrap the callback body in a try/catch and use `server.config.logger.error` to report the failure instead of crashing:

   ```ts
   const rescan = () => {
     try {
       const widgets = discoverWidgetsSync(resolvedWidgetsDir);
       widgetMap = new Map(widgets.map((w) => [w.name, w]));
       writeWidgetsDts(projectRoot, widgets);
     } catch (err) {
       server.config.logger.error(String(err));
     }
   };
   ```

   `````
   </details>
</details>

<sub>Reviews (20): Last reviewed commit: ["Fix docs"](https://github.com/alpic-ai/skybridge/commit/c003b0550a83426572fcaa496f805c6e814b7e14) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29503990)</sub>


harijoe:
@greptileai

packages/create-skybridge/template/README.md
```diff
@@ -44,34 +44,31 @@ This command starts:
...
+│   └── index.css         # Global styles
+├── vite.config.ts
 ├── alpic.json            # Deployment config
-├── nodemon.json          # Dev server config
 └── package.json
 ```
 
 ### Create your first widget
 
 #### 1. Add a new widget
```
paulleseute: i guess we shouldn't call it a widget anymore ?

packages/core/src/commands/build.tsx
```diff
@@ -1,23 +1,53 @@
...
+ * narrowing correct for users with a custom widget dir.
+ */
+async function resolveWidgetsDir(root: string): Promise<string | undefined> {
+  const { loadConfigFromFile } = await import("vite");
+  const loaded = await loadConfigFromFile(
+    { command: "build", mode: "production" },
+    undefined,
+    root,
+  );
+  const raw = (loaded?.config.plugins ?? []) as unknown as unknown[];
```
paulleseute: should we validate instead of asserting ?

packages/core/src/server/server.ts
```diff
@@ -64,7 +63,68 @@ export type ToolDef<
...
+  redirectDomains?: string[];
+  /** Origins allowed in `<base href>` tags (mcp-apps only). */
+  baseUriDomains?: string[];
+}
+
+// Must be exported: TS module augmentation only merges with exported
+// declarations. Without `export`, `.skybridge/widgets.d.ts` augmentation
+// would create a separate interface and `WidgetName` would stay `string`.
+// biome-ignore lint/suspicious/noEmptyInterface: register pattern — augmented by `.skybridge/widgets.d.ts` to narrow WidgetName
+export interface WidgetNameRegistry {}
```
paulleseute: ViewNameRegistry ?

packages/core/src/server/server.ts
```diff
@@ -738,25 +758,112 @@ export class McpServer<
     );
   }
 
-  private lookupDistFile(key: string): string {
-    const manifest = this.readManifest();
-    return manifest[key]?.file;
+  // -------------------------------------------------------------------------
+  // Handler wrapping
+  // -------------------------------------------------------------------------
```
paulleseute: i don't think that kind of comment formatting is necessary

packages/core/src/web/plugin/plugin.ts
```diff
@@ -51,11 +103,66 @@ export function skybridge(): Plugin {
...
+      server.watcher.on("add", rescan);
+      server.watcher.on("unlink", rescan);
+    },
+
     async transform(code, id) {
-      if (WIDGET_ENTRY_RE.test(id)) {
-        for (const warning of validateWidget(code, id)) {
-          this.warn(warning.message);
+      if (widgetEntryRe?.test(id)) {
+        if (!hasDefaultExport(code, id)) {
```
paulleseute: this can me merged

docs/api-reference/generate-helpers.mdx
```diff
@@ -39,21 +39,28 @@ const { callTool } = useCallTool("search-voyage");
...
 ```ts
-// ✅ Works — chain registerWidget/registerTool calls
+// ✅ Works — chain registerTool calls
 const server = new McpServer({ name: "my-app", version: "1.0" }, {})
-  .registerWidget("search-voyage", {}, { inputSchema: { destination: z.string() } }, async ({ destination }) => {
-    return { content: [{ type: "text", text: `Found trips to ${destination}` }] };
-  });
+  .registerTool(
+    {
+      name: "search-voyage",
```
paulleseute: search-trip ?

paulleseute (commented):
Few remarks:
- I'm not sure to understand how we can update core and template at the same time, template referencing a published version of skybridge. As-is the template is broken.
- I think we should get rid of the `widget` term, having both "view" and "widget" in the code/docs is confusing
- How do we plan to rollout this ? From what I understand it's really breaking so should not be merged before v1. Or is this v1 ?
- TBH this is a bit hard to review since it tackling several issues at the same time (registerTool vs structure), IMO it could have been two PRs

harijoe:
> Few remarks:
> 
> * I'm not sure to understand how we can update core and template at the same time, template referencing a published version of skybridge. As-is the template is broken.
> * I think we should get rid of the `widget` term, having both "view" and "widget" in the code/docs is confusing
> * How do we plan to rollout this ? From what I understand it's really breaking so should not be merged before v1. Or is this v1 ?
> * TBH this is a bit hard to review since it tackling several issues at the same time (registerTool vs structure), IMO it could have been two PRs

- If you pull the code and start the new template, it will be broken as long as the new core isn't published. When the new core is published the template will be working properly. There's a test making sure core and template are compatible with each other in the CI. To test it locally, you need to get the 0.0.0-dev.xxxxx version of the core and use it in the new template.
- I agree, I'll make sure we do
- Breaking changes are okay as we are still in v0. So we roll it out with migration instructions in the release. The idea here is to rollout the new structure before v1, as after v1 we won't be able to release breaking changes anymore.
- I agree, even though it was also to make sure that the final state (the one currently proposed) was actually implementable. That was not given at first!

docs/api-reference/generate-helpers.mdx
harijoe: Yes let's fix it

docs/api-reference/mcp-server.mdx
```diff
@@ -295,6 +298,6 @@ import type {
 
 ## Related
 
-- [registerWidget](/api-reference/register-widget) - Widget registration details
+- [registerTool](/api-reference/register-tool) - Tool/view registration details
```
paulleseute: Tool (with or without view)

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+  handler: ToolHandler,
+): McpServer
+```
+
+A single call replaces the previous `registerTool(name, config, handler)` + `registerView(name, resourceConfig, toolConfig, handler)` pair.
+
+## Config
+
+```typescript
+type UnifiedToolConfig = {
```
paulleseute: i'd call this user-facing API `ToolConfig`

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
+---
+title: registerTool
+description: "Register an MCP tool — with or without a React view — in a single unified call."
```
paulleseute: "in a single unified call." seems superfluous

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+## Signature
+
+```typescript
+server.registerTool(
+  config: UnifiedToolConfig,
+  handler: ToolHandler,
+): McpServer
+```
+
+A single call replaces the previous `registerTool(name, config, handler)` + `registerView(name, resourceConfig, toolConfig, handler)` pair.
```
paulleseute: i don't think this is either true of needed

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+  input: Input,
+  extra: RequestHandlerExtra,
+) => Promise<{
+  content?: string | ContentBlock | ContentBlock[]; // Text(s) for the model
+  structuredContent?: StructuredData;               // Data for the view
+  _meta?: Record<string, unknown>;                  // Response metadata
+}>;
+```
+
+`content` can be a plain string, a single `ContentBlock`, or an array — Skybridge normalizes it to an array internally. Use the [content helpers](#content-helpers) to build blocks ergonomically.
```
paulleseute: an array of content block

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+```typescript
+const server = new McpServer(config, {})
+  .registerTool({ name: "a", /* … */ }, handlerA)
+  .registerTool({ name: "b", /* … */ }, handlerB)
+  .registerTool({ name: "c", /* … */ }, handlerC);
+```
+
+<Note>
+Method chaining is **required** for type inference. See [Type Safety](/concepts/type-safety#method-chaining-requirement).
+</Note>
```
paulleseute: If chaining is required then this paragraph should be hoisted and be called "Usage" or something like that. also the whole section could be a oneliner in the 'signature' section

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+
+## Response Fields
+
+Three fields, three audiences:
+
+| Field | Purpose | Consumed by |
+|-------|---------|-------------|
+| `content` | Text description | The host (shown in conversation) |
+| `structuredContent` | Typed data | Host and view (`useToolInfo`, `useCallTool`) |
+| `_meta` | Response metadata | View only (hidden from the model) |
```
paulleseute, line 94: are you sure about that ? my understanding is that it depends of the model. i'm not sure claude accesses structuredContent, or maybe it has changed

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+
+See [Data Flow: Response Fields](/concepts/data-flow#response-fields-explained).
+
+<Info>
+**viewUUID**: Skybridge automatically injects a unique `viewUUID` into `_meta` for every view tool call. This enables `useWidgetState` and `createStore` to persist state across re-renders in MCP Apps via `localStorage`. Any `_meta` you return is preserved alongside the injected `viewUUID`.
+</Info>
+
+## Examples
+
+### View tool
```
paulleseute: Tool with view

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+    async ({ name }) => {
+      return {
+        content: `Greeting ${name}`,
+        structuredContent: { message: `Hello, ${name}!` },
+      };
+    },
+  );
+```
+
+### Plain tool (no UI)
```
paulleseute: Tool without view

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+import { McpServer } from "skybridge/server";
+import { z } from "zod";
+
+const server = new McpServer({ name: "my-app", version: "0.0.1" }, {})
+  .registerTool(
+    {
+      name: "greet",
+      description: "Greet someone by name.",
+      inputSchema: { name: z.string() },
+      view: { component: "greeting", description: "A greeting card" },
```
paulleseute, line 118: do we need the view description ?

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+        content: `Greeting ${name}`,
+        structuredContent: { message: `Hello, ${name}!` },
+      };
+    },
+  );
+```
+
+### Plain tool (no UI)
+
+Omit `view` to register a tool that returns only content to the model:
```
paulleseute: Omit implies it existed in the first place: `Tool without `view` returns content without UI`

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+    },
+  },
+  async ({ city, units }) => {
+    const weather = await fetchWeather(city, units);
+    return {
+      content: `Current weather in ${city}: ${weather.temperature}°, ${weather.conditions}`,
+      structuredContent: weather,
+    };
+  },
+);
```
paulleseute, line 171: we don't need the schema to be that complicated, input: city seems enough

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+```
+
+<Tip>
+If your view makes client-side API calls, add those domains to `connectDomains`. Server-side calls from the handler do not require CSP configuration.
```
paulleseute: the example implies the contrary since we're not showing any component making http calls

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+```typescript
+server.registerTool(
+  {
+    name: "user-profile",
+    description: "Show user profile.",
+    inputSchema: { userId: z.string() },
+    view: { component: "user-profile" },
+  },
+  async ({ userId }, extra) => {
+    const email = extra.authInfo?.extra?.email as string | undefined;
```
paulleseute: i don't think we should encourage type assertions

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+    if (!user) {
+      return {
+        content: `User ${userId} not found`,
+        structuredContent: { error: "not_found", userId },
+      };
+    }
+    return {
+      content: `Showing profile for ${user.name}`,
+      structuredContent: { ...user, viewedBy: email },
+    };
```
paulleseute, line 199: what about isError ? also, should we showcase the try/catch pattern ?

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+| `image(data, mimeType, annotations?)` | `ImageContent` (base64-encoded) |
+| `audio(data, mimeType, annotations?)` | `AudioContent` (base64-encoded) |
+| `embeddedResource(resource, annotations?)` | `EmbeddedResource` |
+| `resourceLink(link, annotations?)` | `ResourceLink` |
+
+You can still return plain `ContentBlock` objects — helpers are opt-in.
+
+## Typed view names
+
+The Vite plugin generates `.skybridge/views.d.ts` at dev/build time, augmenting the internal `ViewNameRegistry` interface. This narrows `ViewName` from `string` to the union of actual view file names in `src/views/`.
```
paulleseute: i don't think discussing the internal makes things clearer

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+
+### Priority & Merging
+
+CSP arrays are **merged** (not overridden). Your domains are added to defaults:
+
+```typescript
+// Skybridge auto-includes: ["https://your-server.com"]
+// You add: ["https://api.example.com"]
+// Result:  ["https://your-server.com", "https://api.example.com"]
+```
```
paulleseute: i don't think we need to specify that

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
+---
+title: registerTool
```
paulleseute, line 2: the structure is a bit odd, not sure why we discuss "response fields" (which should be part of the handler ?) then have a bunch of example, then back to be api, csp are discussed twice, etc. imo we should have a more funneled approach. i don't think we need a "example" section, but have example to illustrate sections

docs/api-reference/register-tool.mdx
```diff
@@ -0,0 +1,373 @@
...
+    name: z.string(),
+    price: z.number(),
+  })),
+  totalCount: z.number(),
+}
+```
+
+<Note>
+Output schema is primarily for documentation and type inference. Runtime validation is not enforced.
+</Note>
```
paulleseute: i think llms reads it too ? or does that fall into "documentation" ?

docs/concepts/data-flow.mdx
```diff
@@ -15,83 +15,98 @@ description: "Learn how data moves between the host, your MCP server, and widget
...
-- **`structuredContent`**: Typed JSON data surfaced to your widget and the host
-- **`_meta`**: Delivered only to the widget and hidden from the model
+- **`structuredContent`**: Typed JSON data surfaced to your view and the host
+- **`_meta`**: Delivered only to the view and hidden from the model
 
-## Tools vs Widgets
+## Tools vs Views
 
-Before diving into the data flow, understand the difference between tools and widgets in Skybridge:
+Before diving into the data flow, understand the difference between plain tools and view tools. Both use `registerTool`, differentiated by the optional `view` field:
```
paulleseute: i don't think "view tools" is semantically correct

docs/concepts/data-flow.mdx
```diff
@@ -15,83 +15,98 @@ description: "Learn how data moves between the host, your MCP server, and widget
...
+);
 ```
 
 ## Data Flow Patterns
 
-### 1. Tool → Widget (Initial Hydration)
+### 1. Tool → View (Initial Hydration)
 
-When the host calls one of your widgets, it returns `structuredContent` to hydrate the React component:
+When the host calls one of your views, it returns `structuredContent` to hydrate the React component:
```
paulleseute: does it call the view ?

docs/concepts/type-safety.mdx
```diff
@@ -146,6 +161,30 @@ function generateHelpers<ServerType>() {
 }
 ```
 
+## Typed `view.component`
+
+The Vite plugin scans `src/views/` and generates `.skybridge/views.d.ts`. That file augments an internal `ViewNameRegistry` interface, narrowing `ViewName` from `string` to the union of your actual view file names.
+
+Result: `view.component` autocompletes and typos become compile errors.
```
paulleseute: i don't think we care about how it works

docs/concepts/type-safety.mdx
```diff
@@ -1,10 +1,10 @@
 ---
 title: Type Safety
```
paulleseute, line 2: i know it's pre-existing but this part is very bloated for a rather simple concept. we should focus on the end result and not the problem or implementation

docs/quickstart/add-to-existing-app/server.mdx
```diff
@@ -33,44 +33,39 @@ import { McpServer } from "skybridge/server";
...
 Keep your existing dependency to @modelcontextprotocol/sdk in your project, skybridge does not replaces it, but depends on it.
 
-Your existing tools, resources, and prompts will continue to work as before. With `skybridge/server`, you will also be able to register **widgets**, a new class encompassing a Tool and a corresponding UI Resource.
+Your existing tools, resources, and prompts will continue to work as before. With `skybridge/server`, `registerTool` gains an optional `view` field so you can bind a tool to a React view in a single call.
 
-## Widget structure
+## Project structure
 
-Skybridge follows a file-based convention for organizing widgets. Here's the recommended project structure:
+Skybridge uses a flat `src/` layout with views in `src/views/`:
```
paulleseute: does not look very flat (maybe vs before, but that we don't care)

docs/quickstart/add-to-existing-app/server.mdx
```diff
@@ -79,54 +74,51 @@ export default defineConfig({
 });
 ```
 
-## Create your first widget
+The plugin auto-mounts every file in `src/views/` — no `mountView` call needed.
```
paulleseute: i don't think this is necessary

docs/quickstart/add-to-existing-app/web.mdx
```diff
@@ -27,34 +27,35 @@ pnpm add skybridge
...
 - **[`useAppsSdkContext`](/api-reference/use-apps-sdk-context)**: Low-level hook to subscribe to `window.openai` state values
-- **[`generateHelpers`](/api-reference/generate-helpers)**: Generate typed helpers for your widgets (requires `skybridge/server`)
+- **[`generateHelpers`](/api-reference/generate-helpers)**: Generate typed helpers for your views (requires `skybridge/server`)
 
 Each hook is documented in the API Reference section in the sidebar.
 
 ## Basic example
 
-Here's a simple widget using Skybridge hooks:
+Here's a simple view using Skybridge hooks. Views live in `src/views/` and `export default` a React component — the Vite plugin auto-mounts them.
```
paulleseute: "— the Vite plugin auto-mounts them." is superfluous

paulleseute (commented):
I reviewed the docs and think we can make some improvements, happy to make a commit if you agree

packages/core/package.json
```diff
@@ -7,6 +7,9 @@
     "url": "https://github.com/alpic-ai/skybridge.git"
   },
   "type": "module",
+  "engines": {
+    "node": ">=22.0.0"
+  },
```
fredericbarthelet, line 12: why enforce node22? node20 is still LTS

packages/core/tsconfig.base.json
```diff
@@ -6,6 +6,8 @@
     "verbatimModuleSyntax": true,
 
     // Output options
+    "rootDir": "${configDir}/src",
+    "outDir": "${configDir}/dist",
```
fredericbarthelet, line 10: aren't those the default already?

docs/api-reference/register-tool.mdx
harijoe: We currently map tool output to structuredContent so yes it should work as described in the doc

docs/api-reference/register-tool.mdx
harijoe: Removed, chaining isnt required

docs/api-reference/register-tool.mdx
harijoe: I agree but rework will be for another PR

docs/concepts/type-safety.mdx
harijoe: I agree, but will do in another PR dedicated to improve documentation.

packages/core/tsconfig.base.json
harijoe: Well, rootDir is by default calculated based on where the input files are so it's not a bad idea to enforce it as we recommend a given file structure;
outDir is by default the current directory : we do need to explicitely state it to build in a dist directory

packages/core/package.json
harijoe: Only for 2 days… 😄
EOL is April 30, 2026

docs/api-reference/register-tool.mdx
harijoe: Indeed it's misleading, I'll clarify

docs/api-reference/mcp-server.mdx
```diff
@@ -235,45 +234,49 @@ const server = new McpServer(
...
 )
-  .registerWidget("search-hotels", {
-    description: "Search for hotels",
-  }, {
-    inputSchema: {
-      city: z.string().describe("City to search"),
-      checkIn: z.string().describe("Check-in date"),
-      checkOut: z.string().describe("Check-out date"),
-      guests: z.number().optional().default(2),
+  .registerTool(
```
paulleseute: i find this code snippet very lengthy

docs/concepts/data-flow.mdx
```diff
@@ -15,83 +15,98 @@ description: "Learn how data moves between the host, your MCP server, and widget
...
-});
+// Plain tool: no UI, returns text
+server.registerTool(
+  { name: "calculate", inputSchema: { /* ... */ } },
+  async (args) => {
+    return { content: "Result: 42" };
+  },
+);
+
+// View tool: has UI, returns structured data
```
paulleseute: tool with view

docs/concepts/data-flow.mdx
```diff
@@ -15,83 +15,98 @@ description: "Learn how data moves between the host, your MCP server, and widget
...
-- **`structuredContent`**: Typed JSON data surfaced to your widget and the host
-- **`_meta`**: Delivered only to the widget and hidden from the model
+- **`structuredContent`**: Typed JSON data surfaced to your view and the host
+- **`_meta`**: Delivered only to the view and hidden from the model
 
-## Tools vs Widgets
+## Tools and views
 
-Before diving into the data flow, understand the difference between tools and widgets in Skybridge:
+Before diving into the data flow, understand the difference between plain tools and tools returning a view. Both use `registerTool`, differentiated by the optional `view` field:
```
paulleseute: they are not returning a view per se, so i'd stick to "tool with a view"

docs/quickstart/add-to-existing-app/server.mdx
```diff
@@ -33,44 +33,39 @@ import { McpServer } from "skybridge/server";
...
-└── web/
-    └── src/
-        └── widgets/
-            ├── my-widget.tsx  # Widget component
-            └── other-widget.tsx
+├── src/
+│   ├── server.ts             # MCP server entry
+│   └── views/
+│       ├── my-view.tsx     # View component
+│       └── other-view.tsx
```
paulleseute: this is a bit misleading, what is other-view ? i'd just remove it

paulleseute (approved):

docs/api-reference/mcp-server.mdx
harijoe: I removed it. Also I believe we don't need to chain to make typings work. I'll give it a check because if we do, it's a serious flaw

PR #711, opened 2026-04-28
fix: support concurrent /mcp requests

paulleseute:
The /mcp middleware crashes on concurrent HTTP requests with "Already connected to a transport". Every request calls server.connect(transport) on a single shared McpServer, but the SDK's Protocol only allows one active transport at a time, so any second request that arrives before the first one finishes blows up with a 500.

Fix is to give each request its own underlying SDK Server bound to its own transport, while sharing the user's tool/resource/prompt handler maps with the main McpServer by reference. That's the SDK's recommended stateless pattern and avoids the slow mutex workaround from the issue (which serializes everything through the shared instance).

Added a connectStatelessTransport method on McpServer that builds the per-request Server, and switched the express middleware to use it. The existing connect override is unchanged so stdio and other long-lived transports still work the same way.

Reproduction test fires 10 concurrent tools/call requests against a 50ms slow tool. On main: 1 success, 9 × 500. With the fix: 10 × 200, no errors logged.

### DIFF
--- a/packages/core/src/server/express.test.ts
+++ b/packages/core/src/server/express.test.ts
@@ -1,7 +1,7 @@
 import http from "node:http";
 import type { ErrorRequestHandler, RequestHandler } from "express";
 import { afterEach, describe, expect, it, vi } from "vitest";
-import type { McpServer } from "./server.js";
+import { McpServer } from "./server.js";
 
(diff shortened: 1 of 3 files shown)


harijoe (approved):
LGTM, well done 👍

PR #721, opened 2026-04-29
fix: log error when view is missing default export

### LINKED ISSUE #534: View validation warnings don't appear in Dev (running "skybridge dev")
### Context

When running `skybridge dev`, if a view file is missing or does not have any component with `export default`, the app fails silently (nothing is displayed to the user)

The `hasDefaultExport` function in the Vite plugin does check for missing export default , but it uses this.warn() in the transform hook, which only surfaces during production builds (npm run build).

### Expected behaviour

Developers should get clear feedback in dev when a view file is missing export default or is missing

### Suggested fix

I can see two options here:

1. Dev server validation: run the existing validation checks at skybridge dev startup and on file change.
2. ESLint plugin (eslint-plugin-skybridge): lint rules that catch these issues directly in the editor. This would create a shorter feedback loop, particularly helpful for AI coding agents that can run eslint and parse the output. Heavier to set up though.

paulleseute:
if a view file under src/views is missing `export default`, skybridge dev silently drops it — your widget never mounts and no log tells you why. the existing this.warn() in the vite plugin lives in the transform hook, which never fires for these files because they're filtered out of the entry list before transform runs.

fix: emit a logger.warn from the plugin's rescan path. dedupe so a rescan triggered by an unrelated edit doesn't re-spam, and log a "resolved" line when the file comes back. covers initial scan + add/change/unlink.

considered a fancier route (react hook + ink panel + chokidar in the parent cli) but it's a lot of plumbing for one log line.

Closes https://github.com/alpic-ai/skybridge/issues/534

Screenshot 2026-04-29 at 10 23 08

### DIFF
--- a/packages/core/src/web/plugin/plugin.ts
+++ b/packages/core/src/web/plugin/plugin.ts
@@ -1,6 +1,7 @@
 import { isAbsolute, relative, resolve } from "node:path";
 import type { Plugin, ViteDevServer } from "vite";
 import {
+  assertUniqueViewNames,
   type DiscoveredView,
   discoverViewsSync,
   scanViewsSync,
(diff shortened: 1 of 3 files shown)


harijoe (approved):
LGTM

PR #724, opened 2026-04-30
feat: add /tunnel to dev server

paulleseute:
today the tunnel only works from the cli. nothing else can start or observe it. we want other consumers such as the devtools to be able to drive it.

architecture: tunnel subprocess ownership lives on the cli. the dev server has a small proxy router that forwards /tunnel routes to the cli over loopback http so devtools can just fetch ${host}/tunnel.

implementation: a TunnelManager + native node:http handler on the cli, listening on 127.0.0.1:<random>. cli passes the port to the dev server via __TUNNEL_CONTROL_PORT env var. the dev server's /tunnel routes (POST start, DELETE stop, GET /tunnel/events SSE) are an express router that proxies to that upstream — POST/DELETE forward as plain json, SSE pipes the readablestream through. the alpic subprocess and its websocket stay up across nodemon restarts because the cli stays up.

the ink hook always subscribes to /tunnel/events regardless of --tunnel, so the cli ui updates from external triggers. --tunnel only controls whether to auto-POST on boot. the hook reconnects SSE indefinitely so dev-server boot and nodemon restarts heal automatically.

to test, start a dev server then:
```
# subscribe to state stream (leave running in another terminal)
curl -N http://localhost:3000/tunnel/events

# start the tunnel
curl -i -X POST http://localhost:3000/tunnel

# stop
curl -i -X DELETE http://localhost:3000/tunnel
```

note: a first iteration of this PR put the TunnelManager inside the user's dev server; it shipped on the branch before we noticed nodemon was killing alpic on every save. the cli-owned + proxy design here is the fix.

<h3>Greptile Summary</h3>

This PR moves tunnel subprocess ownership to the CLI and exposes it to other consumers (devtools, curl) through a small HTTP proxy. The `TunnelManager` lives in the CLI, a loopback control server serves `/__skybridge/tunnel` routes, and the dev server's Express app proxies those routes through to it via `__TUNNEL_CONTROL_PORT`. All four P1 issues from the previous review thread are addressed in this iteration: the timeout message is no longer clobbered on close (child nulled before kill), `observe()` starts unconditionally, `postUntilStarted` retries indefinitely, and the SSE drain-hang on client disconnect is fixed with the inner `res.once("close")` race guard.

Safe to merge — no P0/P1 findings; all prior P1 issues from the review thread are resolved.

All critical issues from the previous round are fixed. The two remaining comments are P2 style concerns (optional chaining consistency and variable name shadowing) that do not affect correctness or safety.

packages/core/src/cli/tunnel-control-server.ts (closeAllConnections without optional chaining), packages/core/src/server/tunnel-proxy-router.ts (inner onClose shadowing)

`````markdown
Fix the following 2 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 2
packages/core/src/cli/tunnel-control-server.ts:37
`server.closeAllConnections()` was added in Node.js 18.2.0. The test file already guards against older runtimes with `server.closeAllConnections?.()`, but the production path calls it unconditionally — on Node < 18.2.0 this throws and the error falls into the `.catch` in `dev.tsx`, logging a spurious message before exit. Adding the optional chain keeps both paths consistent.

```suggestion
        server.closeAllConnections?.();
```

### Issue 2 of 2
packages/core/src/server/tunnel-proxy-router.ts:91-102
**`onClose` variable shadows the outer `onClose`**

Inside the backpressure `Promise`, a new `const onClose` is declared (line 96) that shadows the outer `const onClose = () => controller.abort()` (line 44). When `onDrain` runs and calls `res.off("close", onClose)`, it correctly removes the *inner* Promise-scoped listener — so there's no actual bug. However, the name collision makes the cleanup intent hard to follow and could mislead future editors into thinking the outer `req` listener is being removed here. Renaming the inner pair (e.g., `onResClose` / `onResFlush`) would make the two concerns unambiguous.

`````

</details>

<sub>Reviews (6): Last reviewed commit: ["fix: namespace, kebab-case"](https://github.com/alpic-ai/skybridge/commit/bcab39590399c6ef09cd1b1eebb80ccf9214c38b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30303010)</sub>

### CODE REVIEW

File: packages/core/src/cli/tunnel.ts [hunk 1]
```diff
@@ -0,0 +1,204 @@
...
+  subscribe(listener: (state: TunnelState) => void): () => void {
+    listener(this.state);
+    this.on("state", listener);
+    return () => {
+      this.off("state", listener);
+    };
+  }
+
+  start(): void {
+    if (this.state.status === "starting" || this.state.status === "connected") {
```
harijoe, line 89: No warning instead of returning silently?

packages/core/src/cli/tunnel-handler.test.ts
harijoe, line 1: file names in the CLI currently follow the kebab-case, better keep it consistent for now
- tunnelHandler
- tunnelProxyRouter

packages/core/src/cli/tunnelHandler.ts
```diff
@@ -0,0 +1,53 @@
+import type { IncomingMessage, ServerResponse } from "node:http";
+import type { TunnelActivity, TunnelManager, TunnelState } from "./tunnel.js";
+
+export function createTunnelHandler(manager: TunnelManager) {
+  return (req: IncomingMessage, res: ServerResponse): void => {
+    if (req.url === "/tunnel" && req.method === "POST") {
```
harijoe: Not very likely but still, `/tunnel` could conflict with user code I believe, maybe we could use a namespace like `/__skybridge/tunnel` ?

harijoe (approved):
A few comments but otherwise good to go, nice PR 👍

packages/core/src/cli/tunnel.ts
paulleseute: silent return is intentional because start() is idempotent on purpose: ink hook POSTs on mount, postUntilStarted retries on transient failures, and curl/devtools may also POST; all of them should coalesce into "started" without noise. the response body returns the current state if the caller cares.

PR #725, opened 2026-04-30
Migrate examples to new api

harijoe:
Split by commit:
- Migrate the examples to the new API
- Removed typecheck from ci as we already build the different packages
- Build examples on CI (to replace the typecheck) + Add an override to prefer local packages as we hardcode the pinned version for examples even though we want them to be build against the codebase, not the last published version

_typecheck was creating an issue as the component type file is only generated when running `skybridge dev` and `skybridge build` commands. The ci was not running either of these commands so it didn't have the generated types_

**Note:** The "Type Check" github check must be removed for the repo before this PR is merged


.github/workflows/ci.yml
```diff
@@ -17,12 +17,12 @@ jobs:
         include:
           - name: Unit Tests
             command: pnpm run test:unit
-          - name: Type Check
-            command: pnpm run test:type
```
fredericbarthelet, line 21: don't you want to still run test:type on core and cli ?

examples/auth-auth0/src/server.ts
```diff
@@ -94,6 +84,13 @@ const server = new McpServer(
         openWorldHint: true,
         destructiveHint: false,
       },
+      view: {
+        component: "search-coffee-paris",
```
fredericbarthelet, line 88: no component import finally?

harijoe: Nop, it made the inner workings way too complicated for the benefits and the typesafety already solves the problem we were trying to address in the first place (guaranteeing tool-component connection)

.github/workflows/ci.yml
harijoe: I think running build makes more sense everywhere : it includes types errors
We could keep test:type but I'm not sure of the value it brings

package.json
```diff
@@ -29,6 +28,7 @@
   },
   "pnpm": {
     "overrides": {
+      "skybridge": "workspace:*",
```
fredericbarthelet, line 31: Are you sure this won't be carried over to published packages?

package.json
```diff
@@ -8,7 +8,6 @@
     "test": "pnpm -r --if-present test",
     "format": "pnpm -r --if-present format",
     "test:unit": "pnpm -r --if-present test:unit",
-    "test:type": "pnpm -r --if-present test:type",
```
fredericbarthelet, line 11: not useful ?

fredericbarthelet (approved):
LGTM, just a few small comments

.github/workflows/ci.yml
fredericbarthelet: Type checks is supposed to be build without file emission. What else does it bring g to the table in addition to type checks ?

harijoe: The current build command has three steps :
- Scanning the views _<- it still nice to test that it works correctly_
- Compiling the server (which is running tsc -b --force)
- Running vite build _<- vite may do more than just transpiling the files?_

package.json
harijoe: I will double check before releasing

package.json
harijoe: Not anymore if we choose to remove all `test:type` scripts

PR #731, opened 2026-05-04
feat: support cloudfare worker

paulleseute:
This is a loose adaptation of https://github.com/alpic-ai/skybridge/pull/568. Each commit fixes one runtime gap that surfaced when bundling Skybridge into a Cloudflare Worker.

**Inline `process.env.NODE_ENV` at the express dev guards.** Dev-only imports (devtools, the views dev server) leaked into the worker bundle because the env check sat behind a local variable, defeating bundler dead-code elimination. Inlining the literal at each guard lets the whole dev branch vanish from the worker.

**Expose a Workers fetch handler from `server.run()`.** Workers expect a fetch handler exported from the entry, and `server.run()` returned `void`. The framework now wraps the express app through `cloudflare:node`'s `httpServerHandler` after the listen resolves; on Node, the import throws and we fall through to the existing shutdown wiring.

**Load `package.json` via JSON import.** Reading the framework version used `createRequire(import.meta.url)`, which crashes on workerd because `import.meta.url` is undefined in the bundle. Switched to a static JSON import — Node 22+ ESM and esbuild both handle it natively.

**Precompile Handlebars templates at framework build time.** The view HTML templates were read from disk at module load, which relies on `readFileSync` plus `import.meta.url` — neither works on workerd. The framework now precompiles its templates at its own build time and imports them as plain ESM.

**`enableJsonResponse: true` on the MCP transport.** `tools/call` returned 200 with an empty body because the SDK's default SSE chunks written after an async tool handler don't survive workerd's http bridge. Skybridge's stateless transport never streams server-initiated messages anyway, so a single JSON body is the right default everywhere — not just on workerd.

**Vite manifest as a JS module + `McpServer.setViteManifest()`.** Production view rendering needs Vite's manifest to resolve hashed asset paths, and the framework was reading it from disk at runtime. `skybridge build` now emits the manifest as a JS module; the user entry imports it (NODE_ENV-gated so `skybridge dev` still works without a build) and hands it to the framework before `run()`.

**Emit Cloudflare `_redirects` + `_headers` from `skybridge build`.** Cloudflare's static-asset handler bypasses the worker entirely and maps URLs to files literally — Express's mount-strip and `cors()` middleware don't apply, so the asset URL prefix and CORS need explicit declarations. The build emits both files when a wrangler config is detected, so non-Cloudflare projects don't ship them.

To test, add a `wrangler.jsonc` to a Skybridge app:

```jsonc
{
  "name": "my-app",
  "main": "dist/server.js",
  "compatibility_date": "2025-09-01",
  "compatibility_flags": ["nodejs_compat"],
  "assets": { "directory": "dist/assets" },
  "define": { "process.env.NODE_ENV": "\"production\"" }
}
```

Then `npx skybridge build && npx wrangler dev` (or `wrangler deploy`).

<h3>Greptile Summary</h3>

- Adds Cloudflare Workers support across the core package: precompiled Handlebars templates, a Workers `fetch` handler from `server.run()`, static JSON import for version, `enableJsonResponse` on the MCP transport, and a `setViteManifest()` escape hatch so the user entry can inject the Vite manifest without filesystem access.
- The three new `commandSteps` entries (manifest module, `_redirects`, `_headers`) run unconditionally for every `skybridge build`. The `_redirects` rule would be interpreted by Netlify as a live URL rewrite, breaking non-Cloudflare deployments there; the PR description says these files should only emit when a wrangler config is detected.
- `server.run()` now returns `{ fetch }` on workerd (detected via `navigator.userAgent === \"Cloudflare-Workers\"`) and `undefined` on Node, and the template entry re-exports it as the module default — the correct wiring for a Cloudflare Workers entry point.

Not safe to merge as-is — the unconditional Cloudflare file emission actively breaks Netlify deployments for existing users.

A single P1 defect in the build pipeline: _redirects and _headers are written for every build regardless of target platform. Netlify treats _redirects as live rewrite rules, meaning any existing user deploying to Netlify would have an unintended /assets/assets/* rewrite applied after upgrading. The core Workers wiring (navigator check, httpServerHandler, setViteManifest, template precompilation) looks correct.

packages/core/src/commands/build.tsx — the Cloudflare-specific step block needs a wrangler-config guard matching the behavior stated in the PR description.

`````markdown

---

### Issue 1 of 1
packages/core/src/commands/build.tsx:73-99
**Cloudflare-only output files emitted unconditionally for every build**

The `_redirects` and `_headers` steps have no Cloudflare-config guard, so they run for every `skybridge build`. The PR description explicitly says "The build emits both files when a wrangler config is detected, so non-Cloudflare projects don't ship them." Without that guard, a project deployed to Netlify will have the `/assets/assets/* /assets/:splat 200` rewrite silently applied by Netlify's own `_redirects` processor, which actively rewrites asset URLs and breaks the Netlify deployment. The emitted `dist/vite-manifest.js` is similarly unnecessary for plain Node deployments but is benign there.

`````

</details>

<sub>Reviews (3): Last reviewed commit: ["fix: drop @ts-nocheck on precompiled tem..."](https://github.com/alpic-ai/skybridge/commit/80df5a49fe92445e192f703ab13017cfe269d286) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30701668)</sub>

### CODE REVIEW

File: docs/quickstart/deploy.mdx [hunk 1]
```diff
@@ -48,6 +48,81 @@ Alpic also provides its own [MCP App](https://docs.alpic.ai/features/mcp-server)
...
+    "process.env.NODE_ENV": "\"production\""
+  }
+}
+```
+
+Each setting is load-bearing:
+
+- **`compatibility_date >= 2025-09-01`** + **`nodejs_compat`** — enables `cloudflare:node`'s `httpServerHandler`, which Skybridge uses to bridge the Express app to the Workers fetch event.
+- **`assets.directory`** — points at the views built by `skybridge build`. Cloudflare serves these at the edge before requests reach your worker.
+- **`define.process.env.NODE_ENV`** — forces production mode even under `wrangler dev`. Without it, wrangler defaults to `development` locally and Skybridge's dev tooling (Vite, the devtools server) gets pulled into the worker bundle, where neither runs.
```
fredericbarthelet, line 76: are there any instances where you might want to run `wrangler dev` on your machine? `skybridge dev` is still the default, whatever the deploy runtime. Does this settings also affects `wrangler deploy`?

docs/quickstart/deploy.mdx
```diff
@@ -48,6 +48,81 @@ Alpic also provides its own [MCP App](https://docs.alpic.ai/features/mcp-server)
...
+yarn build
+yarn dlx wrangler deploy
+```
+```bash bun
+bun run build
+bunx wrangler deploy
+```
+</CodeGroup>
+
+### Test locally
```
fredericbarthelet, line 99: I'm not sure about this recommendation

packages/core/scripts/precompile-templates.mjs
```diff
@@ -0,0 +1,36 @@
...
+// import them as plain ESM and we drop `readFileSync` + `fileURLToPath` from
+// the production code path (workerd bundles have neither a usable
+// `import.meta.url` nor the `.hbs` source files).
+import { readdirSync, readFileSync, writeFileSync } from "node:fs";
+import { basename, join } from "node:path";
+import Handlebars from "handlebars";
+
+const ROOT = join(import.meta.dirname, "..");
+const TEMPLATES_DIR = join(ROOT, "src", "server", "templates");
+const OUTPUT_FILE = join(ROOT, "src", "server", "templates.generated.ts");
```
fredericbarthelet: why produce .ts file in precompile build step? You should be fine producing solely .js file and solves at the same time the missing typings issue

packages/core/src/commands/build.tsx
```diff
@@ -1,11 +1,18 @@
...
+import path from "node:path";
 import { Command } from "@oclif/core";
 import { Box, render, Text } from "ink";
 import { useEffect } from "react";
 import { Header } from "../cli/header.js";
 import { type CommandStep, useExecuteSteps } from "../cli/use-execute-steps.js";
 import { scanAndWriteViewsDts } from "../web/plugin/scan-views.js";
 
+function hasWranglerConfig(): boolean {
+  return ["wrangler.jsonc", "wrangler.toml", "wrangler.json"].some((f) =>
```
fredericbarthelet:
```suggestion
  return ["wrangler.jsonc", "wrangler.toml", "wrangler.json"].some((file) =>
```
Avoid one-letter variable names

packages/core/src/commands/build.tsx
```diff
@@ -49,6 +56,56 @@ export const commandSteps: CommandStep[] = [
...
+        {
+          label: "Emitting Cloudflare redirects",
+          // Cloudflare's `assets.directory` maps URL → file literally — no
+          // mount-strip like `app.use("/assets", express.static(...))`. Rewrite
+          // `/assets/assets/*` to `/assets/*` before lookup; status 200 =
+          // server-side rewrite, not HTTP redirect.
+          run: () => {
+            const root = process.cwd();
+            writeFileSync(
+              path.join(root, "dist", "assets", "_redirects"),
```
fredericbarthelet: are there downside from producing the `_redirects` and `_headers` file in assets output directory in non-workerd environments? I'd advocate for a consistent build whatever the project context if it doesn't affect other environments

packages/core/scripts/precompile-templates.mjs
```diff
@@ -0,0 +1,36 @@
+// Precompile Handlebars templates at framework build time so the runtime can
+// import them as plain ESM and we drop `readFileSync` + `fileURLToPath` from
+// the production code path (workerd bundles have neither a usable
+// `import.meta.url` nor the `.hbs` source files).
```
fredericbarthelet: you can remove this comment, this is standard practice to precompile Handlebars template :)

packages/core/src/server/server.ts
```diff
@@ -517,17 +518,29 @@ export class McpServer<
...
 
+    // On workerd, bridge the Node http server to a Workers fetch handler.
+    // The specifier is held in a variable to sidestep tsc's module resolution
+    // (`cloudflare:node` only exists under wrangler/workerd).
+    if (
+      typeof navigator !== "undefined" &&
+      navigator.userAgent === "Cloudflare-Workers"
+    ) {
+      const cloudflareNode = "cloudflare:node";
+      const { httpServerHandler } = await import(cloudflareNode);
```
fredericbarthelet, line 540: why in 2 lines?
```suggestion
      const { httpServerHandler } = await import("cloudflare:node");
```

fredericbarthelet: you don't need to test for userAgent, you can simply try to import and swallow error if any error occurs during import (because package is not available)

packages/create-skybridge/template/_gitignore
```diff
@@ -3,4 +3,5 @@ dist/
 .env*
 .DS_store
 *.tsbuildinfo
-.skybridge/
\ No newline at end of file
+.skybridge/
+src/vite-manifest.js
```
fredericbarthelet: vite-manifest should be written in dist output folder

fredericbarthelet (approved):
A few comments, but LGTM :) Thanks @paulleseute !

docs/quickstart/deploy.mdx
paulleseute: you can run wrangler dev if you want to validate locally the prod build you are about to deploy to the cloud

docs/quickstart/deploy.mdx
paulleseute: it's within the "deploy" page, so it meanings testings your build more than anything - i tried make it explicit with the note but maybe we can get rid of the whole section

packages/core/src/server/server.ts
paulleseute: - otherwise typescript complains since it's injected at runtime
- we might want surface errors that are not module import errors ?

PR #735, opened 2026-05-05
Implement auto-open of devtools after skybridge dev

harijoe:
Just launches devtools on `skybridge dev`

<h3>Greptile Summary</h3>

This PR adds automatic browser launch on `skybridge dev` via a new `useOpenBrowser` hook that TCP-polls the dev-server port and calls `open()` once it is reachable. Users can opt out with `--no-open` or the `SKYBRIDGE_OPEN=false` env var, and the docs are updated to cover the new flag plus two previously undocumented flags.

- **`use-open-browser.ts`**: polls `127.0.0.1:{port}` every 200 ms up to a 5 s deadline, then calls `open()`. The `useRef` guard and `cancelled` flag correctly prevent double-opens across re-renders.
- **`dev.tsx`**: adds the `open` boolean flag with `allowNo: true`; the env-var default (`process.env.SKYBRIDGE_OPEN !== "false"`) is evaluated at module load time, which is the correct point in the lifecycle.
- **`docs/api-reference/cli.mdx`**: documents `--no-open`, the `SKYBRIDGE_OPEN` env var, and the pre-existing `--tunnel` / `-v` flags.

Safe to merge — the change is self-contained and adds opt-in/opt-out controls at both the CLI and environment-variable level.

The new hook is well-guarded against double-opens and re-render races. The only unreported finding is a minor socket-resource nit (using socket.end() instead of socket.destroy() for the port probe), which does not affect correctness. The two more substantive concerns (short timeout, silent open() error) were already captured in prior review threads.

No files require special attention beyond the previously noted timeout and error-handling improvements in use-open-browser.ts.

`````markdown

---

### Issue 1 of 1
packages/core/src/cli/use-open-browser.ts:11-14
Using `socket.end()` initiates a graceful half-close and leaves the socket in a half-open state until the remote side responds — during polling this means up to ~25 lingering sockets, and the HTTP dev-server will receive and log empty/malformed request attempts for each probe. `socket.destroy()` immediately releases the OS resource and avoids this noise.

```suggestion
    socket.once("connect", () => {
      socket.destroy();
      resolve(true);
    });
```

`````

</details>

<sub>Reviews (3): Last reviewed commit: ["Add precision to the docs"](https://github.com/alpic-ai/skybridge/commit/4fe19bc9e833413473a0f28a7eef7bbf655c5a03) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30851808)</sub>

### DIFF
--- a/docs/api-reference/cli.mdx
+++ b/docs/api-reference/cli.mdx
@@ -39,10 +39,14 @@ You can also set the port via the `PORT` environment variable:
 PORT=8080 skybridge dev
 ```
 
-Set `SKYBRIDGE_OPEN=false` to disable auto-opening DevTools by default (equivalent to passing `--no-open` on every run):
+If you'd rather never have DevTools opened automatically, export `SKYBRIDGE_OPEN=false` at the shell level (e.g. in your `~/.zshrc`, `~/.bashrc`, or shell profile). It's equivalent to passing `--no-open` on every run, but lets you opt out once globally instead of per-invocation:
 
 ```bash
+# one-off
 SKYBRIDGE_OPEN=false skybridge dev
+
+# persistent (add to your shell profile)
+export SKYBRIDGE_OPEN=false
 ```
 
 ---
--- a/docs/api-reference/cli.mdx
+++ b/docs/api-reference/cli.mdx
@@ -29,11 +29,20 @@ skybridge dev
 | Flag | Description |
 |------|-------------|
 | `-p, --port <number>` | Port to run the server on. Defaults to `3000`, or the next available port if `3000` is in use. |
+| `--tunnel` | Open an Alpic tunnel for remote testing. |
+| `--no-open` | Don't open DevTools in the browser when the server is ready. |
+| `-v, --verbose` | Show tunnel logs. |
 
 You can also set the port via the `PORT` environment variable:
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by paulleseute — COMMENTED** [review]
It's very likely that some users (including me) will find this annoying so I think we should have some kind of mecanism to disable this globally and document it. I guess the idiomatic way is to respect the BROWSER=none flag.

File: docs/api-reference/cli.mdx [hunk 1]
```diff
@@ -29,13 +29,22 @@ skybridge dev
...
 
 ```bash
 PORT=8080 skybridge dev
 ```
 
+Set `SKYBRIDGE_OPEN=false` to disable auto-opening DevTools by default (equivalent to passing `--no-open` on every run):
+
+```bash
+SKYBRIDGE_OPEN=false skybridge dev
+```
```
paulleseute, line 50: we should make clear that env var is to disable this behavior at shell level. having to prepend the dev with SKYBRIDGE_OPEN=false is worse than the --no-open flag

paulleseute: a code group with the command for each major shell could be nice

paulleseute (approved):

PR #745, opened 2026-05-10
fix(capitals): When asking for a non-capital city on the Capital example app, the widget crashes

### LINKED ISSUE #361: When asking for a non-capital city on the Capital example app, the widget crashes
The widget first shows and then crashes when it cannot resolve where to center the map
Handle the case where no city is found. Default to current center of the map and show a small error message on the widget saying the capital was not found.

udaykakade25:
Fixes https://github.com/alpic-ai/skybridge/issues/361

When a non-capital city is requested, the widget previously got stuck in a permanent loading state and showed no error while newer models would suppress the widget entirely due to `isError: true.`

Now returns `isError: false` with allCapitals in _meta (so the map and widget always render) and surfaces the error message as an overlay on the map, visible in both inline and fullscreen modes.

## Before
Screenshot (359)

## After

Screenshot (362)

<h3>Greptile Summary</h3>

Fixes a widget crash when a non-capital city is queried by adding optional chaining for `output?.capital?.name` and enriching the server-side error response with `_meta: { allCapitals }` and `structuredContent: { error: message }` so the map always has data to render.

- **Crash fix**: `output?.capital.name` → `output?.capital?.name` prevents a runtime TypeError when `output` contains an error shape (no `capital` property).
- **Error enrichment**: The catch block now fetches `allCapitals` and returns it in `_meta` alongside `structuredContent: { error }`, so the widget can render the map and surface the error as an overlay rather than hanging in a loading state or being suppressed.
- **Error overlay**: A new red pill overlay is conditionally rendered below the map using a ternary (`data !== undefined ? data?.structuredContent?.error : output?.error`) that correctly clears itself after a successful `travelTo` call.

Safe to merge — the crash fix is targeted and correct, and the error-path enrichment lets the widget render gracefully instead of hanging.

The optional-chaining fix directly addresses the TypeError root cause. The server-side error path now includes all data the widget needs to display the map, and `isError: true` is correctly preserved so the model still receives a proper MCP error signal. The `errorMessage` ternary correctly distinguishes "no travelTo yet" from "travelTo resolved" and clears the overlay after a successful call.

<sub>Reviews (4): Last reviewed commit: ["fix(capitals): restore isError: true on ..."](https://github.com/alpic-ai/skybridge/commit/f31ebc7c04568333c8b34815d6d0155542cb945c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31487429)</sub>


harijoe:
@udaykakade25 Could you address Greptile’s comments? Please either implement the suggested fixes or reply to Greptile’s review comments explaining why you disagree. Thanks!

udaykakade25:
Hey @harijoe,
I have addressed the greptile comments.
Ready to merge, thanks!

harijoe (approved):
Approved, thanks @udaykakade25 for this contribution!

PR #746, opened 2026-05-10
fix(capitals): capital does not show the same capital as the one selected

### LINKED ISSUE #360: In some conditions, in Capitals example app, fullscreen right pane details on a capital does not show the same capital as the one selected
Image

udaykakade25:
Fixes #360 

## Solution

I was not able to trigger or reproduce this issue but claude suggested following ways

### Scenario 1
 1. Open the widget in inline mode. The initial capital shown is whatever the LLM called the
  tool with (e.g. "Paris").
  2. Click any different capital marker on the map (e.g. "Berlin").
  3. As the right pane slides in, it will briefly show Paris (stale output.capital) while the
  map highlights Berlin.

### Scenario 2
  1. Go to fullscreen, travel to "Berlin" — right pane correctly shows Berlin.
  2. Exit fullscreen.
  3. Click a new capital (e.g. "London") — the right pane slides in showing Berlin (data still
  holds Berlin from step 1) while London is selected.

### Scenario 3
  If you click two different capitals from the NearbyList in rapid succession, two travelTo
  network requests race. If the first one (stale) resolves after the second one (current), data
   gets overwritten with the stale result, and the right pane shows the wrong capital
  persistently until the next click.
  
 ## Why this works
 
 pendingCapital is set synchronously in handleCapitalClick — it's part of the same state batch as setDisplayMode("fullscreen"). So on the very first render after clicking, pendingCapital is already truthy, isLoadingCapital is true, and the spinner shows immediately — before the right pane even starts sliding in. By the time the useEffect runs (after paint), travelTo is called, isTraveling becomes true, and pendingCapital is cleared back to null, keeping the spinner continuously.

<h3>Greptile Summary</h3>

This PR fixes a stale-capital flash in the right pane by including `!!pendingCapital` in the `isLoadingCapital` guard. Without this, there was a one-render gap between when a capital was clicked and when `isTraveling` became true, during which the old capital's detail could briefly render.

- The change ensures the spinner is shown synchronously from the click (when `pendingCapital` is set) rather than waiting for the `useEffect` to fire `travelTo` after paint.
- The race condition described in Scenario 3 (two concurrent `travelTo` requests resolving out of order) is not addressed by this change; that would require request cancellation or a sequence-number guard in `useCallTool`.

Safe to merge — single-line, additive change to a loading guard with no side effects on the happy path.

The change is minimal and logically correct: pendingCapital is set synchronously on click and cleared in a useEffect after travelTo starts, so adding it to the loading flag closes the gap where stale data could flash. No existing state transitions are altered, and the spinner is already the intended behaviour for that window.

<sub>Reviews (2): Last reviewed commit: ["Merge branch &#39;main&#39; into fixCapitalsShow..."](https://github.com/alpic-ai/skybridge/commit/293a94989a20212debf26e379d0a36438d199633) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31507990)</sub>


harijoe (approved):

harijoe:
Thanks @udaykakade25 for this analysis of the problem. We will see if it fixes the underlying issue.

PR #752, opened 2026-05-11
Add toolbar actions

harijoe:
<h3>Greptile Summary</h3>

This PR introduces four toolbar action buttons (Tunnel, Playground, Audit, Deploy) into the devtools header, backed by a new Zustand store that manages tunnel lifecycle state via SSE. The concerns raised in previous review rounds have been addressed: SSE payloads are now validated with a Zod discriminated union before touching the store, and timer handles for both the copy-to-clipboard and hover-close timers are properly stored in `useRef` and cleared in `useEffect` cleanup callbacks.

- **`tunnel-store.ts`**: New Zustand store with `start`/`stop`/`connect` actions; SSE connection established once at the App root via `useConnectTunnel`, cleaned up on unmount.
- **`toolbar-actions.tsx`**: Four new toolbar components using a shared `HoverPopover` pattern; `TunnelButton` handles all four tunnel states in its popover body; `TunnelLinkButton` defers `window.open` until tunnel reaches "connected" via a `pendingOpen` effect.
- **`vite.config.ts`**: Dev-server proxy added for `/__skybridge` path, forwarding to the configured MCP server origin.

Safe to merge — the changes add self-contained UI components and a Zustand store with no impact on existing data paths.

The two previously flagged runtime defects (unvalidated SSE payloads and leaked setTimeout handles) have both been correctly fixed. SSE frames are now filtered through a Zod discriminated union, and timer handles for hover-close and copy-reset are stored in refs with effect cleanup. No new correctness issues were found in the component logic, store lifecycle, or Vite proxy config.

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;main&#39; into julien/sky-338-..."](https://github.com/alpic-ai/skybridge/commit/606dbc290239e4242036a6b22844ea9e64a149a5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31577450)</sub>

### DIFF
--- a/packages/devtools/src/App.tsx
+++ b/packages/devtools/src/App.tsx
@@ -2,8 +2,11 @@ import { QueryClientProvider } from "@tanstack/react-query";
 import { NuqsAdapter } from "nuqs/adapters/react";
 import AppLayout from "./components/layout/app-layout.js";
 import { queryClient } from "./lib/query-client.js";
+import { useConnectTunnel } from "./lib/tunnel-store.js";
 
 function App() {
+  useConnectTunnel();
+
   return (
     <QueryClientProvider client={queryClient}>
       <NuqsAdapter>
--- a/packages/devtools/src/components/layout/header.tsx
+++ b/packages/devtools/src/components/layout/header.tsx
@@ -4,6 +4,12 @@ import { LogOut } from "lucide-react";
 import { useAuthStore } from "@/lib/auth-store.js";
 import { logout, useServerInfo } from "@/lib/mcp/index.js";
 import { StatusBadge } from "./status-badge.js";
+import {
+  AuditButton,
+  DeployButton,
+  PlaygroundButton,
+  TunnelButton,
+} from "./toolbar-actions.js";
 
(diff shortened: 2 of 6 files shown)


paulleseute (approved):
- IMO the double button does not bring much, the pointer is already placed on the button when the tooltip appears, I think it a bit clumsy to have the user move the pointer to click on another button when it's already correctly placed.
- I'd display the tunnel URL inline in the button group like in the mockup, we have a lot an empty space up there, let's use it to display critical data. Also I'd expect clicking on the (opened) tunnel button to stop the tunnel.
- There's no "pending" state for the tunnel, maybe we could add a yellow badge ?
- I'd start the tunnel on Playground / Audit CAT click. We could update the tooltip to inform it'll open the tunnel, but IMO instructing the user to click a button to unblock another one is cumbersome
- The audit icon isn't obvious to me.
- I like the "button group" of the mockup because it brought a bit of structure to this loose set of button, but strong opinion about it. Also, i think it would feel cleaner aligned to the left.

PR #762, opened 2026-05-13
feat: add useRequestSize

paulleseute:
ask host to resize the widget to arbitrary dimensions. can be tested in claude (at least the height). despite what the [apps sdk reference doc says](https://developers.openai.com/apps-sdk/reference), it's not supported by chatgpt (notifyIntrinsicHeight is a dead api)

### DIFF
--- a/packages/core/src/web/bridges/apps-sdk/adaptor.ts
+++ b/packages/core/src/web/bridges/apps-sdk/adaptor.ts
@@ -7,6 +7,7 @@ import type {
   OpenExternalOptions,
   RequestDisplayMode,
   RequestModalOptions,
+  RequestSizeOptions,
   SendFollowUpMessageOptions,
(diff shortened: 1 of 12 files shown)


packages/core/src/web/bridges/apps-sdk/adaptor.ts
```diff
@@ -74,6 +75,10 @@ export class AppsSdkAdaptor implements Adaptor {
     return window.openai.requestClose();
   };
 
+  public requestSize = async (_size: RequestSizeOptions): Promise<void> => {
+    console.warn("[skybridge] requestSize: not supported on Apps SDK");
```
qchuchu, line 79: It might be when we use mcp apps for the bridge of openai in the future

qchuchu (approved):

PR #775, opened 2026-05-15
fix(core): skip CSS modules in asset base URL transform

### LINKED ISSUE #697: CSS `url()` references are corrupted into invalid JS expressions by asset URL transform
## Summary

The `assetBaseUrlTransform` plugin (`dist/server/asset-base-url-transform-plugin.js`) runs a source-level regex over every module and rewrites any quoted asset path into `(window.skybridge?.serverUrl ?? "") + "/path"`. Because Vite serves `.css` imports as JS modules with the CSS embedded as string literals, the regex also matches inside those strings — and the rewritten output is invalid CSS once it's injected into a `<style>` tag.

Net effect: every `@font-face` / `background-image` / `mask-image` / any `url("/foo.<woff2|woff|svg|png|jpg|jpeg|gif|webp|mp3|mp4|ttf|eot>")` silently fails to load in dev, with the asset URL itself being perfectly reachable.

## Repro

1. Scaffold a fresh Skybridge project.
2. Drop any matching file into `web/public/fonts/Brand.woff2`.
3. Add `web/src/fonts.css`:
   ```css
   @font-face {
     font-family: "Brand";
     src: url("/fonts/Brand.woff2") format("woff2");
   }
   ```
4. `import "./fonts.css"` from the widget entry.
5. `pnpm dev`, open the widget, inspect the font.

**Expected:** "Brand" loads.
**Actual:** font is missing. Fetching the CSS module that Vite serves shows the `src:` line transformed to:

```css
src: url((window.skybridge?.serverUrl ?? "") + "/assets/fonts/Brand.woff2") format("woff2");
```

…which is not valid CSS. `curl http://localhost:3000/assets/fonts/Brand.woff2` returns `200 OK` — the asset is reachable, only the reference is broken.

## Root cause

`dist/server/asset-base-url-transform-plugin.js`:

```js
const assetStringPattern = /(?<!https?:\/\/)(["'`])(\/[^"'`]+\.(svg|png|jpeg|jpg|gif|webp|mp3|mp4|woff|woff2|ttf|eot))\1/g;
code = code.replace(assetStringPattern, (_match, _quote, assetPath) => {
    return `(window.skybridge?.serverUrl ?? "") + "${assetPath}"`;
});
```

The `transform` hook runs over every module's source without checking the module type. Vite wraps CSS imports as JS modules where the stylesheet is passed as a string to `__vite__updateStyle(...)`. The regex matches asset paths inside those string literals and rewrites them into JS concatenation expressions, which are syntactically invalid inside a CSS `url(...)`.

## Impact

- Silent: the browser drops the unresolvable reference and falls back to defaults — no hard error in the console beyond occasional `NetworkError`.
- Surface: any CSS file authored in `web/src/**` that references assets by absolute path.
- Workaround requires moving both the CSS file AND its assets into `public/` and injecting via `<link rel="stylesheet">`, which bypasses the Vite asset pipeline entirely (no hashing, no pruning).

## Suggested fix

The same plugin file already has the correct mechanism via `experimental.renderBuiltUrl` in `web/plugin/plugin.js` — that hook is context-aware and works for both JS and CSS referents. The regex-based source transform looks redundant for most cases and wrong for CSS. Options in order of preference:

1. **Remove the source-level transform**, rely solely on `renderBuiltUrl`. Needs verification that dev-mode runtime URLs still work.
2. **Skip CSS in the `transform` hook**: in `transform(code, id)`, bail when `id` indicates a CSS-ish module (`.css`, `.module.css`, `?direct`, `?inline`, `?raw`, framework-specific variants).
3. **Constrain the regex** to not match inside `url(...)`. Fragile; (1) or (2) is cleaner.

Happy to open a PR if a direction is picked.

## Environment

- `skybridge@0.35.21`
- `vite@^8.0.3`, React 19
- Node 24 (`skybridge dev`)
- macOS 15

serhiizghama:
## Problem

The `assetBaseUrlTransformPlugin` runs a source-level regex over every module and rewrites any quoted asset path into `(window.skybridge?.serverUrl ?? "") + "/path"`. Vite serves `.css` imports as JS modules with the stylesheet embedded as a string literal, so the regex matches asset paths inside that string and emits invalid CSS once the styles are injected into a `<style>` tag.

Net effect: every `@font-face` / `background-image` / `mask-image` referencing `/foo.<woff2|svg|png|...>` from a CSS file imported by a view silently fails to load in dev. The reference becomes:

```css
src: url((window.skybridge?.serverUrl ?? "") + "/assets/fonts/Brand.woff2") format("woff2");
```

…which is not valid CSS. Fetching the asset directly works fine — only the reference is broken.

Fixes #697.

## Solution

Skip CSS-family requests in the transform hook (`.css`, `.scss`, `.sass`, `.less`, `.styl`, `.stylus`, `.pcss`, `.postcss`, `.sss`, plus SFC style blocks like `Foo.vue?vue&type=style&lang.css`). Asset URLs in real CSS are already handled at build time by `experimental.renderBuiltUrl` in `web/plugin/plugin.ts`.

The check mirrors Vite's own `isCSSRequest`, exported as `isCssRequest` so the rule is testable in isolation.

## Testing

```bash
pnpm --filter skybridge test
pnpm --filter skybridge build
```

Added unit tests:
- `isCssRequest` — covers `.css`/`.scss`/`.sass`/`.less`/`.styl` extensions, Vite query strings (`?direct`, `?inline`, `?used`), and SFC style-block ids (`?vue&type=style&lang.css`).
- `assetBaseUrlTransformPlugin` — asserts JS modules still get rewritten and CSS modules carrying `url()` references are returned untouched (reproducer for #697).

<h3>Greptile Summary</h3>

This PR fixes a dev-mode bug where the `assetBaseUrlTransformPlugin` was rewriting `url()` references embedded inside Vite's CSS-as-JS modules, producing invalid CSS in `<style>` tags. The fix adds an `isCssRequest` predicate (mirroring Vite's own CSS detection regex) and short-circuits the `transform` hook for any CSS-family module ID.

- `isCssRequest` uses a non-global regex matching `.css|.scss|.sass|.less|.styl|.stylus|.pcss|.postcss|.sss` either at end-of-string or before `?`, correctly handling Vite query strings and SFC style-block IDs (`Foo.vue?vue&type=style&lang.css`).
- New unit tests cover `isCssRequest` and the plugin's transform hook, including a direct reproducer for issue #697.

Safe to merge — the change is a targeted early-return guard in a single plugin hook, with no impact on JS or non-CSS module paths.

The regex is non-global (no lastIndex hazard), correctly handles every CSS-family extension Vite recognises including SFC style blocks, and the existing JS transform path is completely unaffected. The new test suite reproduces the original bug and verifies both the positive and negative cases of the guard.

<sub>Reviews (2): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/css-asset-u..."](https://github.com/alpic-ai/skybridge/commit/b68658d752d794b755115c4478d833dea1b007e4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32254417)</sub>


harijoe (approved):
Hi @serhiizghama , thanks for your contribution 🎉

As it’s your first time contributing to the Skybridge repository, would you mind introducing yourself on [our Discord](https://discord.gg/awV4gu74wK)? We’d love to get to know our contributors.

PR #795, opened 2026-05-20
feat: revamp create cmd

paulleseute:
- add `create`command: so we can have `npx skybridge create`, it's clean and can be eventually advertised without latest because it's a passthrough to npx create-skybridge@latest. also the --help command work without additionnal `--` with npx (not the case with npm, which show actual npm help-
- add blank template without tool. i did not factorize the file intersection of the two template files because it seems overkill. when we add more it'll make sense. also rework the readmes a bit.
- revamp the actual cli, should work nice for llms and humans.
- simplify sanitation: the cli is used by end user, there's no point into preventing them from having absolute / relative path.
- error handling: subprocesses failures (skills, install) get their 10 last lines logged + manual command to repro provided.

Testing: `cd packages/create-skybridge && pnpm build && node index.js`

Dark mode
image

Light mode, error
image

<h3>Greptile Summary</h3>

This PR revamps the scaffolding CLI (`create-skybridge`) and adds a `skybridge create` passthrough command in the core package that delegates to `npx create-skybridge@latest`.

- **CLI rewrite (`index.ts`)**: replaces sync subprocess calls with an async `spawnAsync` wrapper that keeps the spinner alive, adds a `blank` template alongside the existing `demo` template, introduces `--yes` / `--skip-skills` / `--start` / `--pm` flags for non-interactive use, and simplifies path sanitisation to trim-only.
- **`skybridge create` command**: new oclif `Create` command plus a `SkybridgeHelp` class registered as the global `helpClass`; forwards `--help` to `create-skybridge` and delegates all other command help to oclif's default implementation.
- **CI & tooling**: publish workflow, `bump.js`, and `pnpm-workspace.yaml` updated to cover both `demo` and `blank` template directories.

Safe to merge. The --overwrite fallthrough bug from the previous review is correctly fixed, the non-interactive --yes path is well-gated, and the async subprocess wrapper handles both success and failure cleanly.

All changed paths behave correctly: --overwrite immediately clears the directory without re-prompting, --yes skips all interactive steps while providing sensible defaults, and the skills/install error handling degrades gracefully. No logic defects found on the changed paths.

`````markdown

---

### Issue 1 of 1
packages/create-skybridge/src/index.ts:95-108
**`abort` before `intro` renders an orphaned outro**

The two early-exit calls to `abort` (missing target dir and invalid `--pm`) run before `prompts.intro()` on line 111. `abort` ends with `prompts.outro("Aborted")`, so the terminal gets a closing bottom-border with no matching top-border. Moving the `prompts.intro()` call above these two guards (or replacing `prompts.outro` inside `abort` with `prompts.cancel` when no intro has been shown) would make the output consistent.

`````

</details>

<sub>Reviews (4): Last reviewed commit: ["fix: restore else"](https://github.com/alpic-ai/skybridge/commit/09581d2e85e79e83906b7d7c02fa5f92847e8044) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32992894)</sub>

### CODE REVIEW

File: packages/create-skybridge/src/index.ts [hunk 1]
```diff
@@ -1,293 +1,422 @@
...
+      "--yes",
+      "skills",
+      "add",
+      "alpic-ai/skybridge",
+      "--skill",
+      "chatgpt-app-builder",
+      "--agent",
+      "universal",
+      "claude-code",
+      "--yes",
```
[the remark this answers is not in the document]
paulleseute: not true

packages/create-skybridge/src/index.ts
```diff
@@ -1,293 +1,422 @@
...
+    ]);
+    // skills cli always returns 0 so we look for the success message
+    if (status === 0 && output.includes("Done!")) {
+      Spinner.stop(`Installed coding agent skills`);
+    } else {
+      Spinner.error(`Failed to install coding agent skills:
+\x1b[2m${output}\x1b[0m`);
+      prompts.log.error("Try manually: `npx skills add alpic-ai/skybridge`.");
+    }
```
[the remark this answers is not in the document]
paulleseute: it's the last line

packages/create-skybridge/src/index.ts
```diff
@@ -1,293 +1,423 @@
...
-  bitbucket:user/repo#branch
-
-Examples:
-  create-skybridge my-app
-  create-skybridge my-app --repo github:alpic-ai/skybridge/examples/ecom-carousel
-  create-skybridge . --overwrite --immediate
-`;
+  --blank        scaffold a minimal project without demo tools and views
+  --overwrite    remove existing files if target directory is not empty
+  --pm <choice>  package manager to use (choices: ${PACKAGE_MANAGERS.join(", ")}. default: npm)
```
fredericbarthelet: this should be dictated by the package manager you used to actually run the command. It feels strange to have a `package-lock.json` in my generated project as a result of me running `pnpx skybridge create`. I'd only default to npm in case no package manager was invoked to run in a sandbox the initial skybridge command

packages/create-skybridge/src/index.ts
```diff
@@ -1,293 +1,423 @@
...
+const HELP_MESSAGE = `Usage: skybridge create [path] [options]
+
+⛰ Skybridge v${version} - the fullstack framework for building MCP Apps
+
+Arguments:
+  path           Where the project will be created. Prompted when omitted.
 
 Options:
-  -h, --help              show this help message
-  --repo <uri>            use a git repository instead of the built-in template
```
fredericbarthelet, line 19: are we really killing the repo option here? That was the most versatile option

packages/create-skybridge/src/index.ts
```diff
@@ -1,293 +1,423 @@
...
-        overwrite = res;
-      } else {
-        overwrite = "no";
+    if (argv.overwrite) {
+      emptyDir(targetDir);
+    } else if (yes) {
+      prompts.log.error(
+        `Target directory "${targetDir}" is not empty. Use --overwrite to remove existing files.`,
+      );
+      process.exit(1);
```
fredericbarthelet, line 140: can be an early return. `else if` and `else` make code harder to read

packages/create-skybridge/src/index.ts
```diff
@@ -1,293 +1,423 @@
...
-          : ["--yes", "-a", "universal", "-a", "claude-code"]),
-      ],
-      {
-        stdio: "inherit",
-        cwd: targetDir,
-      },
-    );
+  if (installSkills) {
+    Spinner.start("Installing coding agent skills");
+    const { status, output } = await spawnAsync("npx", [
```
fredericbarthelet, line 275: shouldn't be npx, should be the root package manager used to start the command, no?

packages/create-skybridge/src/index.ts
```diff
@@ -1,293 +1,423 @@
...
+      "--skill",
+      "chatgpt-app-builder",
+      "--agent",
+      "universal",
+      "claude-code",
+      "--copy", // something the symlink fails for some reason
+      "--yes",
+    ]);
+    // skills cli always returns 0 so we look for the success message
+    if (status === 0 && output.includes("Done!")) {
```
fredericbarthelet, line 289: that's very britle, if for whatever reason skills.sh change its success message, all our create fails. I'd check for an error message and consider it failed as an early return check, or even consider the skill installation process successful all the time.

packages/create-skybridge/src/index.ts
fredericbarthelet: you should use the method `getPkgExecCmd` that you deleted to run this spawn process in fact

packages/create-skybridge/src/index.ts
```diff
@@ -1,293 +1,423 @@
...
-      });
-      if (prompts.isCancel(immediateResult)) {
-        return cancel();
-      }
-      immediate = immediateResult;
-    } else {
-      immediate = false;
+  // 7. Package manager — autodetect, prompt only if detection fails (interactive)
+  if (!pm) {
+    pm = detectPackageManager() || "npm";
```
fredericbarthelet, line 300: ok now I understand that helper message that says npm is the default is maybe lacking this auto-sourcing information (rather than just say the default is npm)

packages/create-skybridge/src/index.ts
```diff
@@ -1,293 +1,423 @@
...
+      return "deno";
+    case "npm":
+      return "npm";
+    case "pnpm":
+      return "pnpm";
+    case "yarn":
+      return "yarn";
+    default:
+      return undefined;
+  }
```
fredericbarthelet, line 390: I'll be honest, I see the intention of this method, but I find it's implementation funny 😅

You should make a typeguard and add a check on the parser:

```
if (!isValidPackageManager(args.pm)) { 
 // throw...
}

let pm = args.pm

function isValidPackageManager(value): value is PackageManager {
  return PACKAGE_MANAGERS.includes(value)
}
```

WDYT?

packages/create-skybridge/templates/blank/Dockerfile
fredericbarthelet, line 1: should the dockerfile be part of the generated repo? I was expecting it in the docs on the BYOC section, but it being here, and not being used differs from the decision of not putting the `.wrangler` by default and only including it in the docs.

fredericbarthelet (approved):
A few comments, but look good to me otherwise, good work :)

packages/create-skybridge/src/index.ts
paulleseute: i think that PACKAGE_MANAGERS is not an array of string and value is a string so typescript won't like it

packages/create-skybridge/src/index.ts
paulleseute: at this point it's a glorified git clone and i doubt anyone is using this. we are planning to add curated templates with an explicite template flag or similar.

packages/create-skybridge/src/index.ts
paulleseute: the skill installation is non blocking even when it fails. with this implementation worst case scenario "done" disappearing from the success message and we inform user they should install it manually, i think it's acceptable

packages/create-skybridge/src/index.ts
paulleseute: what is the advantage of using the user preferred package manager here ? from my understanding we're just running some script and npx virtually always ship with node so i don't see the benefit of having the getPkgExecCmd bit

packages/create-skybridge/templates/blank/Dockerfile
paulleseute: it'd say it's a bit different, .wrangler is created by cloudfare cli. we both support docker (via dockerfile) and cloudfare (via actual server implementation) so i'd leave it as is. the docker "support" is demonstrated by the presence of the dockerfile. otherwise i think it does not make sense to say "skybridge can be run within docker" since everything can (you just need the right dockerfile)

packages/create-skybridge/src/index.ts
paulleseute: we need it this, otherwise --overwrite would empty then prompt to empty.

PR #797, opened 2026-05-21
feat: add useDownload

paulleseute:
views can't trigger downloads themselves: sandboxed iframes block `<a download>` and blob URLs, so today there's no way to save a CSV/PDF/etc from a view. adds `useDownload`, backed by the MCP App `ui/download-file` spec method. host shows a save dialog and writes the file. supports inline text, inline blob (base64), and resource links (host fetches large files directly).

different from useFiles which handle ChatGPT's attachment store (cf doc)

### DIFF
--- a/docs/api-reference/use-download.mdx
+++ b/docs/api-reference/use-download.mdx
@@ -1,0 +1,109 @@
+---
+title: useDownload
+description: "Trigger file downloads from a view."
+tag: "MCP Apps"
+---
+
+Views run in sandboxed iframes where direct downloads (`<a download>`, `URL.createObjectURL`) are blocked. The `useDownload` hook returns a `download` function that asks the host to save one or more files to the user's filesystem. The host shows a confirmation dialog before initiating the download.
+
+<Note>
+    This hook is not available Apps SDK. Calling `download` from ChatGPT will return `{ isError: true }`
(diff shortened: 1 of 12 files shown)


docs/api-reference/use-download.mdx
```diff
@@ -0,0 +1,109 @@
...
+function ExportButton({ rows }: { rows: Row[] }) {
+  const { download } = useDownload();
+
+  const handleClick = async () => {
+    const csv = rows.map((r) => `${r.id},${r.name}`).join("\n");
+    const { isError } = await download({
+      contents: [
+        {
+          type: "resource",
+          resource: {
```
harijoe, line 27: We provide some helpers on serverside for resource config (Check content-helpers.ts). Maybe we could expose the relevant ones (I guess embeddedResource and resourceLink) to the web package and use them here in the example ?

harijoe (approved):
One suggestion otherwise LGTM

docs/api-reference/use-download.mdx
paulleseute: i see the point for the server side (especially `text` which is nice sugar) but i think this download stuff is specific/rare enough to not have a set of helpers to construct object that are rather simple

PR #804, opened 2026-05-22
feat(devtools): redesign generated tool input form to match sidebar

harijoe:
Generated tool inputs were ugly — this PR redesigns them to match the sidebar's dense, mono aesthetic. Covers every common input type (text, number, textarea, boolean toggle, single/multi-enum, arrays, nested objects), keeps the json fallback tab, and auto-focuses the first empty text-like input when opening a tool.

Multi-select uses a single-pick combobox that appends to a list of removable rows below — the popover stays open across picks.

Also adds an end-to-end test that fills every field type and asserts the values flow back unchanged in `structuredContent`.

image

<h3>Greptile Summary</h3>

This PR redesigns the generated tool input form in the devtools sidebar to match the dense mono aesthetic, introducing custom rjsf templates and widgets for every common input type (text, number, textarea, boolean toggle, single/multi-select enum, plain arrays, and nested objects), along with an end-to-end fixture and test.

- New `form/` module adds custom rjsf templates (`templates.tsx`), widgets (`widgets.tsx`), shared Tailwind class strings (`styles.ts`), and a `buildFormUiSchema` helper that wires array-of-enum fields to the multi-combobox widget.
- `FormBody` in `tool-item.tsx` is cleaned up to use the new module and gains an auto-focus effect for the first empty input when a tool is opened.
- The `every-input-type` e2e test and server fixture exercise all input variants, though the plain-string array field (`interests`) is not filled in the test and therefore does not exercise `ArrayFieldTemplate` rendering.

Safe to merge only after fixing the ArrayFieldTemplate rendering bug, which would crash any tool that uses a plain-array field.

In `ArrayFieldTemplate`, `{items}` is rendered directly. In rjsf v6, `items` is `ArrayFieldTemplateItemType[]` — plain objects, not React elements — so React throws 'Objects are not valid as a React child' the moment any non-enum array field has at least one entry. The fix is `items.map(item => item.children)`. The enum-array field (`colors`) avoids this path via the CheckboxesWidget override, which is why the e2e test does not surface the issue.

packages/devtools/src/components/layout/tools-list/form/templates.tsx — ArrayFieldTemplate rendering

`````markdown

---

### Issue 1 of 1
packages/devtools/src/components/layout/tools-list/form/templates.tsx:216-217
The comment is incorrect and the rendering will fail at runtime. In rjsf v6, `ArrayFieldTemplateProps.items` is `ArrayFieldTemplateItemType[]` — an array of plain objects, each with a `children: ReactElement` property. Rendering `{items}` directly passes plain objects to React, which throws "Objects are not valid as a React child". The correct approach is to call `.map()` and extract `element.children`. In practice this means any non-enum array field (e.g. the `interests: string[]` field in the fixture) crashes when at least one item exists.

```suggestion
      {items.map((item) => (
        <div key={item.key}>{item.children}</div>
      ))}
```

`````

</details>

<sub>Reviews (7): Last reviewed commit: ["Implement Paul&#39;s UI suggestions"](https://github.com/alpic-ai/skybridge/commit/04d89cceb3687fd4645c9b948940159480f705a2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33478975)</sub>


paulleseute (approved):
UI feedbacks:
- select look very big, i'd have something the same same as other text input. also, is placeholder "select..." always enforced ?
- i'd find more harmonious for the description to use non-monospace font, and it could be a tad smaller
- i'm not sure empty array needs to be marked as (empty), the sole presence of the add button seems self explanatory
- list item badges width should match content

PR #806, opened 2026-05-24
feat(devtools): allow collapsing the open tool in the tools list

udaykakade25:
Set the tools accordion to collapsible so clicking the chevron on the open tool closes it. Gate the first-tool auto-open effect behind a mount-only ref so a manual collapse isn't immediately reverted, and normalize Radix's empty-string close signal to null so the URL tool param is cleared. Drop the open-state cursor-default style so the trigger no longer looks non-interactive.

## Demo VIdeo

https://github.com/user-attachments/assets/1546dfe2-310a-4ec3-8837-77af554e4e86

<h3>Greptile Summary</h3>

This PR enables collapsing an open accordion item in the tools list and prevents the first-tool auto-open effect from fighting against a manual collapse.

- `index.tsx`: Adds `collapsible` to the Accordion, introduces a `didInit` ref so the auto-open effect runs at most once (and correctly marks initialization even when a tool is already selected via URL param), and normalizes Radix's empty-string close signal to `null` in `handleToolClick`.
- `accordion-trigger.tsx`: Removes the `data-[state=open]:cursor-default` style so the open trigger continues to look interactive and signals it can be clicked to collapse.

Safe to merge — the changes are small, well-scoped, and the previously identified edge case with URL-preselected tools is correctly handled by the new guard placement.

The didInit.current = true assignment now fires whenever tools.length > 0, regardless of whether openTool is null, so arriving with a URL-selected tool correctly arms the guard before the user can trigger a manual collapse that would previously re-open the first tool. The empty-string-to-null normalization and collapsible prop addition are straightforward and correct.

<sub>Reviews (2): Last reviewed commit: ["Update packages/devtools/src/components/..."](https://github.com/alpic-ai/skybridge/commit/84bf9a081314decd67823086229569371322bb64) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33634467)</sub>

### DIFF
--- a/packages/devtools/src/components/layout/tools-list/accordion-trigger.tsx
+++ b/packages/devtools/src/components/layout/tools-list/accordion-trigger.tsx
@@ -27,7 +27,6 @@ export function AccordionTrigger({
           "type-text-md font-semibold text-foreground text-left",
           "outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 focus-visible:ring-offset-background rounded-sm",
           "disabled:pointer-events-none disabled:opacity-50",
-          "data-[state=open]:cursor-default",
           "[&[data-state=open]>svg]:rotate-90",
           className,
         )}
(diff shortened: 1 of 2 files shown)


harijoe:
Hi @udaykakade25, thanks for the suggestion. We actually considered this behavior during the V1 development, but allowing all tools to be closed would also require clearing the result area on the right, and at the time it didn’t feel necessary.

Could you share a bit more about why you would prefer this behavior over the current one?

harijoe:
Thanks for taking a look at this @udaykakade25 
We won't change the current behavior for now, but if you believe it should be changed, feel free to open a discussion on discord or on an issue. It'd be nice to get other users opinion on this.
Closing it for now.

PR #808, opened 2026-05-25
feat: update auth example

paulleseute:
since 1.0.2 they don't need to import @modelcontextprotocol/sdk anymore. also fixes the auth0 example which was not working with the devtools


fredericbarthelet (approved):
LGTM

PR #820, opened 2026-05-26
Add Vercel deployment support

harijoe:
## Summary

- `skybridge build` emits a Vercel [Build Output API](https://vercel.com/docs/build-output-api) tree under `.vercel/output/` — esbuild-bundled function, static assets, and routing config — alongside the existing Cloudflare/Docker output.
- Runtime: `VERCEL=1` makes `McpServer.run()` return the Express app instead of opening a port.
- No user-owned `vercel.json` needed; deploy with `vercel deploy --prebuilt`.

## Architecture Notes

- **Build Output API over `api/` + `public/`** — the build writes into `.vercel/` (already gitignored by Vercel CLI convention) instead of tracked paths. Routing/CORS live in the emitted `config.json`, so users don't author or maintain a `vercel.json`.
- **esbuild bundling with DCE** — `NODE_ENV` is pinned to `"production"` and `vite` + `@skybridge/devtools` are marked external so dev-only branches (devtools, viewsDevServer) are dead-code-eliminated from the function bundle.

## Verification

Deployed the branch's dev snapshot (`skybridge@0.0.0-dev.9a208db`) to `harijoee/everything` via `vercel deploy --prebuilt`.

URL: https://everything-qfktxwmzz-harijoee.vercel.app

8/8 HTTP assertions green:

- `POST /mcp initialize` → 200, valid `serverInfo`
- `POST /mcp tools/list` → `show-everything` present
- `POST /mcp resources/list` → view URIs returned, CSP rewritten to deploy host
- `POST /mcp resources/read` → HTML references hashed assets from manifest
- `POST /mcp tools/call show-everything` → 200, structured response + `viewUUID`
- `GET /assets/*.js` → 200 + `Access-Control-Allow-Origin: *` + cache HIT
- `GET /assets/*.css` → 200 + CORS + cache HIT
- `GET /mcp` → 405 (POST-only transport)

Two gaps surfaced that aren't fixed in this PR — both need a docs/example follow-up before users hit them:

- Example apps call `server.run()` fire-and-forget; serverless needs `export default await server.run()` so the bundle has a default handler.
- Example apps don't `setViteManifest(manifest)`; without it, `resources/read` crashes with `ENOENT` on `/var/task/dist/assets/.vite/manifest.json` (the same path Cloudflare can't read).

<h3>Greptile Summary</h3>

This PR adds Vercel serverless deployment support to Skybridge. `skybridge build` now always emits a Vercel [Build Output API](https://vercel.com/docs/build-output-api) tree under `.vercel/output/` (esbuild-bundled function + static assets + routing config), and `McpServer.run()` returns the Express app directly when `VERCEL=1` is set, enabling `vercel deploy --prebuilt` without a user-authored `vercel.json`.

- **`build-helpers.ts`** — new module handles esbuild bundling, `.vc-config.json`/`config.json` emission, and static asset copy into the Build Output API directory layout.
- **`server.ts`** — Vercel branch added to `run()`, returning `this.express` rather than binding a port; a dummy `http.Server` is passed to `createApp` only to satisfy the TypeScript parameter signature (safe because the production `createApp` path doesn't use it).
- **Template `.gitignore`** files updated to ignore `.vercel/` in scaffolded projects.

Safe to merge — all changed paths are well-tested and the deployed verification passed 8/8 HTTP assertions.

The Vercel build output path, esbuild bundling, and Express app return from run() are all covered by dedicated unit/integration tests, and the PR author validated the full deployment. The only finding is a minor env-var cleanup asymmetry in a test's finally block that has no real-world impact given Vitest's default NODE_ENV=test.

No files require special attention; express.test.ts has a trivial cleanup inconsistency in the new Vercel mode test.

`````markdown

---

### Issue 1 of 1
packages/core/src/server/express.test.ts:504-509
The `finally` block restores `NODE_ENV` with a direct assignment, but unlike the `VERCEL` variable, it doesn't handle the case where `prevEnv` is `undefined`. In Node.js, `process.env.X = undefined` stringifies to `"undefined"`, which would leave `NODE_ENV` set to the literal string `"undefined"` for any subsequent tests in the same worker. Vitest defaults `NODE_ENV` to `"test"`, so this won't bite in normal CI, but the pattern is asymmetric with the `VERCEL` cleanup above it.

```suggestion
      if (prevVercel === undefined) {
        delete process.env.VERCEL;
      } else {
        process.env.VERCEL = prevVercel;
      }
      if (prevEnv === undefined) {
        delete process.env.NODE_ENV;
      } else {
        process.env.NODE_ENV = prevEnv;
      }
```

`````

</details>

<sub>Reviews (11): Last reviewed commit: ["refactor(core): revert build steps to st..."](https://github.com/alpic-ai/skybridge/commit/4208d177874bd352e3b1a1f8931f3bd52f1f9b5a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33956720)</sub>

### DIFF
--- a/docs/quickstart/deploy.mdx
+++ b/docs/quickstart/deploy.mdx
@@ -123,6 +123,60 @@ bunx wrangler dev
 Skybridge's local dev server (`skybridge dev`) and the Cloudflare path are independent. Use `skybridge dev` for fast iteration on Node with HMR; use `wrangler dev` to validate the production worker bundle before shipping.
 </Note>
 
+## Vercel
+
+Skybridge builds emit a Vercel-compatible function entry (`api/mcp.js`) and a static asset tree (`public/assets/`) on every build. You only need to add a `vercel.json` at your project root so Vercel knows not to auto-detect Vite.
+
+### Configure
+
+Create `vercel.json` at your project root:
+
+```json
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "framework": null,
+  "installCommand": "true",
+  "buildCommand": "true",
+  "outputDirectory": "public",
+  "rewrites": [{ "source": "/(.*)", "destination": "/api/mcp" }],
+  "headers": [
+    {
+      "source": "/assets/(.*)",
+      "headers": [{ "key": "Access-Control-Allow-Origin", "value": "*" }]
+    }
+  ]
+}
+```
+
+Each field is load-bearing:
+
+- **`framework: null`** — without this Vercel detects `vite.config.ts` and runs `vite` instead of your function.
+- **`installCommand` / `buildCommand: "true"`** — no-ops, so Vercel's sandbox doesn't re-run install/build and overwrite the pre-built `public/assets/`.
+- **`outputDirectory: "public"`** — Vercel serves `public/assets/*` via its CDN (filesystem precedence over rewrites keeps `/assets/*` off the function).
+- **`rewrites`** — every other path lands on `api/mcp.js`, which re-exports the Express app from `dist/server.js`.
+- **`headers`** — CORS on `/assets/*` for view iframes loading bundled JS/CSS.
+
+### Build and deploy
+
+<CodeGroup>
+```bash npm
+npm run build
+npx vercel deploy
+```
+```bash pnpm
+pnpm build
+pnpm dlx vercel deploy
+```
+```bash yarn
+yarn build
+yarn dlx vercel deploy
+```
+```bash bun
+bun run build
+bunx vercel deploy
+```
+</CodeGroup>
+
 ## Docker
 
 Projects created with the [Skybridge scaffolder](/quickstart/create-new-app#scaffold-your-project) include a multi-stage `Dockerfile` so you can self-host on any container platform:
--- a/docs/quickstart/deploy.mdx
+++ b/docs/quickstart/deploy.mdx
@@ -125,57 +125,32 @@ Skybridge's local dev server (`skybridge dev`) and the Cloudflare path are indep
 
 ## Vercel
 
-Skybridge builds emit a Vercel-compatible function entry (`api/mcp.js`) and a static asset tree (`public/assets/`) on every build. You only need to add a `vercel.json` at your project root so Vercel knows not to auto-detect Vite.
-
-### Configure
-
-Create `vercel.json` at your project root:
-
-```json
-{
-  "$schema": "https://openapi.vercel.sh/vercel.json",
-  "framework": null,
-  "installCommand": "true",
-  "buildCommand": "true",
-  "outputDirectory": "public",
-  "rewrites": [{ "source": "/(.*)", "destination": "/api/mcp" }],
-  "headers": [
-    {
-      "source": "/assets/(.*)",
-      "headers": [{ "key": "Access-Control-Allow-Origin", "value": "*" }]
-    }
-  ]
-}
-```
-
-Each field is load-bearing:
-
-- **`framework: null`** — without this Vercel detects `vite.config.ts` and runs `vite` instead of your function.
-- **`installCommand` / `buildCommand: "true"`** — no-ops, so Vercel's sandbox doesn't re-run install/build and overwrite the pre-built `public/assets/`.
-- **`outputDirectory: "public"`** — Vercel serves `public/assets/*` via its CDN (filesystem precedence over rewrites keeps `/assets/*` off the function).
-- **`rewrites`** — every other path lands on `api/mcp.js`, which re-exports the Express app from `dist/server.js`.
-- **`headers`** — CORS on `/assets/*` for view iframes loading bundled JS/CSS.
+`skybridge build` emits a [Build Output API](https://vercel.com/docs/build-output-api) tree under `.vercel/output/` — a bundled serverless function plus the static asset tree, along with the routing config. Deploy it with `vercel deploy --prebuilt`; no `vercel.json` is required.
 
 ### Build and deploy
 
 <CodeGroup>
 ```bash npm
 npm run build
-npx vercel deploy
+npx vercel deploy --prebuilt
 ```
 ```bash pnpm
 pnpm build
-pnpm dlx vercel deploy
+pnpm dlx vercel deploy --prebuilt
 ```
 ```bash yarn
 yarn build
-yarn dlx vercel deploy
+yarn dlx vercel deploy --prebuilt
 ```
 ```bash bun
 bun run build
-bunx vercel deploy
+bunx vercel deploy --prebuilt
 ```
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: packages/core/src/cli/build-helpers.ts [hunk 1]
```diff
@@ -0,0 +1,18 @@
...
+
+export function emitManifestModule(
+  manifestPath: string,
+  outPath: string,
+): void {
+  const manifest = readFileSync(manifestPath, "utf-8");
+  writeFileSync(outPath, `export default ${manifest};\n`);
+}
+
+export const VERCEL_API_ENTRY_CONTENT = `export { default } from "../dist/server.js";\n`;
```
fredericbarthelet: in that case, why not use server.js directly as Vercel function? Re-exporting default export from another file is equivalent to changing the file path only

packages/core/src/cli/build-helpers.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export function emitVercelFunction(root: string): void {
+  const apiDir = path.join(root, "api");
```
fredericbarthelet: if you need `api` in the path, just write a symlink to `server.js` here. The problem with this implementation is that it requires you to gitignore `api` to avoid everyone who build his skybridge app to have non-committed changes resulting from build step

packages/core/src/server/express.ts
```diff
@@ -81,7 +81,9 @@ export async function createApp({
         `Ignoring invalid __TUNNEL_CONTROL_PORT=${process.env.__TUNNEL_CONTROL_PORT}`,
       );
     }
-  } else {
+  } else if (process.env.VERCEL !== "1") {
```
fredericbarthelet: you don't need the static mount, but you don't need the static mount not to be here too. I'd recommend not differentiating VERCEL hosting here. In addition, it simplifies runtime code since you can remove the env variable

packages/core/src/server/server.ts
```diff
@@ -720,10 +720,28 @@ export class McpServer<
    *
    * On Cloudflare Workers / workerd, returns an object exposing `fetch` so
    * the runtime can bridge incoming requests to the Node HTTP server. On
+   * Vercel (`VERCEL === "1"`), returns the Express app directly so the
+   * serverless function entry can call it as a `(req, res)` handler. On
```
fredericbarthelet, line 724: but `fetch` is also the recommended interface for Vercel serverless function. Why change this ?

packages/core/src/commands/build.tsx
```diff
@@ -1,89 +1,93 @@
...
+      run: () => {
+        const root = process.cwd();
+        emitManifestModule(
+          path.join(root, "dist", "assets", ".vite", "manifest.json"),
+          path.join(root, "dist", "vite-manifest.js"),
+        );
+      },
+    },
+    {
+      label: "Copying assets to public/ for Vercel",
```
fredericbarthelet: symlink not enough?

docs/quickstart/deploy.mdx
```diff
@@ -123,6 +123,60 @@ bunx wrangler dev
...
+
+Create `vercel.json` at your project root:
+
+```json
+{
+  "$schema": "https://openapi.vercel.sh/vercel.json",
+  "framework": null,
+  "installCommand": "true",
+  "buildCommand": "true",
+  "outputDirectory": "public",
```
fredericbarthelet: why not use `dist` here ?

packages/core/src/commands/build.tsx
harijoe: Symlinks could be a source of trouble if someone builds on windows for example.
Copying the assets is just a few mb and may be not worth the risk

docs/quickstart/deploy.mdx
harijoe: Fixed by using Vercel Build Output API

packages/core/src/server/server.ts
harijoe: https://github.com/alpic-ai/skybridge/pull/828 proposes to only export the server in server.ts instead of calling run on it.

packages/core/src/cli/build-helpers.ts
harijoe: Should be fixed by using Vercel Build Output API

harijoe:
@greptileai

packages/core/src/cli/build-helpers.ts
```diff
@@ -0,0 +1,91 @@
...
+    `${VERCEL_FUNCTION_NAME}.func`,
+  );
+  const staticAssetsDir = path.join(outputDir, "static", "assets");
+
+  rmSync(outputDir, { recursive: true, force: true });
+  mkdirSync(funcDir, { recursive: true });
+
+  const { build } = await import("esbuild");
+  await build({
+    entryPoints: [path.join(root, "dist", "server.js")],
```
fredericbarthelet, line 57: is this normal? should .vercel/output build use `skybridge build` output as input?

packages/core/src/commands/build.tsx
```diff
@@ -1,89 +1,83 @@
...
-export const commandSteps: CommandStep[] = [
-  {
-    label: "Scanning views",
-    run: async () => {
-      const root = process.cwd();
-      const viewsDir = await resolveViewsDir(root);
-      scanAndWriteViewsDts(root, viewsDir);
+export function buildSteps(): CommandStep[] {
+  return [
+    {
```
fredericbarthelet: I'm not sure to understand the intent behind static steps declaration vs nomad change here. Why the switch?

packages/core/package.json
```diff
@@ -7,6 +7,7 @@
     "url": "https://github.com/alpic-ai/skybridge.git"
   },
   "type": "module",
+  "sideEffects": false,
```
fredericbarthelet, line 10: that's gourmand, are you sure none of the core packages file have side effects?

fredericbarthelet (approved):
A few comments but looks neat :)
I haven't checkout `server.ts` since it's discussed in #832 but Express app starting without binding a port when VERCEL=1 feels like a hack. 

I'm approving this one but I'd prefer to use the same `fetch` convention than Cloudflare for Vercel function in this PR, and keep the discussion wether `server.run` should be user-land code or not separate in #832 and eventually refactor the vercel implementation if adopted here too afterwards. WDYT?

packages/core/src/cli/build-helpers.ts
harijoe: Yes, it's intentional : the user's code imports ./vite-manifest.js  that's generated at dist/vite-manifest.js, not present in src/.

packages/core/package.json
harijoe: I ran an audit, it's okay

PR #827, opened 2026-05-27
fix: derive success status when tool output has no input

### LINKED ISSUE #826: isSuccess from useToolInfo always false
Hi, it seems that since yesterday the `isSuccess` property returned by the `useToolInfo` hook is always `false` when the hook returns an output.

I tried to dig a bit deeper, and it looks like the logic here is no longer working as expected:
https://github.com/alpic-ai/skybridge/blob/98fae76c2ee3cf996fd74376514297185ed7cfd9/packages/core/src/web/hooks/use-tool-info.ts#L60

Since `input` appears to be `null`, it always returns `idle`, even when an output is available. In particular, `window.openai.toolInput` is `null` after calling the tool.

I ran some local tests with a slightly modified version of the "flight-booking" example that prints the `useToolInfo` output:

Image

A quick fix might be to change `deriveStatus` to:

```ts
function deriveStatus(
  input: Record<string, unknown> | null,
  output: Record<string, unknown> | null,
  responseMetadata: Record<string, unknown> | null,
): "idle" | "pending" | "success" {
  if (input === null && output === null) { // <-- additional condition
    return "idle";
  }
  if (output === null && responseMetadata === null) {
    return "pending";
  }
  return "success";
}
```

slackerzz:
this should fix https://github.com/alpic-ai/skybridge/issues/826

<h3>Greptile Summary</h3>

This PR fixes the state-derivation logic so tools with no `input` schema (where the host delivers `null` for `toolInput`) no longer get stuck in the deprecated `"idle"` state — they now correctly flow through `"pending"` → `"success"`. The `ToolPendingState` and `ToolSuccessState` types are updated so `input` is `ToolInput | undefined`, and the hook converts the raw `null` from the host to `undefined` via `input ?? undefined`.

- `deriveStatus` no longer receives `input`; "pending" is returned only when both `output` and `responseMetadata` are `null`, otherwise "success".
- `ToolIdleState` is marked `@deprecated` and `isIdle` is hardcoded to `false` with a note about removal in the next major.
- All type-level (`test-d.ts`) and runtime (`test.ts`) tests are updated to reflect the new pending-first behaviour.
- The `docs/api-reference/use-tool-info.mdx` type-signature blocks are correctly updated to `input: ToolInput | undefined`, but the prose code examples still access `input` without optional-chaining or a null guard, which will produce TypeScript errors under `--strictNullChecks` for users following those patterns.

Safe to merge after updating the documentation examples to guard against undefined input.

The runtime and type changes in use-tool-info.ts are correct and the previous concern about input being non-nullable in the success state is properly resolved. The one concrete problem is in docs/api-reference/use-tool-info.mdx: the type-signature sections were updated to ToolInput | undefined but every prose code example still accesses input directly without optional-chaining or a guard. These examples will now fail TypeScript compilation with --strictNullChecks and could cause runtime crashes for developers who copy them verbatim.

docs/api-reference/use-tool-info.mdx — all example code snippets that access input properties need optional-chaining or an if (!input) guard.

<details><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `packages/core/src/web/hooks/use-tool-info.ts`, line 65-71 ([link](https://github.com/alpic-ai/skybridge/blob/89af7af4105e2947c3ac0da14c85aab5e7a6eca4/packages/core/src/web/hooks/use-tool-info.ts#L65-L71)) 

   <a href="#">P1</a> **Type contract broken for no-input tools in success state**

   After this change, a tool with `input === null` and a non-null `output` is now routed to `"success"`. However, `ToolSuccessState` declares `input: ToolInput` (non-nullable), so the `as ToolState<...>` cast on line 121 silently lies to callers: TypeScript believes `input` is a well-typed object, but at runtime it is `null`. Any consumer that narrows on `isSuccess` and then reads a property of `input` will get a null-dereference at runtime.

   The same applies to `responseMetadata` when a no-input tool returns output but no metadata (`input === null, output !== null, responseMetadata === null` falls straight through to `"success"`, yet `ToolSuccessState.responseMetadata` is non-nullable).

   To fix this soundly, `ToolSuccessState` should allow `input` to be optional (e.g. `input: ToolInput | undefined`) for the no-input variant, or a new discriminated union branch for input-less tools should be introduced alongside the types update.

   `````markdown
   Path: packages/core/src/web/hooks/use-tool-info.ts
   Line: 65-71

   Comment:
   **Type contract broken for no-input tools in success state**

   After this change, a tool with `input === null` and a non-null `output` is now routed to `"success"`. However, `ToolSuccessState` declares `input: ToolInput` (non-nullable), so the `as ToolState<...>` cast on line 121 silently lies to callers: TypeScript believes `input` is a well-typed object, but at runtime it is `null`. Any consumer that narrows on `isSuccess` and then reads a property of `input` will get a null-dereference at runtime.

   The same applies to `responseMetadata` when a no-input tool returns output but no metadata (`input === null, output !== null, responseMetadata === null` falls straight through to `"success"`, yet `ToolSuccessState.responseMetadata` is non-nullable).

   To fix this soundly, `ToolSuccessState` should allow `input` to be optional (e.g. `input: ToolInput | undefined`) for the no-input variant, or a new discriminated union branch for input-less tools should be introduced alongside the types update.

   `````
   </details>

2. `docs/api-reference/use-tool-info.mdx`, line 19-22 ([link](https://github.com/alpic-ai/skybridge/blob/8004b946e9e5348b2985e75e351f96d9037b6882/docs/api-reference/use-tool-info.mdx#L19-L22)) 

   **Documentation examples access `input` without a null guard**

   Now that `input` is typed `ToolInput | undefined` in both `pending` and `success` states, the existing usage examples in this file access it directly (e.g. `input.city` on line 20, `input.query` on line 139, `input.endpoint` on line 175, `switch (input.type)` on line 208, `input.name` on line 239) without any optional-chaining or guard. TypeScript will flag these as errors under `--strictNullChecks`, and users copying the examples will either see type errors or crash at runtime when the host hasn't delivered arguments yet. Each example should use `input?.city`, or add an early `if (!input) return <Skeleton />` guard before accessing `input`.

</details>

`````markdown

---

### Issue 1 of 1
docs/api-reference/use-tool-info.mdx:19-22
**Documentation examples access `input` without a null guard**

Now that `input` is typed `ToolInput | undefined` in both `pending` and `success` states, the existing usage examples in this file access it directly (e.g. `input.city` on line 20, `input.query` on line 139, `input.endpoint` on line 175, `switch (input.type)` on line 208, `input.name` on line 239) without any optional-chaining or guard. TypeScript will flag these as errors under `--strictNullChecks`, and users copying the examples will either see type errors or crash at runtime when the host hasn't delivered arguments yet. Each example should use `input?.city`, or add an early `if (!input) return <Skeleton />` guard before accessing `input`.


`````

</details>

<sub>Reviews (5): Last reviewed commit: ["Merge branch &#39;main&#39; into fix\_derived\_sta..."](https://github.com/alpic-ai/skybridge/commit/8004b946e9e5348b2985e75e351f96d9037b6882) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34118265)</sub>

### CODE REVIEW

**Review by harijoe — DISMISSED** [review]
It looks good! Thanks @slackerzz 
Could you make sure Greptiles scores it at 5/5 ? The typing issue seem relevant.
I'll merge it right after 👌

**Comment by slackerzz** [comment]
@harijoe sure, I'll fix the types

File: docs/api-reference/use-tool-info.mdx [hunk 1]
```diff
@@ -100,10 +100,10 @@ Boolean flags derived from `status` for convenience.
 #### `input`
 
 ```tsx
-input: ToolInput
+input: ToolInput | undefined
 ```
 
-The input arguments passed to the tool. Always available.
+The input arguments passed to the tool. May be `undefined` until the host delivers the tool arguments, or for tools that take no input.
```
fredericbarthelet:
```suggestion
The input arguments passed to the tool. May be `undefined` until the host delivers the tool arguments.
```
`input` is always at least an empty array, you can't have an MCP tool without input at all.

docs/api-reference/use-tool-info.mdx
```diff
@@ -71,9 +71,9 @@ The hook returns an object that varies based on the current status:
   status: "success";
   isPending: false;
   isSuccess: true;
-  input: ToolInput;
+  input: ToolInput | undefined;
   output: ToolOutput;
-  responseMetadata: ToolResponseMetadata;
+  responseMetadata: ToolResponseMetadata | undefined;
```
fredericbarthelet: this shouldn't be affected by the change
```suggestion
  responseMetadata: ToolResponseMetadata;
```

docs/api-reference/use-tool-info.mdx
```diff
@@ -119,7 +119,7 @@ The tool's output data. Only available when `status` is `"success"`.
 responseMetadata: ToolResponseMetadata | undefined
 ```
 
-Additional metadata from the tool response. Only available when `status` is `"success"`.
+Additional metadata from the tool response. Only available when `status` is `"success"`, and may still be `undefined` if the tool returned an output without `_meta`.
```
fredericbarthelet:
```suggestion
Additional metadata from the tool response. Only available when `status` is `"success"`.
```

packages/core/src/web/hooks/use-tool-info.test-d.ts
```diff
@@ -5,10 +5,10 @@ test("useToolInfo - TypeScript typing", () => {
   test("should have correct types when no generic parameter is provided", () => {
     const result = useToolInfo();
 
-    expectTypeOf<"idle" | "pending" | "success">(result.status);
+    expectTypeOf<"pending" | "success">(result.status);
```
fredericbarthelet: we can't remove "idle" from the type as this would be a breaking change. We won't be returning it anymore in practice though

packages/core/src/web/hooks/use-tool-info.test-d.ts
```diff
@@ -5,10 +5,10 @@ test("useToolInfo - TypeScript typing", () => {
...
     expectTypeOf<boolean>(result.isPending);
     expectTypeOf<boolean>(result.isSuccess);
-    expectTypeOf<boolean>(result.isIdle);
+    expectTypeOf<false>(result.isIdle);
```
fredericbarthelet: same here, this is a type breaking change, it must remain boolean, but will always be false in practice

fredericbarthelet (changes requested):
@slackerzz thanks for your contribution. Some breaking changes have been introduced though and changes on responseMetadata that shouldn't be there. WDYT?

docs/api-reference/use-tool-info.mdx
slackerzz: We only check `output` in `deriveStatus` before returning "success", so there’s no guarantee that `responseMetadata` is populated.
It depends on whether we want to trust the specs or not. I’m going to revert it

slackerzz:
@fredericbarthelet you're right, changing the return type of 'useToolInfo` would be a breaking change.

About `responseMetadata`, now the state can be "success" only if we have output, without checking `responseMetadata`, so in theory without `responseMetadata`, however I'm reverting it back to non-optional.

edit: now Greptile complains about it :sweat_smile:

fredericbarthelet (approved):
LGTM, thanks for updating @slackerzz

PR #828, opened 2026-05-27
refactor(core): wrap user entry to hide setViteManifest from src/server.ts

harijoe:
## Breaking change

**No.** All migration scenarios are covered:

- Legacy `src/server.ts` (`export default await server.run()`) still boots in dev — the user file's top-level `await` runs the server, and the dev wrapper's `typeof run === "function"` guard is then false, so no double-run.
- `skybridge build` still emits `dist/server.js`, so existing `wrangler.jsonc` / `Dockerfile` entries pointing at it keep working.
- `skybridge start` prefers `dist/__entry.js` but falls back to `dist/server.js` — upgrading skybridge without rebuilding still works.
- Public API (`McpServer.run`, `setViteManifest`) is unchanged.

No on-disk footprint in user repos: the dev wrapper is shipped inside `node_modules/skybridge/dist/cli/dev-entry.js`. `skybridge dev` execs it via tsx, and it dynamically imports `<cwd>/src/server.ts`.

---

<h3>Greptile Summary</h3>

This refactor moves the server lifecycle wiring (`setViteManifest` + `run()`) out of user-authored `src/server.ts` into a generated `dist/__entry.js` wrapper emitted by `skybridge build`. User files now simply `export default server`, and the framework handles the rest.

- A new `dev-entry.ts` (shipped inside the skybridge package) is now exec'd by nodemon via tsx instead of the user's `src/server.ts` directly; it dynamically imports the user module and calls `run()` only when the export is a live McpServer instance, keeping legacy templates that top-level-await `server.run()` from double-running.
- `build-helpers.ts` extracts `emitManifestModule` and `emitEntryWrapper` (with a tested static content constant) so `build.tsx` gains a clean "Emitting entry wrapper" step and no longer manually inlines manifest logic.
- `start.ts` prefers `dist/__entry.js` with a fallback to `dist/server.js`, so projects built with an older skybridge version still boot without a rebuild.

Safe to merge. The backward compatibility path for both dev and production is carefully handled, and no existing deploy targets break without a rebuild.

The refactor correctly handles all three lifecycle scenarios: new-style servers (export default server, run() called by wrapper), legacy servers (top-level await run(), detected by duck-typing in both dev-entry and __entry.js), and wrangler/Docker deploys (prefer __entry.js, fallback to server.js in skybridge start). The static ENTRY_WRAPPER_CONTENT is tested, the dev wrapper is scoped entirely inside the package with no user-repo footprint, and path quoting in nodemon exec handles spaces correctly.

<sub>Reviews (5): Last reviewed commit: ["refactor(examples): switch to export def..."](https://github.com/alpic-ai/skybridge/commit/45494a17d346aeeb98354e429cbc554d38f7eaef) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34136615)</sub>

### DIFF
--- a/docs/quickstart/deploy.mdx
+++ b/docs/quickstart/deploy.mdx
@@ -59,7 +59,7 @@ Add a `wrangler.jsonc` at your project root:
 ```jsonc
 {
   "name": "your-skybridge-app",
-  "main": "dist/server.js",
+  "main": "dist/__entry.js",
   "compatibility_date": "2025-09-01",
   "compatibility_flags": ["nodejs_compat"],
   "assets": { "directory": "dist/assets" },
--- a/examples/auth-auth0/src/server.ts
+++ b/examples/auth-auth0/src/server.ts
@@ -172,6 +172,6 @@ const server = new McpServer(
     },
   );
 
-server.run();
+export default server;
 
 export type AppType = typeof server;
--- a/examples/auth-clerk/src/server.ts
+++ b/examples/auth-clerk/src/server.ts
@@ -99,6 +99,6 @@ const server = new McpServer(
     },
   );
 
-server.run();
+export default server;
 
 export type AppType = typeof server;
--- a/examples/auth-stytch/src/server.ts
+++ b/examples/auth-stytch/src/server.ts
@@ -133,6 +133,6 @@ const server = new McpServer(
     },
   );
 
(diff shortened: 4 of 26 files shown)


harijoe:
Recreated as #832 after branch rename.

PR #829, opened 2026-05-27
refactor(core): collapse host adaptors into single HostAdaptor with per-method routing

harijoe:
## Summary

- Replace `McpAppAdaptor` + `AppsSdkAdaptor` with a single `HostAdaptor` class that composes both transports internally and routes each `Adaptor` method to the better-fit transport.
- MCP App protocol (`ext-apps`) is the always-present baseline; `window.openai` is captured as an overlay when present and claims `setViewState`, `openModal`, the `display`/`viewState` host-context keys, plus the Apps-SDK-exclusive file ops and `setOpenInAppUrl`. `openExternal` and `sendFollowUpMessage` only route to OAI when the caller passes an Apps-SDK-specific option (`redirectUrl: false`, `scrollToBottom`).
- `getAdaptor()` becomes a memoized factory that probes `window.openai` directly. `window.skybridge.hostType` stops being load-bearing; it remains a diagnostic (mismatch logs a warning).
- `NotSupportedError` exported for Apps-SDK-only methods invoked outside the Apps SDK runtime.
- `closeModal()` added to the `Adaptor` interface (no-op when overlay active, dismisses the in-iframe polyfill otherwise).

## Public API

No changes at `src/web/index.ts` and `src/web/hooks/index.ts` (zero diff vs main). Net additions: `NotSupportedError` class and `closeModal()` method on the `Adaptor` interface.

`useMcpAppContext`, `useAppsSdkContext`, `McpAppBridge`, `AppsSdkBridge` are unchanged and still exported. The two old `Adaptor` class names (`McpAppAdaptor`, `AppsSdkAdaptor`) were `@internal` and are removed.

## Architecture notes

- Per-method routing rules live inline in `HostAdaptor` so the routing IS the documentation; the spec at `docs/superpowers/specs/2026-05-27-bridge-refacto-design.md` has the full table.
- `sendFollowUpMessage("hi")` with no options always routes to MCP, even inside ChatGPT. Apps SDK only wins when `scrollToBottom` is explicitly set. Intentional per the spec.
- localStorage LRU persistence for `viewState` (200-entry cap, viewUUID-keyed) is preserved unchanged from the old MCP path; only used when no Apps SDK overlay is active.

## Test plan

- [x] 276 unit tests pass (`pnpm --filter skybridge test:unit`)
- [x] Build clean (`pnpm build`)
- [x] Lint clean (`pnpm --filter skybridge test:format`)
- [ ] Smoke test inside ChatGPT (overlay active path)
- [ ] Smoke test inside an MCP-app host (baseline path)

## Follow-ups (out of scope)

- `docs/fundamentals/mcp-apps.mdx` still mentions `McpAppAdaptor` in prose at lines 9 and 54. Stale after this PR; worth a follow-up docs commit.
- `persistToLocalStorage` could tighten from `protected` to `private` (no subclass uses it).

<h3>Greptile Summary</h3>

This PR replaces the two separate `McpAppAdaptor` / `AppsSdkAdaptor` implementations with a single `HostAdaptor` class that composes both transports internally, routing each `Adaptor` method to MCP (always-present baseline) or `window.openai` (overlay, when present) according to inline routing rules. `getAdaptor()` becomes a memoized factory that probes `window.openai` at construction time, and `window.skybridge.hostType` is demoted to a diagnostic hint. `NotSupportedError` and `closeModal()` are added to the public surface.

- `HostAdaptor` composes `McpAppBridge` and optionally `AppsSdkBridge`, with per-method routing encoded inline; MCP stores are stable singletons, while the polyfill `display`/`viewState` stores (used in the non-OAI path) are created fresh on each `getHostContextStore` call rather than once at construction time.
- The devtools `View` now spins up both an OAI mock and a new `createMcpHostMock` for the MCP ext-apps protocol, with the cleanup returned from the `useEffect` so listeners are removed on re-render.
- Tests across 30+ files are updated to use the unified adaptor rather than host-type-specific instances.

Safe to merge in the OAI overlay path; the polyfill display/viewState stores returned by `getHostContextStore` in the MCP-only path are recreated as new objects on each call, giving `useSyncExternalStore` a different `subscribe` reference every render and causing spurious re-subscription cycles in `ModalProvider`.

The routing logic and store composition are correct and well-tested across both paths. One structural concern remains: `getHostContextStore` constructs a fresh store object (with a new `subscribe` closure identity) on every call for the polyfill `display` and `viewState` keys. `ModalProvider` calls this on every render, so `useSyncExternalStore` sees a new `subscribe` reference each time and unsubscribes then resubscribes every committed render, opening a narrow window where a `openModal`/`closeModal` notification could fire between the two calls in React 18 concurrent mode.

`packages/core/src/web/bridges/adaptor.ts` — the polyfill store construction inside `getHostContextStore` for the `display` and `viewState` keys.

`````markdown

---

### Issue 1 of 1
packages/devtools/src/components/layout/tool-panel/view/create-mcp-host-mock.ts:113-138
**`sendFollowUpMessage` will silently error in devtools** — The MCP mock's switch statement has no case for the `ui/send-message` request (the method `app.sendMessage()` issues). In the new routing, `sendFollowUpMessage(prompt)` with no options always goes through the MCP path even when `window.openai` is present — so any in-devtools call without `scrollToBottom` will hit the `default` branch and come back as a `-32601 Method not found` error. This makes `sendFollowUpMessage` effectively untestable in the devtools sandbox unless `scrollToBottom` is explicitly passed. Adding the missing case (or at least a no-op respond) would restore parity.

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["feat(devtools): add MCP ext-apps postMes..."](https://github.com/alpic-ai/skybridge/commit/3249994efa6148515561979dec982aecb52d2bf4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34180045)</sub>


harijoe:
Closing in favor of a new PR from the SKY-414-tied branch: julien/sky-414-refacto-the-bridge-pattern-to-a-single-ext-app-driven-class. Same commits, no content changes.

PR #832, opened 2026-05-28
refactor(core): wrap user entry to hide setViteManifest from src/server.ts

harijoe:
## Intent

Hide the `setViteManifest` boilerplate from `src/server.ts` while keeping the user entry convention unchanged: `export default await server.run()`. Users get a templated `McpServer` instance, register tools, and `await server.run()` — the framework wires the build manifest in transparently.

## How

- `skybridge build` emits `dist/__entry.js` that calls `__setBuildManifest(manifest)` (a new internal hook), then dynamically imports `dist/server.js`. The dynamic import is load-bearing — a static re-export would evaluate user code before the setter runs.
- `McpServer`'s constructor consumes the side channel: if `__setBuildManifest` was called, `setViteManifest` is invoked automatically. An explicit user call still wins because it happens after construction.
- `skybridge start` prefers `dist/__entry.js` when present (so views get hashed URLs in prod) and falls back to `dist/server.js` otherwise.
- Dev mode is unchanged: nodemon execs `tsx src/server.ts`, the user's top-level `await server.run()` binds the port. No dev wrapper needed.

## Breaking change

**No.** The user-facing convention (`export default await server.run()`) didn't change. Existing projects keep working as-is across every target — Node, Docker, Cloudflare Workers, and any future host whose entry shape is a fetch handler.

Templates and examples drop the `if (NODE_ENV === "production") { setViteManifest(...) }` snippet because the wrapper now handles that — but this is purely a simplification of newly-scaffolded projects, not a breaking change for existing ones (their explicit `setViteManifest` call still works).

---

<h3>Greptile Summary</h3>

This PR abstracts the Vite manifest setup away from user-facing `src/server.ts` files. A new `dist/__entry.js` wrapper is emitted during `skybridge build`; it calls the new internal `__setBuildManifest` hook before dynamically importing user code, ensuring `McpServer`'s constructor finds the manifest via a module-scoped side channel and clears it immediately (consume-once). `skybridge start` prefers the wrapper when present and falls back to `dist/server.js` for older builds.

- `packages/core/src/cli/build-helpers.ts` adds `emitEntryWrapper` / `ENTRY_WRAPPER_CONTENT`; `build.tsx` wires it in as a new build step after the manifest module is emitted, and `emitVercelBuildOutput` now bundles from `dist/__entry.js` rather than `dist/server.js`.
- `packages/core/src/server/server.ts` introduces `pendingBuildManifest` and `__setBuildManifest`; the constructor consumes and clears the side channel so stale state can't bleed across constructions.
- All examples, templates, and the deploy docs are updated to the new `export default await server.run()` convention, removing the `if (NODE_ENV === "production") { setViteManifest(...) }` boilerplate.

Safe to merge. The core mechanism — using a dynamic import to defer user-code evaluation until after __setBuildManifest runs — is sound across all target environments (Node, Vercel/esbuild bundle, Cloudflare/Wrangler bundle). The consume-once reset is correctly in place.

The side-channel pattern is well-understood, the consume-once reset in the constructor prevents cross-construction state leaks, build step ordering ensures the manifest module exists before the entry wrapper is written and before Vercel bundling runs, and the skybridge start fallback maintains backward compatibility for pre-existing builds. New tests cover both emitEntryWrapper and the consume-once semantics of __setBuildManifest.

1. `packages/core/src/cli/use-nodemon.ts`, line 23-33 ([link](https://github.com/alpic-ai/skybridge/blob/45494a17d346aeeb98354e429cbc554d38f7eaef/packages/core/src/cli/use-nodemon.ts#L23-L33)) 

   <a href="#">P2</a> **Custom `nodemon.json` bypasses the new dev wrapper**

   When a user has a `nodemon.json` in their project root, only `{configFile}` is passed to nodemon, so the `exec` is read from their file — typically `tsx src/server.ts`. If that user also migrates their `src/server.ts` to the new `export default server` pattern, `run()` is never called and the dev server starts but never listens. The PR's backward-compat guarantees only hold for the no-`nodemon.json` path. Adding a note in CLAUDE.md or in the migration guide warning users with a custom `nodemon.json` to update their exec to `tsx <path-to-dev-entry.js>` would prevent the silent failure.

   `````markdown
   Path: packages/core/src/cli/use-nodemon.ts
   Line: 23-33

   Comment:
   **Custom `nodemon.json` bypasses the new dev wrapper**

   When a user has a `nodemon.json` in their project root, only `{configFile}` is passed to nodemon, so the `exec` is read from their file — typically `tsx src/server.ts`. If that user also migrates their `src/server.ts` to the new `export default server` pattern, `run()` is never called and the dev server starts but never listens. The PR's backward-compat guarantees only hold for the no-`nodemon.json` path. Adding a note in CLAUDE.md or in the migration guide warning users with a custom `nodemon.json` to update their exec to `tsx <path-to-dev-entry.js>` would prevent the silent failure.

   `````
   </details>
</details>

<sub>Reviews (8): Last reviewed commit: ["fix(core): clear pendingBuildManifest af..."](https://github.com/alpic-ai/skybridge/commit/76bd199a66f5d24892f59d6999b1ddc8992317a8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34334588)</sub>


fredericbarthelet (commented):
Looks good overall. Thanks for putting together this not so easy one :)

I see the point of moving the following snippet out of user-land code
```ts
if (process.env.NODE_ENV === "production") {
  const { default: manifest } = await import("./vite-manifest.js");
  server.setViteManifest(manifest);
}
```

However, I don't see the point of reverting the skybridge entrypoint convention from `export default await server.run()` to `export default server`. This introduce the need to maintain backward compatibility with both declaration on build/start/dev through custom entrypoint file and I don't see the added benefit of not running `server.run()`. Am I missing something in this equation?

packages/core/src/cli/build-helpers.ts
```diff
@@ -7,6 +7,24 @@ import {
...
 
+// Primes the manifest in skybridge's module scope, then dynamically imports
+// `./server.js` so user code runs *after* the side channel is set. The
+// dynamic import is load-bearing: a static `export { default } from ...` is
+// hoisted with the rest of the static graph and would evaluate `server.js`
+// before `__setBuildManifest` runs.
+export const ENTRY_WRAPPER_CONTENT = `import { __setBuildManifest } from "skybridge/server";
+import manifest from "./vite-manifest.js";
+
+__setBuildManifest(manifest);
```
fredericbarthelet, line 18: why creating a new exported API rather than use the existing `setViteManifest` API on server?

packages/core/src/commands/start.ts
```diff
@@ -24,7 +24,13 @@ export default class Start extends Command {
...
     console.clear();
 
-    const indexPath = resolve(process.cwd(), "dist/server.js");
+    // `dist/server.js` is the natural entry (the user's `await server.run()`
+    // binds the port). Prefer `dist/__entry.js` when present because it
+    // primes the Vite manifest first — without it, hashed asset URLs in
+    // views won't resolve.
+    const entryPath = resolve(process.cwd(), "dist/__entry.js");
+    const fallbackPath = resolve(process.cwd(), "dist/server.js");
+    const indexPath = existsSync(entryPath) ? entryPath : fallbackPath;
```
fredericbarthelet, line 33: why do you still need to fallback on priming from `server.js`? Worth case scenario, the manifest is set twice

packages/core/src/server/server.ts
```diff
@@ -431,6 +431,26 @@ const McpServerBaseOmitted = McpServerBase as unknown as new (
...
+let pendingBuildManifest: Record<string, { file: string }> | null = null;
+
+/**
+ * Prime the build-time Vite manifest before user code constructs its
+ * `McpServer`. Called from the generated `dist/__entry.js`; not part of the
+ * user-facing API.
+ *
+ * @internal
+ */
+export function __setBuildManifest(
```
fredericbarthelet, line 448: why not calling this directly from `server.run`?

fredericbarthelet (approved):

PR #833, opened 2026-05-28
SKY-414: collapse host adaptors into single HostAdaptor with per-method routing

harijoe:
Linear: https://linear.app/alpic/issue/SKY-414

## Summary

Two changes that move Skybridge to a single, ext-apps-driven runtime.

**1. One `HostAdaptor`** replaces `McpAppAdaptor` + `AppsSdkAdaptor`. It composes both transports and routes per method: MCP ext-apps is the always-present baseline, and `window.openai` is an overlay for the ChatGPT-only bits (`setViewState`, `openModal`, file ops, `setOpenInAppUrl`, the `display`/`viewState` host-context keys). `openExternal` and `sendFollowUpMessage` use `window.openai` only when given an Apps-SDK option (`redirectUrl: false` / `scrollToBottom`). `getAdaptor()` becomes a memoized factory that probes `window.openai`, so `window.skybridge.hostType` is no longer load-bearing.

**2. One view resource — advertise and serve decoupled.** Each view emits a single ext-apps resource (`ui://views/ext-apps/<component>.html`, `text/html;profile=mcp-app`, nested `ui.*` meta) instead of two:
- **Advertise** via `ui.resourceUri` only. ChatGPT renders from it (verified), so we stop emitting `openai/outputTemplate` — which lets us retire the legacy resource down the line.
- **Serve** the legacy `ui://views/apps-sdk/<component>.html` URL as well (resolve-only — not advertised, not in `resources/list`), returning the same content. ChatGPT caches a view's URL at submission, so this keeps already-published apps working after a Skybridge bump even if they don't re-submit.

## Public API

No breaking change for documented consumers — `src/web/index.ts`, `src/web/hooks/index.ts`, and `src/server/index.ts` are unchanged vs main, and the exports map blocks deep imports.

- **Removed** (both `@internal`): `McpAppAdaptor`, `AppsSdkAdaptor`.
- **Added**: `NotSupportedError`, `closeModal()`.
- **Deprecated no-ops** (still exported): `ViewHostType`, `ViewConfig.hosts`.
- **Internal/observable-but-compatible**: tool `_meta` advertises `ui.resourceUri` (no `openai/outputTemplate`); host-context now reads from the MCP transport (identical in ChatGPT, which speaks both); `setOpenInAppUrl("")` rejects instead of throwing; DevTools resolves views via `ui.resourceUri`.

## Notes

- Routing rules live inline in `HostAdaptor` — the code is the spec.
- The legacy URL is served by aliasing its path onto the canonical ext-apps handler through the `?v=`-agnostic resolver (SKY-435), which echoes the requested URI back on the response.
- `viewState` localStorage LRU persistence is carried over unchanged (used only when no Apps SDK overlay is active).

## Test plan

- [x] 293 unit tests + build + lint (`pnpm test && pnpm build`); DevTools `tsc -b` + format
- [x] `resources/list` returns one ext-apps resource; tool `_meta` advertises `ui.resourceUri` only
- [x] Legacy `apps-sdk` URL read resolves to the ext-apps content, echoing the requested URI
- [x] ChatGPT renders via `ui.resourceUri` alone (no `openai/outputTemplate`)
- [x] ChatGPT honors `ui.csp.redirectDomains` and `ui.description` from the nested `ui.*` form
- [ ] Smoke test inside an MCP-app baseline host

## Follow-ups

- `persistToLocalStorage`: `protected` → `private`.
- Escape-hatch hooks (`useMcpAppContext` / `useAppsSdkContext`) expose a few fields the cross-host hooks don't (`toolCancelled`, `widgetState.privateContent`, …) — revisit whether to deprecate or grow the cross-host hooks.
- Next major: remove the deprecated `ViewHostType` / `ViewConfig.hosts` and the always-`"mcp-app"` `hostType` plumbing.
- Stop serving the legacy `apps-sdk` URL once no published app references it.

<h3>Greptile Summary</h3>

This PR consolidates `McpAppAdaptor` + `AppsSdkAdaptor` into a single `HostAdaptor` that composes both transports with inline per-method routing, and simplifies the view resource layer to emit one `text/html;profile=mcp-app` ext-apps resource instead of two, while keeping the legacy `ui://views/apps-sdk/…` URL resolvable for backward compatibility with already-published apps.

- **`HostAdaptor`** is the new composite adaptor: MCP ext-apps is the always-present baseline for `callTool`, `requestDisplayMode`, `requestClose`, `requestSize`, and `sendFollowUpMessage` (unless `scrollToBottom` is provided); `window.openai` is the overlay for ChatGPT-exclusive features (`setViewState`, file ops, `openModal`, `setOpenInAppUrl`, `openExternal`).
- **Server-side** (`server.ts`): the dual resource registration (`apps-sdk` + `ext-apps`) is replaced by a single `ui://views/ext-apps/…` resource; `serveLegacyAppsSdkUrl` aliases the old path in `viewUriByPath` and a `stripQuery` fix ensures the response echoes the requested (legacy) URI instead of the canonical one.
- **`ViewHostType` / `ViewConfig.hosts`** are now `@deprecated` no-ops, and `NotSupportedError` + `closeModal()` are added to the public surface.

The consolidation is safe to merge. The routing rules are clearly exercised by the 293-test suite, the legacy URL alias is covered by an explicit new integration test, and the three concerns from the previous review round (sync throw in `setOpenInAppUrl`, `window.skybridge` guard in `getAdaptor`, and the discarded unsubscribe in `subscribeToViewUUID`) are all correctly resolved in this revision.

No logic regressions were found in the changed paths. The only finding is a misleading test name in `adaptor.test.ts` whose body already tests the right runtime behavior, but documents an incorrect routing predicate that could mislead future contributors.

No files require special attention. `adaptor.test.ts` has a minor test-description gap (the `openExternal` case where openai is present but no special option is passed is untested), but the production code is correct.

1. General comment 

   <a href="#">P1</a> **Head does not advertise the ext-apps view resource MIME type in resources/list**

   - **Bug**
     - The validation scenario on head returned a single `resources/list` item for `ui://views/ext-apps/widget.html` with nested `ui.*` metadata, and canonical/legacy `resources/read` both returned `mimeType: "text/html;profile=mcp-app"`. But the advertised resource entry itself omitted `mimeType`, failing the requirement that head advertise exactly one ext-apps view resource with MIME type `text/html;profile=mcp-app`.
   - **Cause**
     - The view resource is registered with only `{ description: view.description }` as its resource metadata/options, while the `mimeType` from `ViewResourceConfig` is only attached to `resources/read` contents. As captured in head output, `resources/list` therefore contains `name`, `uri`, `description`, and `_meta`, but no `mimeType`.
   - **Fix**
     - Include the view resource MIME type in the resource registration metadata/options used by `resources/list`, e.g. pass `mimeType` alongside the description when calling `registerResource` for the view resource, while preserving the existing `resources/read` `mimeType`.

   <sub><a href="https://www.greptile.com/trex">T-Rex</a> Ran code and verified through T-Rex</sub>
</details>

<sub>Reviews (19): Last reviewed commit: ["refactor(web): make HostAdaptor a getIns..."](https://github.com/alpic-ai/skybridge/commit/1483bf99fe9be3e6bd784d2817b5fa4d8961c1ca) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34334655)</sub>


harijoe:
@greptileai

packages/core/src/web/bridges/get-adaptor.ts
```diff
@@ -1,16 +1,40 @@
...
 export const getAdaptor = (): Adaptor => {
-  return window.skybridge.hostType === "apps-sdk"
-    ? AppsSdkAdaptor.getInstance()
-    : McpAppAdaptor.getInstance();
+  if (cached) {
+    return cached;
+  }
+
+  const hostType =
+    typeof window !== "undefined" ? window.skybridge?.hostType : undefined;
```
qchuchu: Why did we remove the type of window.skybridge ? 

I thought we typed it bc we are sure in a skybridge app that is being set

packages/core/src/web/bridges/adaptor.ts
```diff
@@ -0,0 +1,403 @@
...
+}
+
+/**
+ * @internal
+ * Single composite implementation of {@link Adaptor}. Composes the MCP App
+ * bridge (always present) with an optional `window.openai` overlay. Per-method
+ * routing rules are encoded inline.
+ */
+export class HostAdaptor implements Adaptor {
+  private readonly mcp: McpAppBridge;
```
qchuchu, line 59: I'm wondering : is it planned, in the future, to remove the bridge layer then ? 

Also, we might want in the future to remove the 2 different ressources structure ?

packages/core/src/web/bridges/get-adaptor.ts
```diff
@@ -1,16 +1,40 @@
-import { AppsSdkAdaptor } from "./apps-sdk/adaptor.js";
-import { McpAppAdaptor } from "./mcp-app/adaptor.js";
+import { HostAdaptor } from "./adaptor.js";
 import type { Adaptor } from "./types.js";
 
+let cached: HostAdaptor | null = null;
```
qchuchu: We used to have a pattern like this before, before moving to singleton pattern. Any reason why ?

packages/core/src/web/bridges/get-adaptor.ts
```diff
@@ -1,16 +1,40 @@
...
+    return cached;
+  }
+
+  const hostType =
+    typeof window !== "undefined" ? window.skybridge?.hostType : undefined;
+  const hasOai = typeof window !== "undefined" && window.openai !== undefined;
+
+  if (hostType === "mcp-app" && hasOai) {
+    console.warn(
+      "[skybridge] hostType is 'mcp-app' but window.openai is present; trusting the probe.",
```
qchuchu: Do you know if by default window.openai uses the ressource openai/outputTemplate over ui/resourceUri ? 

Bc I'm wondering if this is going to be displayed as a warning all the time now

fredericbarthelet: Image

If you look at the app inspector in ChatGPT for the starter app of Skybridge, you see apps-sdk template being sourced first for rendering of tool result. This warning will be displayed all the time

packages/core/src/web/bridges/adaptor.ts
```diff
@@ -0,0 +1,403 @@
...
+
+    if (typeof window !== "undefined" && window.openai !== undefined) {
+      this.openai = window.openai;
+      this.appsSdkBridge = AppsSdkBridge.getInstance();
+      this.appsSdkStores = buildAppsSdkOverlayStores(this.appsSdkBridge);
+    } else {
+      this.openai = null;
+      this.appsSdkBridge = null;
+      this.appsSdkStores = null;
+    }
```
fredericbarthelet: unecessary, just initialize them to null by default

docs/api-reference/register-tool.mdx
```diff
@@ -62,7 +62,7 @@ Binds the tool to a React view.
 type ViewConfig = {
   component: ViewName;                   // File name in src/views/ (typed — see below)
   description?: string;                    // Shown during view discovery
-  hosts?: ("apps-sdk" | "mcp-app")[];      // Defaults to both
+  hosts?: ("apps-sdk" | "mcp-app")[];      // Deprecated, no-op — one resource is always emitted
```
fredericbarthelet:
```suggestion
  hosts?: ("apps-sdk" | "mcp-app")[];      // Deprecated, no-op
```

packages/core/src/server/server.ts
```diff
@@ -217,21 +224,6 @@ type InternalToolMeta = Partial<
...
-/** @see https://developers.openai.com/apps-sdk/reference#component-resource-_meta-fields */
-type OpenaiViewCSP = {
-  connect_domains: string[];
-  resource_domains: string[];
-  frame_domains?: string[];
-  redirect_domains?: string[];
-};
-
-type OpenaiResourceMeta = {
-  "openai/widgetDescription"?: string;
```
fredericbarthelet, line 229: `openai/widgetDescription` is not advertise as a legacy compatiblity feature, people can still use it for ChatGPT only added information (see https://developers.openai.com/apps-sdk/reference#component-resource-_meta-fields)

packages/core/src/server/server.ts
```diff
@@ -217,21 +224,6 @@ type InternalToolMeta = Partial<
   OpenaiToolMeta & McpAppsToolMeta & SecuritySchemesToolMeta
 >;
 
-/** @see https://developers.openai.com/apps-sdk/reference#component-resource-_meta-fields */
-type OpenaiViewCSP = {
-  connect_domains: string[];
-  resource_domains: string[];
-  frame_domains?: string[];
-  redirect_domains?: string[];
```
fredericbarthelet, line 225: `redirect_domains` is ChatGPT only feature. It's currently being discussed in ext-apps to bring this to the protocol, but until it's included, this CSP should still be present in order to configure links without confirmation modal - see https://github.com/modelcontextprotocol/ext-apps/issues/678

packages/core/src/server/server.ts
```diff
@@ -726,7 +716,7 @@ export class McpServer<
           for (const content of result.contents ?? []) {
             if (
               typeof content.uri === "string" &&
-              stripQuery(content.uri) === path
+              stripQuery(content.uri) === stripQuery(canonical)
```
fredericbarthelet, line 709: I don't understand why this needs to be changed, can you explain?

packages/core/src/server/server.ts
```diff
@@ -958,134 +948,79 @@ export class McpServer<
     view: ViewConfig,
     toolMeta: InternalToolMeta,
   ): void {
-    const hosts = view.hosts ?? (["apps-sdk", "mcp-app"] as const);
+    // `hosts` is deprecated and ignored — every view emits the single ext-apps
+    // resource. Kept for backwards compatibility; see ViewConfig.hosts.
```
fredericbarthelet: this comment has no meaning when read here on its own, can be removed

packages/core/src/server/server.ts
```diff
@@ -958,134 +948,79 @@ export class McpServer<
...
+        }
+        return base;
+      },
+    };
+    this.registerViewResource({ name: toolName, viewResource, view });
+
+    // Advertise via the MCP Apps standard pointer only — ChatGPT renders from
+    // ui.resourceUri (verified), and not emitting openai/outputTemplate lets us
+    // retire the legacy apps-sdk resource later. The legacy apps-sdk URL is still
+    // served (see registerViewResource) so already-published apps keep resolving.
```
fredericbarthelet, line 1014: this comment is relevant, but you miss the actual call to `registerViewResource` no?

fredericbarthelet: ok, now I've seen the addition of the new entry with `viewUriByPath`. Now that I see it, I find it less clear. I'd prefer another call to `this.registerViewResource` to make it clear 2 resources are registered (rather than wiring through a set and a middleware the same content for both resources). How will that impact resource/list operation by the way?

packages/core/src/web/bridges/get-adaptor.ts
fredericbarthelet: indeed, prefer the clean static implementation on HostAdaptor class with a `getInstance` method to avoid side effect

packages/core/src/web/bridges/get-adaptor.ts
```diff
@@ -1,16 +1,25 @@
...
+  }
+  cached = new HostAdaptor();
+  return cached;
+};
+
+/** @internal Test-only. */
+export const _resetAdaptor = (): void => {
+  cached?.cleanup();
+  cached = null;
 };
```
fredericbarthelet: this can completely be removed and use a `getInstance` cleaner method on `HostAdaptor` directly

packages/core/src/web/bridges/adaptor.ts
```diff
@@ -0,0 +1,408 @@
...
+      },
+      getSnapshot: () => this._viewState,
+    };
+
+    this.subscribeToViewUUID();
+  }
+
+  public hasAppsSdkOverlay(): boolean {
+    return this.openai !== null;
+  }
```
fredericbarthelet: not a big fan of the name, or just introducing this API at all, why is it useful for downstream consumer?

packages/core/src/web/bridges/adaptor.ts
```diff
@@ -0,0 +1,408 @@
...
+
+  /** @internal Release any subscriptions held on the underlying bridges. */
+  public cleanup(): void {
+    this.unsubscribeViewUUID?.();
+    this.unsubscribeViewUUID = null;
+  }
+
+  // ---- Adaptor interface ----
+
+  public getHostContextStore = <K extends keyof HostContext>(
```
fredericbarthelet, line 124: maybe as a best backward compatibility tentative we can still wire `useAppsSdkContext` and `useMcpAppContext` to there respective. `this.appsSdkBridge.use` and `this.mcp.use` no?

packages/core/src/web/bridges/adaptor.ts
```diff
@@ -0,0 +1,408 @@
...
+    }
+    const app = await this.mcp.getApp();
+    await app.sendMessage({
+      role: "user",
+      content: [{ type: "text", text: prompt }],
+    });
+  };
+
+  public openExternal = (href: string, options?: OpenExternalOptions): void => {
+    if (this.openai && options?.redirectUrl === false) {
```
fredericbarthelet: why not prefer openai method (with withelisted URLs from CSP) even if redirectUrl is true?

packages/core/src/server/server.ts
harijoe: You're right, I missed that

packages/core/src/web/bridges/adaptor.ts
fredericbarthelet: or did you meant to make this API internal? There are a few places in the actual implementation where you still check `this.openai` rather than using this method.

fredericbarthelet (commented):
I see a lot of variations in naming across this PR `AppsSdk`, `openai` for Apps SDK or even `mcp`, `ext-apps`, `mcp-app` for MCP App. Can we align on a single naming (except where it can't be changed without breaking like in the resource URI) :)?

packages/core/src/web/bridges/adaptor.ts
```diff
@@ -0,0 +1,408 @@
...
+    } catch (error) {
+      console.error("Failed to update view state in MCP App.", error);
+    }
+  };
+
+  public uploadFile = async (
+    file: File,
+    options?: UploadFileOptions,
+  ): Promise<FileMetadata> => {
+    if (!this.openai) {
```
fredericbarthelet, line 256:
```suggestion
    if (!this.hasAppsSdkOverlay()) {
```

packages/core/src/web/bridges/adaptor.ts
```diff
@@ -0,0 +1,408 @@
...
+      );
+      return { isError: true };
+    }
+    return app.downloadFile(params);
+  };
+
+  public setViewState = async (
+    stateOrUpdater: SetViewStateAction,
+  ): Promise<void> => {
+    if (this.openai) {
```
fredericbarthelet, line 215:
```suggestion
    if (this.hasAppsSdkOverlay()) {
```

packages/core/src/web/bridges/host-context-stores.ts
```diff
@@ -0,0 +1,136 @@
...
+type PickContext<K extends readonly McpAppContextKey[]> = {
+  [P in K[number]]: McpAppContext[P];
+};
+
+/**
+ * Build a memoized {@link HostContextStore} backed by one or more
+ * `McpAppBridge` context keys. Recomputed values are deep-compared via
+ * `dequal` so unchanged snapshots return the cached reference.
+ */
+function createMcpStore<const Keys extends readonly McpAppContextKey[], R>(
```
fredericbarthelet: you can still ship those methods in dedicated `createStore` method on apps-sdk bridge and mcp-app bridge. It would still confine each store implementation in it's respective folder structure (where the adaptor used to live) and avoid this new file that I took a bit of time understanding. The maid adaptor shouldn't store 2 stores on its state, but a single one, polymorphically built in the constructor

packages/core/src/web/bridges/types.ts
```diff
@@ -247,8 +247,30 @@ export interface Adaptor {
   selectFiles(): Promise<FileMetadata[]>;
   openModal(options: RequestModalOptions): void;
   setOpenInAppUrl(href: string): Promise<void>;
+  /**
+   * Close a modal opened via {@link Adaptor.openModal}. In the Apps SDK
+   * runtime this is a no-op (the host owns modal lifecycle). In the MCP App
+   * runtime it dismisses the in-iframe polyfill.
+   */
```
fredericbarthelet: this is implementation detail writted in adaptor.ts, the comment is unecessary here

packages/devtools/src/components/layout/tool-panel/index.tsx
```diff
@@ -88,9 +88,9 @@ export const ToolPanel = () => {
       setPreference("displayMode", "inline");
     }
   });
-  const templateUri = tool?._meta?.["openai/outputTemplate"] as
-    | string
-    | undefined;
+  const templateUri =
+    (tool?._meta?.ui as { resourceUri?: string } | undefined)?.resourceUri ??
+    (tool?._meta?.["openai/outputTemplate"] as string | undefined);
```
fredericbarthelet: this is unecessary, you'll never have a skybridge devtools version reading an older skybridge core version, and even if you do, this fallback is not required

fredericbarthelet (approved):
I made a bunch of comments, feel free to implement/discard before merging. Awesome work!

packages/core/src/server/server.ts
harijoe: That's the problem, if we call registerViewResource twice, we will advertise 2 resources and that's not what we want. It's not possible using the MCP sdk to register a read handler without listing it.
I still agree that the new entry in viewUriByPath could be a bit clearer

packages/core/src/server/server.ts
harijoe: In one word : because of the alias we introduced in registerResource

> === path only matched when the requested path equalled the canonical's path (true for ?v= variants of the same URL). The apps-sdk→ext-apps alias is cross-path, so it didn't match and the response leaked the canonical ext-apps URL. === stripQuery(canonical) compares against what the handler actually returns, so the consumer always gets its own requested URL back, identical for the ?v= case, fixed for the alias.

packages/core/src/web/bridges/adaptor.ts
harijoe: I removed hasAppsSdkOverlay, it doesn't bring much value over this.openai

packages/core/src/web/bridges/adaptor.ts
harijoe: I remove it, checking on this.openai is good enough

packages/core/src/web/bridges/adaptor.ts
```diff
@@ -0,0 +1,408 @@
...
+      role: "user",
+      content: [{ type: "text", text: prompt }],
+    });
+  };
+
+  public openExternal = (href: string, options?: OpenExternalOptions): void => {
+    if (this.openai) {
+      this.openai.openExternal({ href, ...options });
+      return;
+    }
```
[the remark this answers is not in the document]
harijoe: Intentional — this restores pre-refactor behavior. The old `AppsSdkAdaptor` always routed `openExternal` through `window.openai.openExternal({ href, ...options })` in ChatGPT; the "openai only for `redirectUrl: false`" routing was a regression introduced by the adaptor consolidation, not the intended contract. A maintainer requested it, and it's required for `openai/widgetCSP.redirect_domains` to take effect — that allowlist is only honored on the `window.openai.openExternal` path, so routing the common case to MCP `openLink` would make it inert. Keeping it.

harijoe:
> I see a lot of variations in naming across this PR `AppsSdk`, `openai` for Apps SDK or even `mcp`, `ext-apps`, `mcp-app` for MCP App. Can we align on a single naming (except where it can't be changed without breaking like in the resource URI) :)?

The naming inconsistencies that wouldn't break public api aren't obvious to me. I'd keep digging this outside of this PR which is already massive.

PR #842, opened 2026-06-01
examples/chess

valentinbeggi:
- **feat(core): add view-provided tools (useRegisterViewTool)**
- **feat(examples): add chess app showcasing view-provided tools**

<h3>Greptile Summary</h3>

This PR introduces `useRegisterViewTool`, a React hook that lets a view register tools the host and model can call directly against the running widget (the MCP Apps "app-provided tools" feature), and adds a chess example built entirely around it.

- **Core hook (`use-register-view-tool.ts`)**: Clean ref-based design keeps the handler always in sync with current state while re-registering only when the tool `name` changes. Both adaptors implement `registerViewTool`; the Apps SDK no-ops with a warning as documented.
- **Bridge upgrade**: `ext-apps` is bumped to v1.7.3 to gain `app.registerTool`. Test coverage spans capability advertisement, tool listing, arg validation, handler invocation, and unregister/notify.
- **Chess example**: The `chess_make_move` handler doesn't guard against it being White's turn — chess.js validates legality for the *current mover*, so a valid White move passed on White's turn succeeds, letting the model advance the board without any user drag.

Safe to merge after adding a Black-turn guard in the chess_make_move handler; the core hook and bridge are solid.

The `useRegisterViewTool` hook and bridge implementation are well-tested and architecturally sound. The one real defect is in the chess example: `chess_make_move` never checks that it is actually Black's turn before forwarding the request to chess.js, which validates only legality for the current mover. A White move supplied on White's turn succeeds, letting the model advance the board on the player's behalf without any user drag gesture.

examples/chess/src/views/chess.tsx — the chess_make_move handler needs a turn guard

`````markdown

---

### Issue 1 of 1
examples/chess/src/views/chess.tsx:129-147
The `chess_make_move` handler says "it must be Black's turn" in its description but never enforces it. Chess.js validates legality against the *current side to move*, not against a fixed color — so if the model (or host) calls the tool when it's White's turn and passes a valid White move (`{ san: "e4" }` at the start position, for example), the move succeeds and the assistant effectively plays White's pieces. The game state advances on White's behalf without any user drag, leaving the player with no turn to take.

Add an explicit turn guard before delegating to `play`.

```suggestion
    ({ san, from, to, promotion }) => {
      const { fen } = useMatch.getState();
      const position = readPosition(fen);
      if (position.toMove !== "b") {
        return {
          content: [
            {
              type: "text",
              text: "It is not Black's turn — wait for White (the player) to move first.",
            },
          ],
          isError: true,
        };
      }

      const request = san
        ? { san }
        : from && to
          ? { from, to, promotion }
          : null;
      if (!request) {
        return {
          content: [
            {
              type: "text",
              text: "Pass either `san`, or both `from` and `to`.",
            },
          ],
          isError: true,
        };
      }

      const result = useMatch.getState().play(request);
```

`````

</details>

<sub>Reviews (2): Last reviewed commit: ["feat(examples): add chess app showcasing..."](https://github.com/alpic-ai/skybridge/commit/3151cd9014cd3b6d22743b98722585c464a3b861) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34966338)</sub>

### DIFF
--- a/docs/api-reference/use-register-view-tool.mdx
+++ b/docs/api-reference/use-register-view-tool.mdx
@@ -1,0 +1,147 @@
+---
+title: useRegisterViewTool
+description: "React hook for exposing app-provided (view) tools that the host and model can call directly against the running view."
+---
+
+`useRegisterViewTool` registers a **view-provided tool** — a tool whose handler
+runs _inside the view_ rather than on the MCP server. The host discovers it via
+`tools/list` and invokes it via `tools/call`, and your handler executes against
+the view's live state. This is the MCP Apps ["app-provided
+tools"](https://github.com/modelcontextprotocol/ext-apps/pull/72) feature.
+
+It is the mirror image of [`useCallTool`](/api-reference/use-call-tool): with
+`useCallTool` the view _calls_ a server tool; with `useRegisterViewTool` the
+view _exposes_ a tool for the model to call.
+
+<Note>
+  View tools are an **MCP Apps** feature. The ChatGPT Apps SDK
+  (`window.openai`) runtime has no equivalent, so on that runtime
+  `useRegisterViewTool` is a no-op (a warning is logged). Use it when targeting
+  MCP Apps hosts.
+</Note>
+
+## Basic usage
+
+```tsx
+import { useRegisterViewTool } from "skybridge/web";
+import * as z from "zod";
+
+function Counter() {
+  const [count, setCount] = useState(0);
+
+  useRegisterViewTool(
+    {
+      name: "counter_increment",
+      description: "Increment the on-screen counter by an amount.",
+      inputSchema: { by: z.number().int().default(1) },
+      annotations: { readOnlyHint: false },
+    },
+    ({ by }) => {
+      setCount((current) => current + by);
+      return {
+        content: [{ type: "text", text: `Counter is now ${count + by}.` }],
+        structuredContent: { count: count + by },
+      };
+    },
+  );
+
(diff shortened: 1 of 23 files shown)

### CODE REVIEW

File: docs/api-reference/use-register-view-tool.mdx [hunk 1]
```diff
@@ -0,0 +1,147 @@
...
+  structuredContent?: Record<string, unknown>;
+  isError?: boolean;
+  _meta?: Record<string, unknown>;
+};
+```
+
+**Required.** Runs when the host calls the tool. `args` is typed from
+`config.inputSchema`. Return a standard MCP `CallToolResult` — `content`
+blocks plus optional `structuredContent`. Set `isError: true` to report a
+failure to the model.
```
thleonard, line 103: No _meta?

examples/chess/src/views/chess.tsx
```diff
@@ -0,0 +1,248 @@
...
+      };
+    },
+  );
+
+  useRegisterViewTool(
+    {
+      name: TOOL.legalMoves,
+      description:
+        "List the legal moves for the side to move. Optionally restrict to moves originating from a single square (e.g. `from: 'e7'`).",
+      inputSchema: {
```
thleonard, line 75: Isn't it way too big for the context?

valentinbeggi: no it's fine, list of ~20 moves max

thleonard (approved):

PR #848, opened 2026-06-04
Added Descope MCP Server auth

mrunankpawar:
Adds `examples/auth-descope`, a complete MCP server example that uses Descope as the auth provider, along with accompanying documentation.

Developers can configure an MCP Server in the Agentic Identity Hub and point it to their MCP server URL. The example implements the complete OAuth 2.1 flow, including server discovery, Dynamic Client Registration (DCR), token exchange, and JWT verification using Descope's JWKS endpoint.


examples/auth-descope/src/server.ts
```diff
@@ -0,0 +1,164 @@
...
+  McpServer,
+  mcpAuthMetadataRouter,
+  requireBearerAuth,
+} from "skybridge/server";
+import * as z from "zod";
+import { verifyAccessToken } from "./auth.js";
+import { searchCoffeeShops } from "./coffee-data.js";
+import { env } from "./env.js";
+
+const DESCOPE_MCP_BASE = `https://api.descope.com/oauth2/v1/apps/agentic/${env.DESCOPE_PROJECT_ID}/${env.DESCOPE_MCP_SERVER_ID}`;
```
gaokevin1: This will change to being a project level issuer URL, with an audience claim.

examples/auth-descope/src/server.ts
```diff
@@ -0,0 +1,164 @@
...
+)
+  .use(cors())
+  // Proxy DCR to Descope without the x-forwarded-host header the playground injects
+  .use("/register", registrationProxy)
+  // Serve AS metadata from this server so playground doesn't hit Descope directly
+  .use(
+    mcpAuthMetadataRouter({
+      oauthMetadata: {
+        issuer: env.SERVER_URL,
+        authorization_endpoint: `${DESCOPE_MCP_BASE}/authorize`,
```
gaokevin1: Add `audience` parameter to this after next week.

examples/auth-descope/src/server.ts
```diff
@@ -0,0 +1,164 @@
...
+        jwks_uri: `https://api.descope.com/${env.DESCOPE_PROJECT_ID}/.well-known/jwks.json`,
+        response_types_supported: ["code"],
+        code_challenge_methods_supported: ["S256"],
+        token_endpoint_auth_methods_supported: [
+          "none",
+          "client_secret_basic",
+          "client_secret_post",
+        ],
+        grant_types_supported: ["authorization_code", "refresh_token"],
+        scopes_supported: ["openid", "email"],
```
gaokevin1, line 80: This I don't think should be statically defined. This will change depending on the MCP server configuration in Descope right?

@csonigo how would we fetch this from Descope dynamically, if you're serving the AS metadata on our behalf? Is there a way to do this within Skybridge or should the developer just manage this themselves?

examples/auth-descope/src/server.ts
```diff
@@ -0,0 +1,164 @@
...
+      },
+      resourceServerUrl: new URL("/mcp", env.SERVER_URL),
+    }),
+  )
+  .use(
+    "/mcp",
+    requireBearerAuth({
+      verifier: { verifyAccessToken },
+      requiredScopes: ["openid"],
```
gaokevin1, line 89: This is just the basic required scope for the server correct?

We should add a comment making that clear.

examples/auth-descope/src/server.ts
```diff
@@ -0,0 +1,164 @@
...
+        },
+      },
+      _meta: {
+        "openai/widgetAccessible": true,
+      },
+    },
+    ({ query, minRating }, extra) => {
+      const auth = extra.authInfo as AuthInfo;
+
+      const name = auth.extra?.name as string | undefined;
```
gaokevin1, line 132: I would include in your example a way to show how to extract custom claims from the JWT as well and how to use them in the rest of the server.

This is something I assume you can pass through to your tools via `auth.extra` correct?

examples/auth-descope/src/auth.ts
```diff
@@ -0,0 +1,42 @@
...
+    return {
+      token,
+      clientId: payload.sub as string,
+      scopes:
+        typeof payload.scope === "string"
+          ? payload.scope.split(" ").filter(Boolean)
+          : [],
+      expiresAt: payload.exp,
+      extra: {
+        sub: payload.sub,
```
gaokevin1, line 32: We are not fetching these from `/userinfo`, they are actually being taken from the JWT claims. 

@csonigo if you want to do some kind of token introspection I assume you would hit the `/userinfo` endpoint within this function and set these values to whatever is from that, regardless of the token contents correct?

@mrunankpawar we should make sure to comment here about token introspection and link to our docs.

examples/auth-descope/README.md
```diff
@@ -0,0 +1,86 @@
+# Auth Example — Descope
+
+An example MCP app built with [Skybridge](https://docs.skybridge.tech/home): a personalized coffee shop finder demonstrating full OAuth authentication with [Descope](https://descope.com/).
+
+## What This Example Showcases
```
gaokevin1, line 5: Image

We're missing the Deploy to Alpic button and also the additional resources as well I think. @mrunankpawar

examples/auth-descope/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+- **Transport-Level Auth**: Auth is enforced at the `/mcp` transport level — unauthenticated requests receive HTTP 401 before reaching any tool handler
+- **Descope OAuth**: Full OAuth2 setup with JWT verification via Descope's JWKS endpoint
```
gaokevin1, line 8: Do any examples @mrunankpawar @csonigo show tool level scoping, or just basic JWT validation at the `/mcp` entry point to the server?

thleonard: Hi @gaokevin1, we have no example app with mixed auth. I'm not sure it's natively supported by Skybridge but I've never experimented with it. Maybe @harijoe could tell you more about it

examples/auth-descope/src/auth.ts
thleonard: Hi @gaokevin1, I'd say that the token is the source of truth here. I believe we can trust the claims once it's been verified above. Do you have push back on this ?

thleonard:
Hi @mrunankpawar @gaokevin1, thanks for the PR! Looking forward to merge it!
The PR is still in draft, is it ready for review on our end or do you need more time? Either way you'll need to rebase on main

fredericbarthelet:
Hey @mrunankpawar @gaokevin1 , happy to help get this merge soon.
@harijoe is working on upcoming minor Skybridge release with the addition of OAuth helpers to empower user with drop-in solution to protect their MCP App.
We'd love to also have the exemple app as part of the release. Let us know how we can help :)

examples/auth-descope/src/auth.ts
gaokevin1: > Hi @gaokevin1, I'd say that the token is the source of truth here. I believe we can trust the claims once it's been verified above. Do you have push back on this ?

Technically the dynamic claims can change, for example if the session is valid for 10 mins and the user is deactivated or something within that timeframe. Not the most common use case by any means, but it is something that we hear from customers.

It's ok if we don't support something like this for now, but it's something to consider for the future.

gaokevin1:
> Hi @mrunankpawar @gaokevin1, thanks for the PR! Looking forward to merge it! The PR is still in draft, is it ready for review on our end or do you need more time? Either way you'll need to rebase on main

Our changes finally were deployed and we're doing a final review today. After that we can take it out of draft for you guys to review. Thanks!

examples/auth-descope/src/server.ts
gaokevin1:
```suggestion
const DESCOPE_MCP_BASE = `${env.DESCOPE_MCP_SERVER_ISSUER_URL}`;
```

examples/auth-descope/src/server.ts
gaokevin1: Audience should be the MCP server URL (aka the audience of the resource)

harijoe:
Hi @mrunankpawar, hi @gaokevin1, thanks a lot for this, and sorry for the slow turnaround. Your Descope example was genuinely useful and it directly informed where we landed.

We've folded Descope into a broader PR, #879. For consistency we're going to favor that implementation, so we'd like to close this PR in favor of #879, but your contribution is carried over and will still be credited there.

A few notes on how it compares, in the spirit of full transparency:
  - The demo app from your PR is carried over almost verbatim (the coffee-data, view, and helpers are byte-identical).
  - Your manual src/auth.ts (hardcoded JWKS + aud = ${SERVER_URL}/mcp) is replaced by the discovery-based descopeProvider.
  - One correction we made while testing live against Descope: the access token's aud is actually [DCR client id, project id] — not the MCP server URL — so the audience is the Project ID. We verified this end-to-end through the Alpic playground.
  - The one thing your PR has that #879 is currently missing is the docs-site integration. We'd love to bring those over in the follow-up PR that is already in the pipe.

  Would you be up for reviewing #879 (especially the Descope bits). Really appreciate the contribution either way. 🙏

PR #850, opened 2026-06-08
feat(devtools): add saved queries

paulleseute:
This adds saved queries so user can stash a tool's input under a name and pull it back in one click. 

How it works:
- The input header now has a selector where the "Input" label used to be. Pick a saved query to load it into both the form and JSON views; pick "none" to clear and start fresh.
- "save" stashes the current input under a name. Existing name → overwrites it, new name → creates one. The modal prefills with the current selection (so re-saving over it is one click), and Enter submits.
- Trash icon on hover deletes a query.
- Queries live in localStorage, scoped per server + tool, so they survive reloads and don't bleed across projects.
- If a tool's schema changed since you saved, loading keeps only the fields that still exist.

The input header got a small refresh while I was in there: form/json is now a single toggle, copy floats in the top-right of the JSON box, and the tool name is a touch bigger/bolder.

Implementation: a small zustand store persisted to localStorage (serverName → toolName → queries). The active selection is tracked per-tool on the existing in-memory tool store, and since the form and JSON views already share one input field, both stay in sync for free. New saved-queries.tsx holds the dropdown + save dialog; tool-item.tsx wires it into the header.

Coming next: run button + saved queries always in the tool header, regardless of the accordion state.

https://github.com/user-attachments/assets/083c22cc-9016-4258-8019-049e1f9f2799

<h3>Greptile Summary</h3>

This PR adds a saved-queries feature to the devtools panel: users can stash a tool's current input under a name, reload it in one click, and delete stale entries. Queries are scoped per server+tool and persisted to localStorage via a new zustand store.

- A new `SavedQueriesDropdown` and `SaveQueryDialog` are added in `saved-queries.tsx`; `tool-item.tsx` wires them into the existing `ToolBody` header, also fixing `meta+enter`/`enter` keyboard shortcuts to ignore keypresses inside any dialog.
- The input header is refreshed: the old tab list becomes a single form/JSON toggle button, and the copy button moves into the JSON textarea overlay.
- `intersectInputToSchema` defensively trims saved inputs to fields that still exist in the current schema, avoiding broken state after tool-schema changes.

The change is self-contained and additive — no existing behaviour is removed, only new optional UI is layered on top.

The store actions are immutable and guard against missing entries cleanly. The active-key derivation handles stale references without leaking into the UI. The `inDialog` keyboard guard correctly prevents the tool-run shortcut from firing inside the save dialog. No data-loss or incorrect-state paths were found.

<sub>Reviews (3): Last reviewed commit: ["fix: input instead of query"](https://github.com/alpic-ai/skybridge/commit/e871d34d3b0e4385af92496ae9e155ed204fbca7) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36220931)</sub>

### DIFF
--- a/packages/devtools/src/components/layout/tools-list/saved-queries.tsx
+++ b/packages/devtools/src/components/layout/tools-list/saved-queries.tsx
@@ -1,0 +1,190 @@
+import { Button } from "@alpic-ai/ui/components/button";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from "@alpic-ai/ui/components/dialog";
+import {
+  DropdownMenu,
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuTrigger,
+} from "@alpic-ai/ui/components/dropdown-menu";
+import { Input } from "@alpic-ai/ui/components/input";
+import { Label } from "@alpic-ai/ui/components/label";
+import { Check, ChevronDown, Trash2 } from "lucide-react";
+import { type FormEvent, useEffect, useId, useState } from "react";
+import type { SavedQuery } from "@/lib/saved-queries-store.js";
+import { cn } from "@/lib/utils.js";
+
+export function SavedQueriesDropdown({
+  queries,
+  activeKey,
+  onSelect,
+  onClear,
+  onDelete,
+}: {
+  queries: SavedQuery[];
+  activeKey: string | null;
+  onSelect: (query: SavedQuery) => void;
+  onClear: () => void;
+  onDelete: (key: string) => void;
+}) {
+  return (
+    <DropdownMenu>
+      <DropdownMenuTrigger asChild>
+        <button
+          type="button"
+          className="inline-flex max-w-52 cursor-pointer items-center gap-1 rounded px-1 py-0.5 font-mono text-xs text-muted-foreground transition-colors hover:text-foreground focus-visible:outline-none focus-visible:ring-1 focus-visible:ring-ring aria-expanded:text-foreground"
+        >
+          <span className="truncate">{activeKey ?? "Input"}</span>
+          <ChevronDown className="size-3 shrink-0" />
+        </button>
+      </DropdownMenuTrigger>
+      <DropdownMenuContent align="start" className="min-w-48">
+        {queries.length === 0 ? (
+          <DropdownMenuItem
+            disabled
+            className="font-mono text-xs font-normal text-muted-foreground"
+          >
+            No saved queries yet
+          </DropdownMenuItem>
+        ) : (
+          <>
+            {/* "none" unselects the active query and clears the input. */}
+            <DropdownMenuItem
+              onSelect={onClear}
+              className="cursor-pointer font-mono text-xs font-normal"
+              disabled={queries.length < 0}
+            >
+              {queries.length > 0 && (
+                <Check
+                  className={cn(
+                    "size-3.5 shrink-0",
+                    activeKey === null ? "opacity-100" : "opacity-0",
+                  )}
+                />
+              )}
+
+              <span className="text-muted-foreground italic">
+                {queries.length > 0 ? "none" : "no saved input yet"}
+              </span>
+            </DropdownMenuItem>
+            {queries.map((query) => (
+              // The trash sits as a sibling overlay (not a child of the item)
+              // so its click never enters the item's selection path: clicking
+              // it deletes without also loading the query.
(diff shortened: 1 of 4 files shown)


packages/devtools/src/lib/saved-queries-store.ts
```diff
@@ -0,0 +1,104 @@
...
+          const forServer = state.queries[serverName] ?? {};
+          const forTool = forServer[toolName] ?? [];
+          const query: SavedQuery = { key, input, createdAt: Date.now() };
+          return {
+            queries: {
+              ...state.queries,
+              [serverName]: { ...forServer, [toolName]: [...forTool, query] },
+            },
+          };
+        }),
```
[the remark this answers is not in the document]
paulleseute: useless today

valentinbeggi:
wdyt of 

- colocating the saved query with the save cta + add update state 
image

- just show save button if none exist (1 click less to get to save your query) 
image

paulleseute:
> wdyt of
> 
> * colocating the saved query with the save cta + add update state
> 
> image
> * just show save button if none exist (1 click less to get to save your query)
> 
> image

@valentinbeggi I strongly considered having an update button but I went for a single save because:
- We want to cover two patterns: update existing and save a new one. You can save a new query from an existing one.
- I think having all the controls on the same row (and not have this specific layout dynamically change) brings structure and clarity ("here is everything you can do with inputs")
- Having json|form + update + save buttons on the same row make it crowded (especially on 13inch), which deteriorates the UI (ellipsis & co)
- The "one click" update is fair concern but I think the saved queries ops are rare enough to have two clicks for updates.

valentinbeggi:
@paulleseute what I find a bit off is the dropdown, existing even when no saved query exist, that isn't clearly linked to the save action.
maybe we could use a third eye on this 👁️ (cc @harijoe @qchuchu )

paulleseute:
> @paulleseute what I find a bit off is the dropdown, existing even when no saved query exist, that isn't clearly linked to the save action. maybe we could use a third eye on this 👁️ (cc @harijoe @qchuchu )

@valentinbeggi it's actually by design, clicking on it nudges the user into creating the first one. but i see your concern. maybe having the save button next to the saved query label is clearer. my only concern is that its position will be not be fixed and related to the saved query name length, which is not great imo - i'd expect a cat to always have the same position

packages/devtools/src/lib/store.ts
```diff
@@ -23,6 +23,8 @@ type ToolData = {
   openaiLogs: OpenAiLog[];
   openaiObject: AppsSdkContext | null;
   openInAppUrl: string | null;
+  // Key of the saved query the current input was loaded from (session-only).
```
qchuchu, line 26: comment not needed

packages/devtools/src/lib/saved-queries-store.ts
```diff
@@ -0,0 +1,104 @@
...
+    }),
+    {
+      name: "skybridge-devtools-saved-queries",
+      version: 1,
+      partialize: (state) => ({ queries: state.queries }),
+    },
+  ),
+);
+
+/** A tool's saved queries (stable empty array when none). */
```
qchuchu, line 97: Same here

qchuchu:
The "Input" Label is a bit weird

paulleseute:
seen with @valentinbeggi and @qchuchu, we'll settle for this:
image
changed to be implemented in a upcoming pr.

packages/devtools/src/components/layout/tools-list/tool-item.tsx
```diff
@@ -84,8 +89,13 @@ export function ToolItem({ tool, open }: { tool: Tool; open: boolean }) {
     await callTool({ toolName: tool.name, args: formData });
   };
 
-  useKeyPress("meta.enter", () => {
-    if (!open) {
+  // Enter/⌘Enter run the tool — but not while a dialog (e.g. the save-query
```
qchuchu: Remove the comment not necessary

packages/devtools/src/components/layout/tools-list/saved-queries.tsx
```diff
@@ -0,0 +1,190 @@
...
+                  )}
+                />
+              )}
+
+              <span className="text-muted-foreground italic">
+                {queries.length > 0 ? "none" : "no saved input yet"}
+              </span>
+            </DropdownMenuItem>
+            {queries.map((query) => (
+              // The trash sits as a sibling overlay (not a child of the item)
```
qchuchu: Remove comment

qchuchu (approved):
Ok if useless comments are removed + we use the terminology savedInputs instead of queries.

PR #853, opened 2026-06-09
feat: add saved inputs and run button to tool header 

paulleseute:
The main objective of this PR is to let users run queries without having to open the tool accordions, which makes playing sequences easier.

I also added a bunch of improvements on the tool pane because they update the same code and I want them to be in the next release which is very soon. Each is encapsulated within a commit. I'd recommend to review one by one since they are all pretty small.

https://github.com/user-attachments/assets/6ddeba11-218e-4388-8206-30f08f7b8d54

<h3>Greptile Summary</h3>

This PR adds a run button and saved-inputs dropdown directly in each tool's accordion header so users can execute queries without expanding the accordion, and switches the accordion from single- to multiple-open mode so several tools can be visible at once.

- The `saved-queries-store` is renamed to `saved-inputs-store` (localStorage key `skybridge-devtools-saved-queries` → `skybridge-devtools-saved-inputs`); a `TruncatedDescription` component replaces the old line-clamp + dialog pattern throughout the tool pane; and `ToolItem` drops the `open` prop, sourcing run/disabled state from the store rather than accordion open/closed state.
- E2E selectors are tightened to scope `[data-tool-name]` so the newly multiplied run buttons don't cause ambiguous matches.

Safe to merge once the persisted-data migration is addressed; all other changes are well-scoped UI improvements.

The rename of the Zustand persist storage key from skybridge-devtools-saved-queries to skybridge-devtools-saved-inputs means every existing user's saved queries are silently discarded on upgrade — the store initialises empty with no migration path. The rest of the PR is clean.

packages/devtools/src/lib/saved-inputs-store.ts — the storage key rename and missing migration.

`````markdown

---

### Issue 1 of 1
packages/devtools/src/lib/saved-inputs-store.ts:90
**Silent loss of all previously-saved queries on upgrade**

The localStorage key was renamed from `"skybridge-devtools-saved-queries"` to `"skybridge-devtools-saved-inputs"`, and the persisted state shape changed from `{ queries: … }` to `{ inputs: … }`. Because Zustand's `persist` middleware sees this as a brand-new store (different `name`), it initialises with an empty `inputs: {}` and never reads — nor offers to migrate — anything from the old key. Every user who had saved queries before this release will silently lose them with no warning and no recovery path.

A `migrate` callback keyed off `version` could read the old data and rewrite it under the new shape, or the PR could keep the original storage key and only rename the in-memory property.

`````

</details>

<sub>Reviews (3): Last reviewed commit: ["feat: s/query/input"](https://github.com/alpic-ai/skybridge/commit/d015d77965acb1b2b7616deb9f3a9a2ec4eb6572) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36431122)</sub>

### CODE REVIEW

File: packages/devtools/src/components/layout/tools-list/tool-item.tsx [hunk 1]
```diff
@@ -58,11 +55,28 @@ export function ToolItem({ tool, open }: { tool: Tool; open: boolean }) {
...
+    const hasNoInput =
+      !schema?.properties || Object.keys(schema.properties).length === 0;
+    if (hasNoInput) {
+      return false;
+    }
+    if (tab === "json" && jsonError) {
+      return true;
+    }
+    return validator.validateFormData(formData, schema).errors.length > 0;
+  }, [tool.inputSchema, tab, jsonError, formData]);
```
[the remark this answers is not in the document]
paulleseute: real cost, but it's the store, not type=multiple: useCallToolResult subscribes with no selector and updateNestedState cloneDeeps the whole tools map, so every keystroke re-renders + re-validates every tool (open or not). a debounce/touched-guard is just a band-aid — even adding a selector wouldn't help while cloneDeep churns every ref. proper fix is the store update strategy, out of scope here. negligible at our tool counts so leaving it.

packages/devtools/src/components/layout/tools-list/tool-item.tsx
```diff
@@ -135,41 +172,59 @@ export function ToolItem({ tool, open }: { tool: Tool; open: boolean }) {
...
-      {!open && tool.description ? (
-        <div aria-hidden="true" className="px-3 pb-3 -mt-1 font-sans">
-          <div className="rounded-md border border-border bg-muted/40 px-2.5 py-2 text-xs text-muted-foreground/70">
-            <span className="line-clamp-2">{tool.description}</span>
-          </div>
+      {tool.description ? (
+        <div className="px-3 pb-3 font-sans">
+          <ToolDescription name={tool.name} description={tool.description} />
         </div>
       ) : null}
```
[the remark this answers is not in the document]
paulleseute: intentional — consolidated the description into a single always-visible instance in the header so it shows whether the tool is collapsed or expanded. not a regression.

packages/devtools/src/components/layout/tools-list/truncated-description.tsx
```diff
@@ -0,0 +1,118 @@
...
+        const mid = (lo + hi) >> 1;
+        el.textContent = `${text.slice(0, mid).trimEnd()}... more`;
+        if (fits()) {
+          best = mid;
+          lo = mid + 1;
+        } else {
+          hi = mid - 1;
+        }
+      }
+      setTruncatedPrefix(text.slice(0, best).trimEnd());
```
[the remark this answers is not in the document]
paulleseute: agreed, bounded — ~8 reflows in useLayoutEffect (no paint flicker), only on mount + container resize, against one offscreen measurer. fine for a devtool, leaving as is.

packages/devtools/e2e/tests/auth.spec.ts
```diff
@@ -21,7 +21,11 @@ test.describe("devtools auth", () => {
       page.getByRole("button", { name: "whoami", exact: true }),
```
qchuchu, line 21: nothing but I remember that in the alpic project we enforced the file name of tests with test.ts

packages/devtools/src/components/layout/tools-list/truncated-description.tsx
```diff
@@ -0,0 +1,118 @@
...
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogHeader,
+  DialogTitle,
+} from "@alpic-ai/ui/components/dialog";
+import { useLayoutEffect, useRef, useState } from "react";
+import { cn } from "@/lib/utils.js";
+
+// A description clamped to two lines. When it overflows, the text is cut so an
```
qchuchu, line 11: WTF all this code to truncate text ?

qchuchu: We might have a lib for that no ?

qchuchu: And why is it part of this PR ?

packages/devtools/src/components/layout/tools-list/saved-inputs.tsx
```diff
@@ -33,14 +33,16 @@ export function SavedQueriesDropdown({
   onClear: () => void;
```
qchuchu, line 33: why does the component is still named saved-queries

packages/devtools/src/components/layout/tools-list/truncated-description.tsx
paulleseute: I also added a bunch of improvements on the tool pane because they update the same code and I want them to be in the next release which is very soon. Each is encapsulated within a commit.

Having the ellipsis + more link on the same row needs a bit of code. I don't think bringing a third party lib is any better than a custom implementation tho, the actual algo implementation is like 50 lines

packages/devtools/src/components/layout/tools-list/saved-inputs.tsx
paulleseute: @qchuchu fixed [d015d77](https://github.com/alpic-ai/skybridge/pull/853/commits/d015d77965acb1b2b7616deb9f3a9a2ec4eb6572)

packages/devtools/src/lib/saved-inputs-store.ts
```diff
@@ -0,0 +1,104 @@
...
+              [serverName]: {
+                ...forServer,
+                [toolName]: forTool.filter((s) => s.key !== key),
+              },
+            },
+          };
+        }),
+    }),
+    {
+      name: "skybridge-devtools-saved-inputs",
```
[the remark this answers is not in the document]
paulleseute: not merged

qchuchu (approved):

PR #857, opened 2026-06-10
feat: allow configuring the built-in express.json() body parser

slackerzz:
## Why
The `express.json()` middleware that `McpServer` pre-applies is hardcoded with no arguments, so its defaults (notably the 100kb body size limit) can't be changed. My app needs a larger limit, and today the only way to get one is patching the package (I'm currently using a `yarn patch` for exactly this).

## What

`McpServer` now accepts an optional third constructor argument with Skybridge-specific options, forwarded to the built-in `express.json()`:

```ts
const server = new McpServer(
  { name: "my-app", version: "1.0.0" },
  {},
  { json: { limit: "10mb" } },
);
```
 - New exported types: `SkybridgeServerOptions` and `JsonOptions`. `JsonOptions` is derived from Express's own types (`Parameters<typeof express.json>[0]`), so users get full IntelliSense on the options without skybridge depending on body-parser.
- The SDK's `ServerOptions` second argument is untouched, and the new parameter leaves room for future framework-level options.

Fully backward compatible: omitting the new argument results in `express.json(undefined)`, which behaves identically to `express.json()`.

<h3>Greptile Summary</h3>

This PR adds an optional third constructor argument to `McpServer` that forwards options to the built-in `express.json()` body parser, allowing users to configure things like the body size limit without patching the package.

- New exported types `SkybridgeServerOptions` and `JsonOptions` (derived via `Parameters<typeof express.json>[0]`) give users full IntelliSense without introducing a direct `body-parser` dependency.
- The change is fully backward-compatible: `express.json(undefined)` behaves identically to `express.json()` because `body-parser` internally does `var opts = options || {}`.
- Two integration tests verify the happy path (custom 10 MB limit accepts a 200 KB body) and the default behavior (no options → 100 KB limit → 413).

Safe to merge — the change is a small, additive constructor extension with no modifications to existing behavior.

The only runtime impact is `express.json(undefined)` replacing `express.json()`, which is semantically identical. The new code path (passing a user-supplied options object) is covered by an integration test. Types are derived directly from Express's own signature, so the public API will stay in sync automatically. No edge cases, no error-handling gaps, no side-effects on the rest of the class.

<sub>Reviews (2): Last reviewed commit: ["Merge branch &#39;main&#39; into express\_json\_co..."](https://github.com/alpic-ai/skybridge/commit/53c22499b84010f15c5c1c2ee4f1b2172f782d15) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36709353)</sub>


harijoe (approved):
Thanks for taking care of this @slackerzz, nice work!

PR #862, opened 2026-06-11
feat(doc): add build section

paulleseute:
## build

the build section is an opinionated walkthrough sitting between the architecture concept page and the api reference. the pages share one running example, a personal shopper (search-products, create-checkout, a carousel view), so the code composes across pages, and each one ends with a hand-off to the next (tools → views → state → auth → test).

what they are not: api references. they don't enumerate every option or host caveat, they link to api-reference instead, and `@todo` comments mark the spots where deeper references still need to be written.

every page has the same shape: an intro stating the design problem, one complete hero snippet, a roadmap sentence, then one section per facet. inside a section the rhythm is always generic explanation first, then the canonical snippet re-shown with line highlights on the lines being discussed, then commentary on the example. pages close with a Go Further card grid to the sibling pages. it's heavy on show-not-tell because agents, also more ~fun to read for humans (imo).

## test

the test section covers the tooling for testing an mcp app: devtools, tunnel, playground, audit. the organizing idea: each tool puts one more actor in the loop. devtools is your app alone (no model, no host), the playground adds a real model, the tunnel adds the real hosts, and the audit (powered by alpic beacon) checks the whole thing headlessly. each page's limitations name what the tool can't test, which is exactly what the next one covers.

tool pages are deliberately lighter than build guides: what problem it solves, setup, a quick feature overview (users will discover the rest by clicking around), then limitations.

the section has a home with a comparison table routing by need (real model / real host / interactive). build got the same treatment: a home page with the four questions an app answers and their core APIs. both homes are wired as group roots, so clicking "Build" or "Test" in the sidebar opens them without adding a tree item.

## ship

a single standalone page at `/ship`, reworked from the legacy deploy + build-for-production pages.

## also

- reworked the tree so it's easier to know what's new and what's legacy
- added a bunch of illustrations which are more of some kind of proposal/placeholder. very loosely inspired by [[bierstadt](https://fr.wikipedia.org/wiki/Albert_Bierstadt)](https://fr.wikipedia.org/wiki/Albert_Bierstadt), slop edition.

## reviewing

best reviewed when started locally. make sure your mintlify is up to date so all components are rendered correctly:
```
mint update
rm -rf ~/.mintlify
```


harijoe (approved):
Definitely cleaner than the previous state, well done 👍
Reading flow is smooth and the explanations are clear. The coding examples with highlighted sections are neat.

Since this part of the docs is dedicated to beginners, a few Excalidraw schemas could really help with understanding, such as:
Image

I'm not 100% convinced by the illustration images. They're beautiful and make the overall reading experience more pleasant, but they could lower the perceived quality, as they're explicitly AI-generated and might lead readers to think the docs are too. Also, beyond being beautiful, they don't add much value in helping the user understand the concepts.
I still think it's mergeable as-is, but there's room for improvement.

The images introduced are quite heavy, and Mintlify doesn't compress them automatically when published, so it could be worth asking Claude to compress them before shipping (around ~50% bytes saved).

The main challenge is keeping this doc in sync with the actual codebase. It would be very handy to ship a skill that checks whether what the doc states is accurate against the actual codebase (like what we did for the Alpic CLI). Probably part of another PR though.

The "Authenticate Users" section and its wording are a bit inaccurate: it's not only about "who's behind every tool call," since it covers both authentication and authorization. It could be worth splitting these two concepts into dedicated sections.

paulleseute:
@harijoe
- i agree we need a simpler diagram, i'll see what I can do. not a big fan of excalidraw for this use case since it looks a bit cheap imo.
- images: they are proposal/placeholder. style is interchangeable but i think having nice, playful illustration with a story arc is really important for branding/differentiation purposes. it's already compressed, i think ~200kb it's acceptable for having nice visuals.
- skill: in the works, alongside the work i'm doing on the doc
- auth: agreed, will rework this page to make it clearer

also note this pr does not merge in main

PR #871, opened 2026-06-16
fix(devtools): restore cached outputs

paulleseute:
'selecting' a tool means displaying the output. it can be achieved by
- running the tool
- clicking on the tool header, granted we have an previous output to display
selected tools are flagged with a cyan left border (on both tool header and body)

also fixed the auto-scroll on focused input: .focus() auto-scrolls the input into view, but it fires before the list layout has settled, so the browser scrolls against a half-built layout and overshoots to the bottom (where it then stays).

Screenshot 2026-06-16 at 12 07 20

<h3>Greptile Summary</h3>

This PR introduces "cached output selection" for the devtools tool list: clicking a tool's header now restores its previous output in the right pane (when a result exists), visualized via a 5 px cyan left border on the selected row. It also fixes an autofocus scroll overshoot caused by `.focus()` firing before the list layout settles, and changes the initial accordion state so only the first tool starts expanded (preventing all inputs from claiming autofocus simultaneously).

- `tool-item.tsx`: adds `hasCachedOutput` guard on the `AccordionTrigger` `onClick` to call `setSelectedTool`, adds left-border selection indicator, and switches to `focusable.focus({ preventScroll: true })`.
- `index.tsx`: narrows the initial `openTools` state to the first tool only.
- E2e tests: each test now explicitly opens the relevant accordion item before interacting, and the Vite iframe timeout is raised to 45 s for parallel runs.

The changes are narrowly scoped to UI interaction and autofocus behaviour with no data-layer or auth-path changes; safe to merge.

All three change areas (selection state, border styling, focus fix) are self-contained, the e2e tests are correctly updated to reflect the new default-open behaviour, and no store, network, or auth logic is touched.

<sub>Reviews (4): Last reviewed commit: ["fix: remove comment"](https://github.com/alpic-ai/skybridge/commit/3e6e08587d8a3838219ef21aa839f0d8ee4b24f6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37873183)</sub>

### DIFF
--- a/packages/devtools/src/components/layout/tools-list/tool-item.tsx
+++ b/packages/devtools/src/components/layout/tools-list/tool-item.tsx
@@ -52,6 +52,7 @@ export function ToolItem({ tool }: { tool: Tool }) {
   const requiresAuth = useAuthStore((s) => s.requiresAuth);
   const formRef = useRef<Form<unknown, RJSFSchema>>(null);
   const result = useCallToolResult(tool.name);
+  const hasCachedOutput = Boolean(result?.response);
   const { setToolData } = useStore();
   const formData = (result?.input ?? {}) as Record<string, unknown>;
   const setFormData = (data: Record<string, unknown> | null) => {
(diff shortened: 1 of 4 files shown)


packages/devtools/src/components/layout/tools-list/tool-item.tsx
```diff
@@ -176,9 +177,17 @@ export function ToolItem({ tool }: { tool: Tool }) {
...
       data-tool-name={tool.name}
-      className="border-b border-border transition-colors last:border-b-0 data-[state=closed]:hover:bg-muted/40"
+      className={cn(
+        // 5px left border always reserved (light grey) so the row never shifts;
+        // turns cyan when selected.
+        "border-b border-l-[5px] border-border transition-colors last:border-b-0 data-[state=closed]:hover:bg-muted/40",
+        isSelected && "border-l-cyan-400",
+      )}
     >
+      {/* Clicking the header toggles the input panel (Radix) and, when the tool
```
harijoe: Not convinced by this comment, I think the code is already quite safe-explanatory

harijoe (approved):

PR #874, opened 2026-06-17
fix: serve view resource no matter the ?v= cache key (#872)

harijoe:
Ports the #872 fix (`711743f`) onto this branch by hand instead of cherry-picking, to avoid conflicts from the 74-commit divergence.

## Differences vs `711743f`

- **`server.ts`** — additions are byte-identical (`stripQuery` helper, `viewUriByPath` map, the `resources/read` middleware that resolves any `?v=` to the canonical URI and echoes back the requested URI, and the map population in `registerViewResource`).
- **`view-resource-resolution.test.ts`** — only real difference: this branch predates the refactor that introduced the module-level `__setBuildManifest` helper, so the prod test sets the manifest via the `server.setViteManifest(...)` instance method inside the `connect` callback instead.

<h3>Greptile Summary</h3>

This PR ports the `?v=` cache-key fix from `711743f` onto the `v0` branch by hand. It adds a `stripQuery` helper and a `viewUriByPath` map (path → canonical URI) populated at tool registration time, plus a `resources/read` middleware that rewrites any incoming URI to its canonical form, delegates to the underlying handler, then restores the original URI on both the response contents and the shared request params.

- **`server.ts`**: `viewReadResolveEntry` is unconditionally inserted into the middleware chain, so the dead-code guard at lines 505-507 (flagged in a prior review) is now unreachable.
- **`view-resource-resolution.test.ts`**: Three integration scenarios — dev (no version param), prod (stale key resolves against manifest-derived canonical), and unknown path (must throw) — with `setViteManifest` correctly called before `registerWidget` inside the `connect` callback.

Safe to merge; the middleware logic is correct and the tests cover the key resolution scenarios.

The rewrite logic is sound: unknown paths pass through unmodified, the `finally` block restores request state on both success and error paths, and response content URIs are echoed back to the consumer as originally requested. No new functional defects were found beyond the pre-existing dead-code guard noted in a previous review cycle.

1. `packages/core/src/server/server.ts`, line 505-507 ([link](https://github.com/alpic-ai/skybridge/blob/180fd9832f3a6e209524d0dc1eb304a318c1c171/packages/core/src/server/server.ts#L505-L507)) 

   <a href="#">P2</a> Dead code guard — `entries.length === 0` can never be true now. `viewReadResolveEntry` is unconditionally included in the array, so the array always has at least one element and this early return is unreachable. The guard was meaningful before this PR (when the array was conditionally empty), but it's now misleading since it implies there is still a code path where no middleware is applied.

   

   `````markdown
   Path: packages/core/src/server/server.ts
   Line: 505-507

   Comment:
   Dead code guard — `entries.length === 0` can never be true now. `viewReadResolveEntry` is unconditionally included in the array, so the array always has at least one element and this early return is unreachable. The guard was meaningful before this PR (when the array was conditionally empty), but it's now misleading since it implies there is still a code path where no middleware is applied.

   

   `````
   </details>

   Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!
</details>

<sub>Reviews (4): Last reviewed commit: ["Update packages/core/src/server/view-res..."](https://github.com/alpic-ai/skybridge/commit/7410adc730211b5a29fa438df540b8a70eed3a40) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38107558)</sub>

### DIFF
--- a/packages/core/src/server/view-resource-resolution.test.ts
+++ b/packages/core/src/server/view-resource-resolution.test.ts
@@ -1,0 +1,129 @@
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
+import { afterEach, describe, expect, it } from "vitest";
+import { McpServer } from "./index.js";
+
+// Mirror what the Skybridge Vite plugin generates in `.skybridge/views.d.ts`:
+// narrow `ViewName` so `component: "widget"` typechecks against the registry.
+declare module "./server.js" {
+  interface ViewNameRegistry {
+    widget: true;
+  }
+}
+
+// SKY-435: a view resource must resolve no matter the `?v=` query param value.
+// The version param is a content-derived cache key for external consumers
+// (it busts host/CDN caches when the bundle changes); it is not part of the
+// resource's identity, so a stale, absent, or arbitrary param must still serve
+// the underlying asset.
+
+async function connect(register: (server: McpServer) => void) {
+  const server = new McpServer({ name: "test", version: "1.0.0" });
+  register(server);
+  const client = new Client({ name: "test-client", version: "1.0.0" });
+  const [clientTransport, serverTransport] =
+    InMemoryTransport.createLinkedPair();
+  await server.connect(serverTransport);
+  await client.connect(clientTransport);
+  return {
+    client,
+    teardown: async () => {
+      await client.close();
+      await server.close();
+    },
+  };
+}
+
+function registerWidget(server: McpServer) {
+  server.registerTool(
+    { name: "show", description: "show", view: { component: "widget" } },
+    () => ({ content: [{ type: "text" as const, text: "ok" }] }),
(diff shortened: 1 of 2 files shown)


harijoe:
@greptileai

packages/core/src/server/view-resource-resolution.test.ts
```diff
@@ -0,0 +1,129 @@
...
+
+// Mirror what the Skybridge Vite plugin generates in `.skybridge/views.d.ts`:
+// narrow `ViewName` so `component: "widget"` typechecks against the registry.
+declare module "./server.js" {
+  interface ViewNameRegistry {
+    widget: true;
+  }
+}
+
+// SKY-435: a view resource must resolve no matter the `?v=` query param value.
```
paulleseute: let's strip the ticket #

```suggestion
- a view resource must resolve no matter the `?v=` query param value.
```

paulleseute (approved):

PR #876, opened 2026-06-17
feat: add guides

paulleseute:
## guides

the guides are cross-cutting how-to. four new pages in the house style: csp, ux, files, migrate. auth provider tbd when the feature ships. same shape as the build pages, a hero illustration, a design-problem intro, then the generic-explanation → highlighted-snippet → example-commentary rhythm. they link to api-reference instead of enumerating every option, with `@todo` markers where deeper refs are still missing.

## also
- the everything example now exercises the full chatgpt file surface to match the guide: upload/pick, resolve a download url, hand a file to a tool, and an inspect-file tool that echoes the file back (
- dropped a `{ library: true }` upload flag that turned out to be a no-op (ticket filed to remove it from the api)

## reviewing

best reviewed started locally. make sure mintlify is current so components render:
```
mint update
rm -rf ~/.mintlify
```

<h3>Greptile Summary</h3>

- Adds a new guides section covering CSP, UX, files, and migration workflows.
- Updates docs navigation and related guide pages with cross-links and guide assets.
- Expands the everything example to demonstrate file upload, selection, download URL resolution, and passing files to a tool.

The changes are documentation-focused with example updates and no identified blocking issues.

The reviewed changes add guides, navigation entries, assets, and broaden an example’s file handling coverage without any reported defects.

<details><summary><h3><a href="https://www.greptile.com/trex">T-Rex</a> T-Rex Logs</h3></summary>

**What T-Rex did**
- Captured the base docs preview before changes and the head docs preview after changes to verify the new Guides index and route cycling.
- Inspected the guides verification log to confirm file presence, docs.json structure, hero image references, internal links, and code-fence counts for the new guides.
- Compared the everything-files UI flow before and after, noting that the before state lacks pickLibrary, preview, inspect, and result display, while the after state includes Upload, Pick from library, Preview, Inspect with tool, and Resolve.
- Verified the inspect-file tool behavior before and after, with the after state showing contract checks true and a runtime FileRef including file\_name and sizeBytes.
- Reviewed the related logs to confirm UI and tool state coverage aligned with the expected outcomes.

<a href="https://app.greptile.com/trex/runs/11497162/artifacts"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifactsDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/ViewAllArtifacts.svg?v=1">View all artifacts</picture></a>

<sub><a href="https://www.greptile.com/trex">T-Rex</a> Ran code and verified through T-Rex</sub>
</details>

<sub>Reviews (3): Last reviewed commit: ["feat: review"](https://github.com/alpic-ai/skybridge/commit/38c486973fe7575ae49877cdd0ce739e64163b94) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38126798)</sub>


harijoe (approved):
A few comments:
- I was a bit confused that the Guides section wasn't clickable, since that's not the case for the other sections. Not a blocker, but I lean slightly towards having a page for each section. For the Guides section, that could be the right spot to put the same explanation as the one in this PR description that explains what the guides are.
- I really like the new docs overall, they look way more professional and polished.

### For CSP
- I think we could hide the details of the CSP key in the first code snippet. They make the snippet complex, even though its only purpose is to show where the CSP key goes. I'd rather put something like `csp: { ... }` in highlighted syntax for this specific first snippet.
- I think for all the CSP sections, a code snippet with a minimal component/frontend that would trigger the corresponding CSP would be very valuable for understanding.
- I'm not sure about the "Allowed Automatically" section: I think this info could be an "info alert" style message within the connectDomains and resourceDomains sections instead.
- The CSP section illustration is a bit bland, it's just showing a random landscape. Maybe CSP isn't easy to represent in this image style, but I still think we can do something, since CSP evokes the idea of guarding something… open to brainstorming.

### For UX
- "host owns and can reshape at any moment." → What does it mean that the host can reshape the canvas? That's not true as far as I understand it. I think I get what it's trying to explain, but this whole section intro looks confusing to me. Maybe too generic.
- Same issue as CSP with the big code snippet at the start of the section. I think a good rule of thumb would be that there shouldn't be a code snippet without any highlighted text. As a reader, a code snippet without highlighted text makes me search for what I'm supposed to be looking at, which isn't ideal.
- We could show some CSS reacting to the `dark` class added in dark mode, to make it easier to understand what's actually expected.
- Inline/PiP/Fullscreen could use `code` syntax to highlight them nicely.
- "Preview Every Surface": maybe add a screenshot of the DevTools button bar to create a visual connection?

### For files
- The section description repeats the "Files in ChatGPT" section a bit. It's okay, but it's not very pleasant to read, since we get the same info twice within a very short span.
- I fear this section will be hard to keep accurate, though it brings a lot of value. It's probably the only place on the internet where Claude and ChatGPT file handling are compared and explained.

paulleseute:
@harijoe thanks for the feedback, applied most of your suggestions.
- "host owns and can reshape at any moment" -> it means that, at the end of the day, the host is controlling the env the view is run on and the view should be prepared to adapt. if the host for any reason want to trigger any display mode, resize the frame or wipe the shared state it theoretically could. rephrased tho.
- about the "big" snippets: the pattern introduced by this revamp for "walk-throughs":
```
1. Intro paragraph: the design problem in generic terms, no example nouns yet
2. Hero: one complete, runnable snippet
4. One `##` section per facet, in build order
```
this is the shape used by both "build" and "guides" pages. i'd to keep it that way for consistency, can be revisited later.
- "CSS reacting to the dark": i think having a class is explicit enough, handling dark mode isn't specific to mcp apps

PR #889, opened 2026-06-26
feat: add conformance app

paulleseute:
cf readme
image


harijoe:
Observations from tests on VSCode:
- All downloads are flagged approved when clicked but it's actually not doing anything. Maybe not only rely on a function not throwing but enable the tester to specify explicitely that a test didn't work for the final report ?
- Same for useRequestClose : it says Supported even though the view wasn't closed. Btw this one raises the question of how to get the report if the view has been actually closed

harijoe (approved):

PR #905, opened 2026-06-30
feat: support x-forwarded-prefix

paulleseute:
as of today assets are fetched on scheme+host. openai requires path prefix for versioning, so we need to support reverse proxy routing based on path for the assets to be fetched on the right url. x-forwarded-prefix is an unofficial but accepted way of achieving this. it'll be injected by alpic cloud for instance. refer to doc for more details

docs get a new "Ship" group: deploy + versioning


harijoe (approved):
I think we can shrink tests by ~90% of lines while testing most of it using only unit tests instead of the full server setup.

paulleseute:
> I think we can shrink tests by ~90% of lines while testing most of it using only unit tests instead of the full server setup.

issues i had during dev were bugs in wiring (header key, origin+prefix, assets/assets doubling), only visible in rendered HTML from a real request. units wouldn't catch them, so i'd keep the setup.

PR #910, opened 2026-07-02
fix: remove minimumReleaseAge, bump skybridge & alpic

paulleseute:
minimumReleaseAge  it makes things difficult wrt skybridge or alpic version bump. more annoying than useful


harijoe (commented):
A leaner solution to fix the problem : 
- Add recharts to minimumReleaseAgeExclude temporarily
- Remove pnpm-lock and add it to gitignore

We should also add the same minimumReleaseAge policy to Alpic to avoid this problem in the future (as the root cause is Alpic enforcing in its package.json a version that only resolves to too recent releases)

Issue #533, opened 2026-03-06
Skybridge build fails when no view is registered


neil-ac:
### Context

Hey there :) I wanted to use Skybridge as a framework for a regular MCP server (no ChatGPT/MCP App). 

I registered a tool in the server/ folder and have no web/ folder, and tried to build the project but got the following error:
```
✘ [ERROR] Could not resolve "/Users/neil/Build/folder/apps/mcp-server/web/vite.config.ts"

failed to load config from /Users/neil/Build/folder/apps/mcp-server/web/vite.config.ts
error during build:
Error: Build failed with 1 error:
error: Could not resolve "/Users/neil/Build/folder/apps/mcp-server/web/vite.config.ts"
    at failureErrorWithLog (/Users/neil/Build/folder/node_modules/.pnpm/esbuild@0.27.3/node_modules/esbuild/lib/main.js:1467:15)
    at /Users/neil/Build/folder/node_modules/.pnpm/esbuild@0.27.3/node_modules/esbuild/lib/main.js:926:25
    at runOnEndCallbacks (/Users/neil/Build/folder/node_modules/.pnpm/esbuild@0.27.3/node_modules/esbuild/lib/main.js:1307:45)
    at buildResponseToResult (/Users/neil/Build/folder/node_modules/.pnpm/esbuild@0.27.3/node_modules/esbuild/lib/main.js:924:7)
    at /Users/neil/Build/folder/node_modules/.pnpm/esbuild@0.27.3/node_modules/esbuild/lib/main.js:951:16
    at responseCallbacks.<computed> (/Users/neil/Build/folder/node_modules/.pnpm/esbuild@0.27.3/node_modules/esbuild/lib/main.js:603:9)
    at handleIncomingPacket (/Users/neil/Build/folder/node_modules/.pnpm/esbuild@0.27.3/node_modules/esbuild/lib/main.js:658:12)
    at Socket.readFromStdout (/Users/neil/Build/folder/node_modules/.pnpm/esbuild@0.27.3/node_modules/esbuild/lib/main.js:581:7)
    at Socket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:561:12)
```

### Expected behaviour

I should be able to use Skybridge only for the server part without having to add a placeholder vite.config and widget in the web/ folder.

PR #913, opened 2026-07-03
fix(build): skip vite build for view-less MCP servers

### LINKED ISSUE #912: skybridge build fails for view-less servers: empty Rollup input ([INVALID_OPTION]) — regression of #533
### Summary

A Skybridge app with **no views** (a headless MCP server that only exposes tools/resources — e.g. one serving `skill://` resources) cannot be built. `skybridge build` crashes at the `Building views` step because the Vite plugin passes an empty `rollupOptions.input`, which rolldown rejects.

`skybridge dev` works fine — **this only affects `skybridge build`** (and therefore `deploy`).

This looks like a regression of the (closed) #533 "Skybridge build fails when no view is registered". That report was against the old `web/` architecture and surfaced a different error (`Could not resolve web/vite.config.ts`); the current v1.2.3 architecture reintroduces the same "no views" gap with a new failure mode.

### Environment

- `skybridge` v1.2.3
- `vite` v8.1.3 / `rolldown` v1.1.4
- Node 24, pnpm
- Reproducible on a brand-new blank template with zero views (not caused by any user code).

### Steps to reproduce

1. Scaffold a blank Skybridge server (or take the template as-is) with **no** files under `src/views/`.
2. Run `pnpm build`.

### Actual behavior

```
vite v8.1.3 building client environment for production...
✗ Build failed in 11ms
[INVALID_OPTION] You must supply `options.input` to rolldown, you should at least
provide one entrypoint via `options.input` or `this.emitFile({type: 'chunk', ...})`
```

`Scanning views` and `Compiling server` pass; `Building views` fails.

### Expected behavior

Building a Skybridge app with no views should succeed. A view-less server is a legitimate use case (headless MCP server exposing only tools/resources).

### Root cause

In `dist/web/plugin/plugin.js`, the `config()` hook builds `input` from discovered views and returns it unconditionally:

```js
const input = {};
for (const view of views) {
  input[view.name] = `${VIRTUAL_PREFIX}${view.name}`;
}
return { build: { rollupOptions: { input } /* … */ } };
```

When there are no views, `input` is `{}`, and rolldown throws `[INVALID_OPTION]`.

### Proposed fix

When there are **no views**, skip the `vite build` (client) step entirely rather than invoking rolldown with an empty input. The build should then emit an empty-but-valid manifest so the downstream steps still run.

**Open question — do the downstream deploy steps handle a skipped/empty client build?** The build pipeline currently runs, after `Building views`:

- `Emitting manifest module` — reads `dist/assets/.vite/manifest.json`
- `Emitting entry wrapper`
- `Emitting Cloudflare redirects` / `Emitting Cloudflare headers`
- `Emitting Vercel build output` (esbuild-bundles `dist/__entry.js`, copies `dist/assets`)

If `vite build` is skipped, `dist/assets/.vite/manifest.json` and `dist/assets/` won't exist. Please confirm whether the **Cloudflare** and **Vercel** build outputs are impacted and need to tolerate a missing/empty manifest and assets directory (e.g. emit an empty manifest module and skip the assets copy).

### Current workaround

Supply a minimal client entry in the app's own `vite.config.ts`:

```ts
build: {
  rollupOptions: {
    input: { index: fileURLToPath(new URL("./src/client.ts", import.meta.url)) },
  },
}
```

(where `src/client.ts` is an empty `export {}`). Skybridge's `input: {}` deep-merges under this, letting the build complete.

syf2211:
## Summary

Skip the client `vite build` step when a Skybridge app has no registered views, and emit an empty Vite manifest so downstream deploy steps still succeed.

## Motivation

`skybridge build` fails for headless MCP servers that only expose tools/resources (no views). Rolldown rejects an empty `rollupOptions.input` during the "Building views" step. `skybridge dev` works fine — this only affects production builds (and therefore deploy).

Fixes #912 (regression of #533).

## Changes

- Add `getCommandSteps()` to discover views before assembling the build pipeline
- Skip the "Building views" step when no views are found
- Add `emitEmptyManifestModule()` for view-less builds
- Add `ensureAssetsDir()` so Cloudflare redirect/header files can be written without a prior vite output
- Make `emitVercelBuildOutput()` tolerate a missing `dist/assets` directory

## Tests

```bash
pnpm --filter skybridge test   # 308 passed (48 files)
pnpm --filter skybridge build  # success
```

New unit tests cover:
- view-less step list omits "Building views"
- view-present step list includes "Building views"
- empty manifest emission
- `ensureAssetsDir` / missing assets in Vercel output

## Notes

- composer-2.5 review: **APPROVE**
- Direct `vite build` with empty input is out of scope; this fixes the `skybridge build` CLI path described in the issue
- Invalid view files (missing default export) are treated as no views, matching the vite plugin behavior


harijoe:
Hi @syf2211, thanks for your contribution!

As the contributor guidelines state, we require that either Greptile gives your PR a 5/5, or that you respond to the issues it raised explaining why you disagree with its review.

Also, since this is your first time contributing to the Skybridge repository, would you mind introducing yourself on [our Discord](https://discord.gg/awV4gu74wK)? We'd love to get to know our contributors.

syf2211:
Thanks for the review @harijoe — I've addressed both Greptile findings in 43b3e83:

1. **rmSync path**: the "Compiling server" step now uses `path.join(root, "dist")` so cleanup respects the `root` parameter, consistent with every other step.
2. **Duplicate view error UX**: `hasViews` now uses non-throwing `scanViewsSync(...).valid.length`; duplicate-name validation still runs inside the "Scanning views" step via `scanAndWriteViewsDts` → `discoverViewsSync`, so `useExecuteSteps` can render the styled Ink error instead of an unhandled rejection before `render()`.

All 308 unit tests pass (`pnpm --filter skybridge test:unit`). Happy to join Discord as well — will drop an intro there shortly.

harijoe (approved):

harijoe:
Thanks @syf2211 for your contribution!

PR #931, opened 2026-07-07
fix(devtools): stop widget e2e flake and surface flakes in CI

### LINKED ISSUE #860: 1 flaky test in devtools e2e suits
Just noticed we have a flaky test for the iframe rendering on views.
We should :
- make flaky test fail the E2E test suite
- fix this flaky test

```
1) [chromium] › e2e/tests/smoke.spec.ts:31:3 › devtools smoke › calls a widget tool and renders inside the iframe 

    Error: expect(locator).toBeVisible() failed

    Locator: locator('iframe[title="html-preview"]').contentFrame().getByText('card-3083a5e2-0db0-4228-9573-abf9003e3cd9')
    Expected: visible
    Timeout: 20000ms
    Error: element(s) not found

    Call log:
      - Expect "toBeVisible" with timeout 20000ms
      - waiting for locator('iframe[title="html-preview"]').contentFrame().getByText('card-3083a5e2-0db0-4228-9573-abf9003e3cd9')


      43 |     // First-time view compilation by Vite can take a few seconds.
      44 |     const widget = page.frameLocator('iframe[title="html-preview"]');
    > 45 |     await expect(widget.getByText(token)).toBeVisible({ timeout: 20_000 });
         |                                           ^
      46 |   });
      47 | });
      48 |
        at /home/runner/work/skybridge/skybridge/packages/devtools/e2e/tests/smoke.spec.ts:45:43

    Error Context: test-results/smoke-devtools-smoke-calls-4b660-d-renders-inside-the-iframe-chromium/error-context.md

  1 flaky
    [chromium] › e2e/tests/smoke.spec.ts:31:3 › devtools smoke › calls a widget tool and renders inside the iframe 
  16 passed (51.2s)
```

harijoe:
Closes #860

## Summary
- `packages/core` vite plugin: drop `exclude: ["skybridge/web"]` from view dep optimization (net core change: one deleted line)
- e2e: isolate each dev server's Vite dep-optimizer cache (per-port `cacheDir` for the 3 fixtures and the 3 devtools instances, parked under `node_modules/`)
- e2e: `retries: 0` so a flake fails the suite; `trace: "retain-on-failure"` so failures keep their trace without retries
- `deploy.spec`: retrying `hoverDeployPopover` helper + `dispatchEvent("click")` for in-popover buttons

## Root causes
1. **User-facing framework bug.** `exclude`-ing `skybridge/web` (introduced silently in b6d4c14 / #707) also hides it from Vite's dep scanner, so its transitive deps (`ext-apps`, `sdk/types.js`, `superjson`, …) are discovered lazily on first view mount. On a cold `.vite` cache that triggers `optimized dependencies changed → full-reload`, which blanks the `doc.write` iframe: the view silently never mounts, no matter how long the test waits. This affects any user app on first `skybridge dev`. With no mention of `skybridge/web` at all, Vite does the right thing in both worlds: workspace-linked (framework contributors) → served fresh from source and crawled by the scanner; npm-installed (users) → auto-discovered from the view entries and prebundled with transitives inlined. (The "first-time compilation is slow" comment in the spec was a misdiagnosis of this.)
2. **e2e cache collision.** The fixture app has no own `package.json`, so its default Vite `cacheDir` resolved up to `packages/devtools/node_modules/.vite` — shared by all 6 concurrent dev servers. Optimizers clobbered each other's dep files mid-run (one artifact was a 0-byte chunk), breaking module fetches and forcing reloads. Test-only: a real `skybridge dev` runs one views server per project with its own `node_modules/.vite`, and devtools is served prebuilt static.

Separately, the deploy popover tests raced the hover-close timer: a toolbar re-render after load steals hover and closes the popover; a real click also moves the mouse off the trigger. Hence the retrying hover helper and programmatic clicks.

## Verification
- Pre-fix: 1/10 local repeats failed, matching CI
- Post-fix: 7× full-suite cold-cache runs (all `.vite` caches wiped, `CI=1`), 27/27 each — covering the linked-package case
- Installed-package case: packed `skybridge` into a tarball, installed it in a scratch app with one view; on cold start `skybridge/web` is prebundled at boot (present in optimizer metadata) and serving the view triggers no re-optimization
- Ablations confirm each fix is load-bearing: scanner fix without cache isolation → `Failed to fetch .../echo-card`; cache isolation without the scanner fix → late `sdk/types.js` optimization reload; both vite fixes without the deploy helper → 3 hover failures. A `test.slow()` workaround proved unnecessary once the reload was fixed and was dropped.
- Core unit tests: 311 passed; CI green on cold runners with `retries: 0`


packages/devtools/e2e/tests/deploy.spec.ts
```diff
@@ -235,16 +247,15 @@ test.describe("deploy button", () => {
...
     await expect(page.getByText(/last deployed from git/i)).toBeVisible();
     await expect(page.getByText("main")).toBeVisible();
     await expect(page.getByText("octocat")).toBeVisible();
     await expect(page.getByText(/Fix the thing/)).toBeVisible();
 
-    // force: the popover is mid-animation in headless CI, skip the stability wait.
     await page
       .getByRole("button", { name: /deploy anyway/i })
-      .click({ force: true });
+      .dispatchEvent("click");
```
fredericbarthelet, line 258: why isn't the original `click` API not working on this specific button?

packages/devtools/e2e/tests/deploy.spec.ts
```diff
@@ -277,14 +288,14 @@ test.describe("deploy button", () => {
...
-    await deployTrigger(page).hover();
+    await hoverDeployPopover(page);
     await expect(page.getByText(/last deployed from git/i)).toBeVisible();
     await expect(
       page.getByText("https://stale-session.alpic.live"),
     ).toHaveCount(0);
     await page
       .getByRole("button", { name: /deploy anyway/i })
-      .click({ force: true });
+      .dispatchEvent("click");
```
fredericbarthelet, line 298: same here

fredericbarthelet (approved):
Small comments wondering what's wrong with the click API, but good to go otherwise :)

packages/devtools/e2e/tests/deploy.spec.ts
harijoe: click() moves the cursor before clicking, problem is that we're in a popover so it was a potential source of flakiness. Using dispatchEvent avoids this.

PR #938, opened 2026-07-08
feat: mixed-auth on branded/custom providers

harijoe:
### Summary

- Lets branded/custom OAuth providers serve public and authenticated tools from one server.
- Per-tool `auth` field `{ public?: boolean; scopes?: string[] }`, mutually exclusive with the raw `securitySchemes` escape hatch.
- Anonymous by default once any tool is public; a protected tool returns `401/403 + WWW-Authenticate` (Claude) or an in-band `mcp/www_authenticate` result (ChatGPT) to trigger sign-in.
- With a server `oauth` provider set, undeclared tools default to sign-in required (secure default); with no provider, every tool is public.
- Adds the `auth-descope-mixed` example and refreshes the auth docs.

### Notes

- Per-tool scopes are enforced whenever an `oauth` provider is set (was advertised-only). Changelog + minor bump.
- Verified end to end on Claude and ChatGPT via tunnel. Does not yet work when deployed behind Alpic (see below).
- Supersedes #904 (same work, branch renamed to match Linear SKY-486).

### Why it doesn't work on Alpic yet

The server is correct: deployed as-is it serves `/.well-known/oauth-protected-resource` and emits `WWW-Authenticate` challenges (verified on the built dev package in production mode). The blocker is Alpic's connect-time auth detection.

Alpic's runtime probes the server once at startup and classifies it purely by the HTTP status of an unauthenticated `initialize`:

- `200` → treated as fully public, OAuth discovery never surfaced.
- `401 + WWW-Authenticate` → OAuth discovered and served.

A mixed-auth server returns `200` on `initialize` (it must serve anonymous callers; only individual tools require auth), so Alpic files it as fully public, never advertises its OAuth, and ChatGPT reports "does not implement OAuth" then connects anonymously. Protected tools then can't trigger sign-in.

The model treats `public` and `has-OAuth` as mutually exclusive; mixed auth is `public` **and** `has-OAuth`. Fix is on the Alpic platform side (detection + environment metadata model), not in this framework. Works on Claude / direct tunnel because the client discovers the server's own `/.well-known` directly.

### Conscious decision: a public tool challenges an under-scoped signed-in caller

For `{ public: true, scopes }`, scopes are enforced **once a bearer token is present**:

- No token → anonymous path, allowed.
- Token with the scopes → allowed.
- Token missing a scope → `insufficient_scope` (403 / in-band `mcp/www_authenticate`), so the host prompts a step-up.

Chosen deliberately over pure OR-semantics (where a signed-in-but-under-scoped caller would fall through to the anonymous path). A signed-in user should be prompted to complete the missing scopes rather than silently downgraded to the guest experience.

Tradeoff accepted: a signed-in caller can no longer use the anonymous/degraded path on such a tool. If a tool should stay fully open to signed-in callers regardless of scope, don't list scopes on the public form. If a scope must be a hard gate for everyone, make the tool non-public with `{ scopes }`.

Reverses the earlier advertise-only revert. Spec references (RFC 6750 §3.1 `insufficient_scope`, MCP runtime scope challenge): greptile thread on `security-schemes.ts` R27-R29.

Closes SKY-486 · https://linear.app/alpic/issue/SKY-486/implement-mixed-auth

### DIFF
--- a/docs/api-reference/register-tool.mdx
+++ b/docs/api-reference/register-tool.mdx
@@ -154,7 +154,7 @@ The recommended way to declare a tool's auth. With the [`oauth`](/api-reference/
 | `{ public: true }` | Callable signed out; uses the token when one is present. |
 | `{}` | Requires sign-in. |
 | `{ scopes: [...] }` | Requires sign-in with every listed scope. |
-| `{ public: true, scopes: [...] }` | Callable signed out; advertises the scopes for signed-in callers. |
+| `{ public: true, scopes: [...] }` | Callable signed out; a signed-in caller must carry every listed scope. |
 
 Declaring any `{ public: true }` tool turns the server [mixed](/build/auth#mix-public-and-authenticated-tools): it serves anonymous callers, and every other tool stays sign-in-gated by default. Requiring sign-in (`{}` or `{ scopes }`) needs the `oauth` provider and throws without it; `{ public: true }` works either way. Omit `auth` entirely for the secure default (sign-in required whenever the server has auth).
 
--- a/docs/api-reference/register-tool.mdx
+++ b/docs/api-reference/register-tool.mdx
@@ -154,7 +154,7 @@ The recommended way to declare a tool's auth. With the [`oauth`](/api-reference/
 | `{ public: true }` | Callable signed out; uses the token when one is present. |
 | `{}` | Requires sign-in. |
 | `{ scopes: [...] }` | Requires sign-in with every listed scope. |
-| `{ public: true, scopes: [...] }` | Callable signed out; a signed-in caller must carry every listed scope. |
+| `{ public: true, scopes: [...] }` | Callable signed out; advertises the scopes for signed-in callers (step-up hint, not enforced). |
 
 Declaring any `{ public: true }` tool turns the server [mixed](/build/auth#mix-public-and-authenticated-tools): it serves anonymous callers, and every other tool stays sign-in-gated by default. Requiring sign-in (`{}` or `{ scopes }`) needs the `oauth` provider and throws without it; `{ public: true }` works either way. Omit `auth` entirely for the secure default (sign-in required whenever the server has auth).
 
--- a/docs/api-reference/register-tool.mdx
+++ b/docs/api-reference/register-tool.mdx
@@ -147,7 +147,7 @@ type ViewCsp = {
 
 ### `auth`
 
-The recommended way to declare a tool's auth. Requires the [`oauth`](/api-reference/mcp-server#constructor) provider option; with it, Skybridge enforces the declaration for you (anonymous or under-scoped calls are rejected before the handler runs).
+The recommended way to declare a tool's auth. With the [`oauth`](/api-reference/mcp-server#constructor) provider option set, Skybridge enforces the declaration for you (anonymous or under-scoped calls are rejected before the handler runs).
 
 | Value | Meaning |
 | --- | --- |
@@ -155,11 +155,11 @@ The recommended way to declare a tool's auth. Requires the [`oauth`](/api-refere
 | `"required"` | Requires sign-in. |
 | `{ scopes: [...] }` | Requires sign-in with every listed scope. |
 
-Declaring any `"public"` tool turns the server [mixed](/build/auth#mix-public-and-authenticated-tools): it serves anonymous callers, and every other tool stays sign-in-gated by default. Setting `auth` without an `oauth` provider throws, as does setting both `auth` and `securitySchemes`.
+Declaring any `"public"` tool turns the server [mixed](/build/auth#mix-public-and-authenticated-tools): it serves anonymous callers, and every other tool stays sign-in-gated by default. `"required"` and `{ scopes }` need the `oauth` provider and throw without it; `"public"` works either way. If both `auth` and `securitySchemes` are set, `securitySchemes` wins.
 
 ### `securitySchemes`
 
-The low-level form behind `auth`, for cases it can't express (several alternative `oauth2` schemes, or pure `noauth` that never accepts a token). Mutually exclusive with `auth`.
+The low-level form behind `auth`, for cases it can't express (several alternative `oauth2` schemes, or pure `noauth` that never accepts a token). Takes precedence over `auth` when both are set.
 
 ```ts
 type SecurityScheme =
--- a/docs/api-reference/register-tool.mdx
+++ b/docs/api-reference/register-tool.mdx
@@ -61,7 +61,7 @@ type ToolConfig = {
   outputSchema?: ZodRawShape; // Zod shape; types structuredContent
   annotations?: ToolAnnotations; // standard MCP hints
   view?: ViewConfig; // bind a view to render the result
-  auth?: { public?: boolean; scopes?: string[] }; // per-tool auth
+  auth?: { allowsAnonymous?: boolean; scopes?: string[] }; // per-tool auth
   securitySchemes?: SecurityScheme[]; // low-level alternative to `auth` (mutually exclusive)
   _meta?: ToolMeta; // extra metadata
 };
@@ -151,12 +151,12 @@ The recommended way to declare a tool's auth. With the [`oauth`](/api-reference/
 
 | Value | Meaning |
 | --- | --- |
-| `{ public: true }` | Callable signed out; uses the token when one is present. |
-| `{}` | Requires sign-in. |
+| _omitted_ (default) | Requires sign-in, the secure default when the server has an `oauth` provider. |
+| `{ allowsAnonymous: true }` | Callable signed out; uses the token when one is present. |
 | `{ scopes: [...] }` | Requires sign-in with every listed scope. |
-| `{ public: true, scopes: [...] }` | Callable signed out; a signed-in caller missing a listed scope gets a step-up challenge (`insufficient_scope`) instead of the anonymous path. |
+| `{ allowsAnonymous: true, scopes: [...] }` | Callable signed out; a signed-in caller missing a listed scope gets a step-up challenge (`insufficient_scope`) instead of the anonymous path. |
 
-Declaring any `{ public: true }` tool turns the server [mixed](/build/auth#mix-public-and-authenticated-tools): it serves anonymous callers, and every other tool stays sign-in-gated by default. Requiring sign-in (`{}` or `{ scopes }`) needs the `oauth` provider and throws without it; `{ public: true }` works either way. Omit `auth` entirely for the secure default (sign-in required whenever the server has auth).
+Declaring any `{ allowsAnonymous: true }` tool turns the server [mixed](/build/auth#mix-public-and-authenticated-tools): it serves anonymous callers, and every other tool stays sign-in-gated by default. Requiring sign-in (`{ scopes }`) needs the `oauth` provider and throws without it; `{ allowsAnonymous: true }` works either way. Omit `auth` entirely for the secure default (sign-in required whenever the server has auth).
 
 ### `securitySchemes`
 
--- a/docs/api-reference/register-tool.mdx
+++ b/docs/api-reference/register-tool.mdx
@@ -160,7 +160,7 @@ Declaring any `{ public: true }` tool turns the server [mixed](/build/auth#mix-p
 
 ### `securitySchemes`
 
-The low-level form behind `auth`, for cases it can't express (several alternative `oauth2` schemes, or pure `noauth` that never accepts a token). Mutually exclusive with `auth` — setting both is a compile error.
+The low-level form behind `auth`, for cases it can't express (several alternative `oauth2` schemes, or pure `noauth` that never accepts a token). Mutually exclusive with `auth`.
 
 ```ts
 type SecurityScheme =
--- a/docs/api-reference/register-tool.mdx
+++ b/docs/api-reference/register-tool.mdx
@@ -61,7 +61,8 @@ type ToolConfig = {
   outputSchema?: ZodRawShape; // Zod shape; types structuredContent
   annotations?: ToolAnnotations; // standard MCP hints
   view?: ViewConfig; // bind a view to render the result
-  securitySchemes?: SecurityScheme[]; // declare per-tool auth
+  auth?: "public" | "required" | { scopes: string[] }; // per-tool auth
+  securitySchemes?: SecurityScheme[]; // low-level alternative to `auth`
   _meta?: ToolMeta; // extra metadata
 };
 ```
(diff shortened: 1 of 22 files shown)


packages/core/src/server/auth/setup.ts
```diff
@@ -36,42 +64,79 @@ export function setupOAuth(app: Express, config: OAuthConfig): void {
...
+  app.use("/mcp", (req, res, next) => {
+    const body = req.body as
+      | {
+          id?: string | number | null;
+          method?: string;
+          params?: { name?: string };
+        }
+      | undefined;
+    const tool = body?.params?.name;
+    if (body?.method !== "tools/call" || !tool || !schemesByTool.has(tool)) {
```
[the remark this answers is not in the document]
harijoe: Fixed in the latest push. The `/mcp` gate now normalizes the body to an array and evaluates every `tools/call` in it, so an anonymous protected call inside a batch returns a transport `401` + `WWW-Authenticate` for spec/non-OpenAI clients (in-band clients fall through to the per-invocation backstop, which emits the challenge per call). Added a test asserting the batch path returns `401` with `resource_metadata`.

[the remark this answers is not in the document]
harijoe: Fixed in the latest push. The `/mcp` gate now normalizes the body to an array and evaluates every `tools/call` in it, so an anonymous protected call inside a batch returns a transport `401` + `WWW-Authenticate` for spec/non-OpenAI clients (in-band clients fall through to the per-invocation backstop, which emits the challenge per call). Added a test asserting the batch path returns `401` with `resource_metadata`.

packages/core/src/server/auth/security-schemes.ts
```diff
@@ -0,0 +1,78 @@
...
+  scopes: string[];
+};
+
+export function evaluateSecuritySchemes(
+  schemes: SecurityScheme[] | undefined,
+  authInfo: AuthInfo | undefined,
+): SchemeFailure | undefined {
+  if (securitySchemesAllowAnonymous(schemes)) {
+    return undefined;
+  }
```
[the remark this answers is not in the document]
harijoe: Intended behavior, not a bypass. `{ public: true }` means anonymous is allowed, so the tool can never hard-require a scope — a signed-in caller lacking the scope must still be able to run it (otherwise an authenticated user would be blocked from something a guest can do). The scopes on a public tool are advertised (for the host consent screen / step-up hint), and the handler branches on `extra.authInfo.scopes` to decide the enhanced path. This is the documented contract: `{ public: true, scopes }` = "callable signed out; advertises the scopes for signed-in callers". Use `{ scopes }` without `public` when a scope must be enforced before the handler runs.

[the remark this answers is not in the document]
harijoe: Fixed in the latest push. `evaluateSecuritySchemes` now only short-circuits on `noauth` when there is **no** token; once a bearer token is present it must satisfy the `oauth2` scopes, so an under-scoped signed-in caller gets `insufficient_scope` on `{ public: true, scopes: [...] }`. Anonymous callers still pass (the point of `public`). Both the `/mcp` gate and the handler backstop go through this helper, so the transport and in-band paths are consistent. Added a test covering anonymous-pass / under-scoped-reject / scoped-pass.

examples/auth-descope-mixed/src/server.ts
```diff
@@ -0,0 +1,75 @@
...
+    }),
+  },
+)
+  .registerTool(
+    {
+      name: "browse-catalog",
+      description:
+        "Browse the public coffee catalog. Works signed out, and greets you by name when a token is present.",
+      inputSchema: {},
+      auth: { public: true },
```
fredericbarthelet: do you really need a separate exemple app? Can't do with the existing one instead?

examples/auth-descope-mixed/src/server.ts
```diff
@@ -0,0 +1,75 @@
...
+        ],
+      };
+    },
+  )
+  .registerTool(
+    {
+      name: "whoami",
+      description: "Return the signed-in user. Requires sign-in.",
+      inputSchema: {},
+      auth: {},
```
fredericbarthelet: this feel strange, is it the same as providing no `auth` config?

docs/api-reference/register-tool.mdx
```diff
@@ -61,7 +61,8 @@ type ToolConfig = {
   outputSchema?: ZodRawShape; // Zod shape; types structuredContent
   annotations?: ToolAnnotations; // standard MCP hints
   view?: ViewConfig; // bind a view to render the result
-  securitySchemes?: SecurityScheme[]; // declare per-tool auth
+  auth?: { public?: boolean; scopes?: string[] }; // per-tool auth
```
fredericbarthelet: now that we had the conversation over on alpic infra implementation, I feel like `allowsAnonymous` is more inline with what this settings do than public. WDYT? @paulleseute you were part of the interface discussion as well, your feedback on this last minute naming change is valuable too :)

docs/api-reference/register-tool.mdx
```diff
@@ -144,9 +145,22 @@ type ViewCsp = {
...
 | `baseUriDomains` | `<base href>` origins (MCP Apps only). |
 
+### `auth`
+
+The recommended way to declare a tool's auth. With the [`oauth`](/api-reference/mcp-server#constructor) provider option set, Skybridge enforces the declaration for you (anonymous or under-scoped calls are rejected before the handler runs).
+
+| Value | Meaning |
+| --- | --- |
+| `{ public: true }` | Callable signed out; uses the token when one is present. |
+| `{}` | Requires sign-in. |
```
fredericbarthelet: this options should not be documented I think (not that it should be enforced at runtime too, but it feels strange to put an empty object). Or we could make `allowsAnonymous` not optional to guarantee that `auth` config is never an empty object, WDYT?

packages/core/src/server/server.ts
```diff
@@ -325,25 +333,31 @@ type AddTool<
...
-   * the right scopes through the OAuth flow. Listing both `noauth` and
-   * `oauth2` signals that the tool works for anonymous callers and gives
-   * enhanced behavior to authenticated ones.
-   */
-  securitySchemes?: SecurityScheme[];
   _meta?: ToolMeta;
 }
 
+type ToolConfig<TInput extends ZodRawShapeCompat | AnySchema> =
+  | (ToolConfigBase<TInput> & { auth?: ToolAuth; securitySchemes?: never })
```
fredericbarthelet: you should provide js doc on auth too and on each of its param as well

packages/core/src/server/server.ts
```diff
@@ -325,25 +333,31 @@ type AddTool<
...
-   * Lets clients label tools that require sign-in before calling, and pass
-   * the right scopes through the OAuth flow. Listing both `noauth` and
-   * `oauth2` signals that the tool works for anonymous callers and gives
-   * enhanced behavior to authenticated ones.
-   */
-  securitySchemes?: SecurityScheme[];
   _meta?: ToolMeta;
 }
 
+type ToolConfig<TInput extends ZodRawShapeCompat | AnySchema> =
```
fredericbarthelet: provide a dedicated intermediary interface with auth or securitySchemes named `ToolAuthConfig` without type parameters to use as base or intersection for ToolConfig

packages/core/src/server/server.ts
```diff
@@ -483,6 +497,30 @@ export function __setBuildManifest(
   pendingBuildManifest = manifest;
 }
 
+// Collapses registerTool's two overloads into a single { config, cb } shape.
+//   registerTool("greet", { description }, handler)
+//     -> { config: { name: "greet", description }, cb: handler }
+//   registerTool({ name: "greet", description }, handler)
+//     -> { config: { name: "greet", description }, cb: handler }
+function normalizeRegisterToolArgs(args: unknown[]): {
```
fredericbarthelet, line 565: why is this needed now?

docs/api-reference/register-tool.mdx
fredericbarthelet: now that I see it in action, it feels weird to have this field named `oauth` on server config, and `auth` on tool config. Worth aligning?

packages/core/src/server/server.ts
```diff
@@ -1291,7 +1394,10 @@ export class McpServer<
       this.registerViewResources(name, view, toolMeta);
     }
 
-    const wrappedCb = this.wrapHandler(cb, { attachViewUUID: Boolean(view) });
+    const wrappedCb = this.decorateToolHandler(cb, {
```
fredericbarthelet, line 1477: is there a specific reason we use a custom wrapper here instead of our middleware system to decorate tool/call requests?

packages/core/src/server/server.ts
```diff
@@ -745,18 +794,39 @@ export class McpServer<
       },
     };
 
+    // ChatGPT reads `securitySchemes` at the tool descriptor top level (SEP-1488,
+    // still Draft), but the SDK's registerTool strips unknown top-level fields, so
+    // it's stashed in `_meta` at registration. This restores it to the top level
+    // on tools/list output. Remove once SEP-1488 lands and the SDK preserves it.
+    //   { name: "checkout", _meta: { securitySchemes: [{ type: "oauth2" }] } }
+    //     -> { name: "checkout", _meta: {…}, securitySchemes: [{ type: "oauth2" }] }
+    const toolsListSecuritySchemesEntry: McpMiddlewareEntry = {
```
fredericbarthelet, line 882: this is not needed until the SEP lands (and therefore, until the SDK implements it)

packages/core/src/server/auth/setup.ts
```diff
@@ -36,42 +66,78 @@ export function setupOAuth(app: Express, config: OAuthConfig): void {
...
+      method?: string;
+      params?: { name?: string };
+    };
+    const body = req.body as Call | Call[] | undefined;
+    const calls = Array.isArray(body) ? body : [body];
+    const auth = (req as Request & { auth?: AuthInfo }).auth;
+    let failure: SchemeFailure | undefined;
+    for (const call of calls) {
+      const tool = call?.params?.name;
+      if (call?.method === "tools/call" && tool && schemesByTool.has(tool)) {
```
fredericbarthelet, line 113: wouldn't this be easier defined using a middleware too?

fredericbarthelet (commented):
Made a few comments. Overall good, some last minute adjustments that might be worth taking.
Also on the implementation side, some could be easier done via middleware, especially the auth info to security schemes evaluation

docs/api-reference/register-tool.mdx
paulleseute: yea that might be clearer

docs/api-reference/register-tool.mdx
paulleseute: the empty object is weird, i guess we should have something like 

```
undefined (default): requires sign in (when server has auth)
```

examples/auth-descope-mixed/src/server.ts
harijoe: I think it deserves a dedicated example app :
- The behavior is quite different from the server-auth example as it allows anonymous users
- Mixed-auth is not easy to grasp so having a dedicated example advertises different tools to show the different cases (public, with scopes etc.)

Not a strong opinion on this though.

docs/api-reference/register-tool.mdx
harijoe: Problem if we consider undefined as requiring sign-in is that we will introduce a change of behaviour for users bumping Skybridge.

But that's okay if documented properly in the changelog. Okay to give a try to @paulleseute suggestion.

packages/core/src/server/server.ts
harijoe: But Chatgpt docs say to put securitySchemes at top level, better to follow it even if it doesn't have consequences, no?

packages/core/src/server/server.ts
harijoe: Needed now because per-tool auth enforcement is new, and the legacy string-first overload used to early-return past all per-tool logic.

On `main`, `registerTool` branched like this:

```ts
if (typeof args[0] === "string") {
  baseFn.call(this, args[0], args[1], args[2]); // straight to the SDK
  return this;
}
// only the object-first path below extracts securitySchemes,
// mirrors _meta, registers the view, wraps the handler…
```

That early return was harmless before, since auth was enforced purely at the transport layer. But this branch adds two per-tool steps that have to run for every tool: `securitySchemesByTool.set(name, …)` (feeds the gate) and `decorateToolHandler(…, { securitySchemes })` (the handler backstop). The string overload skipped both, so a tool registered the legacy way was left completely unenforced.

Concretely, given:

```ts
server.registerTool(
  "checkout",
  { description: "Place an order", auth: { scopes: ["checkout"] } },
  handler,
);
```

On `main`'s code path this hits the early return: `securitySchemesByTool` never gets a `checkout` entry, so the gate's `schemesByTool.has("checkout")` is false and the call sails through, and the handler is never decorated. An anonymous or under-scoped caller reaches `handler`. The object-first form of the exact same tool was correctly gated. So auth depended on which overload you happened to use.

`normalizeRegisterToolArgs` collapses both overloads into one `{ config, cb }` shape at the top, so both run the identical registration + decorate path:

```ts
const { config, cb } = normalizeRegisterToolArgs(args);
// string-first and object-first now indistinguishable from here down
```

The alternative was duplicating the auth-registration into the string branch, but that just invites the two paths drifting apart again. This is the fix behind `21df39e`.

packages/core/src/server/server.ts
harijoe: It's just a way to not be affected by the onion of middlewares, as we already do it for attachViewUUID it felt okay to put it here.

docs/api-reference/register-tool.mdx
harijoe: I think it actually makes sense to use different words here : `oauth` is specific to configuring the oauth protocol (there could be others in the future) whereas `auth` on tools describe configuration about authotization in general, not really tight to oauth.

I can change it if you have a strong opinion on the matter but the change isn't obvious to me

packages/core/src/server/auth/setup.ts
harijoe: From Claude, it doesn't seem that the middleware is a suitable alternative way : 

>   It reads like it should be, but our `mcpMiddleware` can't do the job here — it's a layer-capability gap, not a style choice.
> 
>   `mcpMiddleware` wraps the SDK's `_requestHandlers`: it runs *inside* the SDK on already-parsed JSON-RPC, receives `(request, extra, next)`, and returns a JSON-RPC **result**. It has no handle on the Express `res`. This gate needs two things it
>   structurally can't give:
> 
>   1. **Transport-level HTTP output.** Every non-OpenAI client gets a real `401`/`403` + `WWW-Authenticate` header — that's how MCP OAuth discovery is bootstrapped. A middleware can only return a JSON-RPC body, never set the status or response
>   headers. Only the OpenAI in-band path (200 + `_meta["mcp/www_authenticate"]`) is expressible as middleware, and that's the minority case.
>   2. **Batch-level rejection.** This loop fails a whole JSON-RPC *array* at the transport boundary (the fail-closed batch behavior). `mcpMiddleware` runs per-method-handler *after* the SDK splits the batch, so "reject the entire batch" isn't
>   expressible there.
> 
>   So the gate lives at the Express layer by necessity. The handler-level backstop (`decorateToolHandler`) is the piece that *does* run in-band, and it's what catches anything that reaches a handler without passing this gate.
> 
>

examples/auth-descope-mixed/src/server.ts
fredericbarthelet: It's just that people may not see the possibility to do mixed auth if not advertise in auth examples themselves. Fine as is, but it might be worth reconsolidating example apps

docs/api-reference/register-tool.mdx
fredericbarthelet: Agree with your point, I was just afraid that people don't catch this in practice and typo one or the other

fredericbarthelet (approved):
LGTM :)

PR #939, opened 2026-07-08
feat(core): serve Agent Skills over MCP (SEP-2640)

harijoe:
Implements [SKY-485](https://linear.app/alpic/issue/SKY-485). Serves [Agent Skills](https://agentskills.io/) from a Skybridge server as `skill://` MCP resources per [SEP-2640](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2640). PRD: [Skills over MCP](https://app.notion.com/p/Skills-over-MCP-3951e5ce535b8037b314fc4570c97a33).

> **Experimental** while SEP-2640 is under review.

## Summary

- New `skills?: boolean | string` on `SkybridgeServerOptions` gates the feature (presence enables it, declares the `io.modelcontextprotocol/skills` capability). No new chainable API.
- All SEP-facing logic isolated in `packages/core/src/server/skills.ts`: frontmatter parse + validate, traversal-safe `skill://` URI resolution, discovery, `skill://index.json` generation, and the optional `resources/directory/read` method.
- Each `SKILL.md` and supporting file is served as a resource; the index carries `url` + `sha256` digest + verbatim frontmatter.
- Build emits `dist/skills.js`; the entry wrapper injects it in memory (never read from disk at runtime).
- Dev reads live from `src/skills/` so edits need no restart.
- Docs guide + API reference + `chatgpt-app-builder` skill updated.

## Cloudflare / Vercel note

Cloudflare Workers is a real server runtime here (no filesystem), and Vercel functions don't ship `dist/`. So skills are **not** copied to `dist/skills/` and read via `node:fs` — they are emitted as a JS module and injected in memory, mirroring the existing Vite-manifest pattern. Works on every target.

## Test plan

- `pnpm test` (core: unit + integration for discovery, validation, URI traversal, index shape, directory-read, and constructor wiring) and `pnpm build` pass.

## Manual verification

Tested a fresh consumer repo depending on the published dev build of this branch (`skybridge@0.0.0-dev.006848a`), with one skill at `src/skills/greeter/SKILL.md` and a server using `{ skills: true }`.

**Production (`skybridge build` → `start`)** — `build` emits `dist/skills.js` (frontmatter + `sha256` digest) and the entry wrapper primes it in memory. The running server advertises the `io.modelcontextprotocol/skills` capability and serves `skill://index.json` and `skill://greeter/SKILL.md` over MCP. ✅

**Dev (`skybridge dev`) + auto-reload** — server advertises the capability and serves the skill; editing `SKILL.md` triggers a nodemon restart (~2s) and the next read returns the updated content. ✅


docs/guides/skills.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+That is all. Skybridge validates every skill at build and dev startup (a bad `name`, missing `description`, or name/directory mismatch fails loudly), then exposes:
+
+- `skill://refunds/SKILL.md` and each supporting file as MCP resources.
+- `skill://index.json`, the discovery index (each entry carries the skill's `url`, a `sha256` digest, and its frontmatter verbatim).
+- the `io.modelcontextprotocol/skills` capability, so hosts know to look.
+
+## How hosts use them
+
+Skills are **pull-based**: the server never pushes them. A host reads `skill://index.json`, surfaces each skill's `name` and `description`, and lets the model read the full `SKILL.md` only when it's relevant ([progressive disclosure](https://agentskills.io/specification#progressive-disclosure)). To point the model at a skill, reference its URI from a tool description or a tool result, e.g. *"see `skill://refunds/SKILL.md`"*.
```
thleonard, line 56: Should we add warning saying that most hosts don't support this yet? The best would be a table of compatibility but it's hard to maintain

packages/core/src/cli/use-nodemon.ts
```diff
@@ -35,7 +35,8 @@ export function startNodemon(
       }
     : {
         watch: ["src"],
-        ext: "ts,json",
+        // `md` so editing a skill's SKILL.md restarts and re-discovers it.
```
thleonard: Remove the comment

packages/core/src/server/server.ts
```diff
@@ -483,6 +496,36 @@ export function __setBuildManifest(
...
+let pendingSkillsManifest: SkillsManifest | null = null;
+
+/** @internal Primes the build-time skills snapshot; called by `dist/__entry.js`. */
+export function __setSkillsManifest(manifest: SkillsManifest): void {
+  pendingSkillsManifest = manifest;
+}
+
+// Pure and `this`-free so it can run inside the `super(...)` call, before `this`
+// exists — the capability must be present for the `initialize` response.
+function withSkillsCapability(
```
thleonard, line 507: I think we can remove this comment also?

packages/core/src/server/server.ts
```diff
@@ -526,9 +569,13 @@ export class McpServer<
...
   ) {
-    super(serverInfo, options);
+    const mergedOptions = withSkillsCapability(options, skybridgeOptions);
+    super(serverInfo, mergedOptions);
     this.serverInfo = serverInfo;
-    this.serverOptions = options;
+    // Store the merged options: the per-request stateless server is built from
+    // these, so it must carry the skills capability too (a host keys off
+    // `capabilities.extensions` to decide whether to look for skills).
+    this.serverOptions = mergedOptions;
```
thleonard, line 572: We can't achieve `this.serverOptions = mergedOptions;` inside `super`?

packages/core/src/server/skills.ts
```diff
@@ -0,0 +1,311 @@
...
+const mimeTypeForFile = (name: string): string => {
+  switch (extname(name)) {
+    case ".md":
+      return "text/markdown";
+    case ".json":
+      return "application/json";
+    default:
+      return "text/plain";
+  }
+};
```
thleonard: We can write skills in any file format?

packages/core/src/server/skills.ts
```diff
@@ -0,0 +1,311 @@
...
+  try {
+    data = parseYaml(match[1] ?? "");
+  } catch (error) {
+    const detail = error instanceof Error ? error.message : String(error);
+    throw new Error(`Cannot parse frontmatter of ${source}: ${detail}`);
+  }
+  if (data === null || typeof data !== "object" || Array.isArray(data)) {
+    throw new Error(`Frontmatter of ${source} must be a YAML mapping`);
+  }
+  return data as Record<string, unknown>;
```
thleonard: What's the actual type of `data`?

packages/core/src/server/skills.ts
```diff
@@ -0,0 +1,311 @@
...
+
+const readSkillDir = (root: string, rel = ""): Record<string, SkillFile> => {
+  const files: Record<string, SkillFile> = {};
+  for (const entry of readdirSync(join(root, rel), { withFileTypes: true })) {
+    const childRel = rel ? `${rel}/${entry.name}` : entry.name;
+    // Skip symlinks so a link can't pull content from outside the skills tree
+    // into the served manifest.
+    if (entry.isSymbolicLink()) {
+      continue;
+    }
```
thleonard: Should this be specified in the doc?

packages/core/src/server/skills.ts
```diff
@@ -0,0 +1,311 @@
...
+    const skillMd = readFileSync(skillMdPath, "utf8");
+    const source = `${entry.name}/SKILL.md`;
+    const raw = parseFrontmatter(skillMd, source);
+
+    const parsed = SkillFrontmatterSchema.safeParse(raw);
+    if (!parsed.success) {
+      throw new Error(
+        `Invalid skill frontmatter in ${source}: ${z.prettifyError(parsed.error)}`,
+      );
+    }
```
thleonard, line 104: Can we throw later (after the loop) so that we output all the errors at once instead of having the user fix them one by one?

packages/core/src/server/skills.ts
```diff
@@ -0,0 +1,311 @@
...
+  name: string;
+  relPath: string;
+} {
+  const match = /^skill:\/\/([^/]+)(?:\/(.*))?$/.exec(uri.replace(/\/+$/, ""));
+  if (!match) {
+    throw new McpError(ErrorCode.InvalidParams, `Invalid skill uri: ${uri}`);
+  }
+  const [, name = "", relPath = ""] = match;
+  const segments = relPath.split("/").filter(Boolean);
+  if (name === "" || segments.includes("..") || segments.includes(".")) {
```
thleonard, line 131: You can have directories with dots inside the name so it will fail for them too, but I think it's the easiest way escape this indeed. So I guess it's fine

packages/core/src/server/skills.ts
```diff
@@ -0,0 +1,311 @@
...
+  }
+  return { name, relPath: segments.join("/") };
+}
+
+// Lists the immediate children of `relPath` within a skill. `null` means the
+// path is absent (empty `relPath` always matches the skill root).
+function listDir(
+  skill: Skill,
+  relPath: string,
+): { name: string; mimeType: string }[] | null {
```
thleonard, line 140: There's no utilities that explore a directory and provides an object representing it? I feel like we develop quite a lot

thleonard (changes requested):

packages/core/src/server/server.ts
harijoe: We can't use `this` before `super` returns, I don't see how we can simplify this, any idea?

packages/core/src/server/skills.ts
harijoe: It's accepted by the SEP but I agree that for now we could simplify and only accept markdown

packages/core/src/server/skills.ts
harijoe: I simplified to remove the cast string. It's a Record<string, unknown> because that's how [frontmatter](https://docs.github.com/fr/contributing/writing-for-github-docs/using-yaml-frontmatter) is designed (key-value)

packages/core/src/server/skills.ts
harijoe: Not sure it's worth it as it's likely that skills will be added one at a time and it respects the fail-fast principle

packages/core/src/server/skills.ts
harijoe: No strict equivalent unfortunately. I guess it's fine as the code is within skills.ts so no contamination to the rest of the codebase and it's well tested

packages/core/src/server/skills.ts
thleonard: Ok fair 👍

thleonard (approved):

PR #971, opened 2026-07-15
feat: add render-carousel tool

paulleseute:
- render-carousel tool: takes the curated product IDs and returns the products to show. Full data (variants, options, media) goes in _meta for the view; a trimmed grounding subset goes to structuredContent for the model. getProducts() is a stub you plug your catalog into.
- shared product model in types.ts (Price, Attribute), used by both tools. Product = one carousel card, its variants, and the option axes; variants are sparse so "this combo doesn't exist" just means it's not in the list.
- search-products output cleaned up to a flat product shape (id, title, price, outOfStock, attributes) instead of the old name/value properties bag.
- skill's ecommerce reference updated: documents both tools, the shared types, and a "mapping ids to products" section walking through the strategies (no variants / grouped / one card per variant) since that's the main thing a dev has to decide.

### DIFF
--- a/packages/create-skybridge/templates/ecom/src/server.ts
+++ b/packages/create-skybridge/templates/ecom/src/server.ts
@@ -1,6 +1,10 @@
 import { existsSync } from "node:fs";
 import { McpServer } from "skybridge/server";
-import { MIN_SEARCH_ITERATIONS } from "./config.js";
+import { CAROUSEL_RANGE, MIN_SEARCH_ITERATIONS } from "./config.js";
+import {
+  renderCarouselDefinition,
+  renderCarouselHandler,
+} from "./tools/render-carousel.js";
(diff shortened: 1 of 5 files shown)


neil-ac (approved):

neil-ac (commented):
Do you plan on adding data-llm in later commits ? It's a good practice to sync what the user sees - when they scroll the carousel, or click on a given item - and what the LLM knows about it

 ->+ would love to see the View of the render-carousel tool :)

paulleseute:
> Do you plan on adding data-llm in later commits ? It's a good practice to sync what the user sees - when they scroll the carousel, or click on a given item - and what the LLM knows about it
> 
> * would love to see the View of the render-carousel tool :)

yes, this pr scope the render server tool

PR #979, opened 2026-07-17
feat: add claude to conformance test

paulleseute:
What we fixed: the ChatGPT run never matched its own hand-checked baseline. openExternal and the follow-up message always read unsupported, because the old Notte function poked the widget over postMessage, which carries no user gesture, and both hosts only fire those effects on a real click. The modal test trusted a broadcast the Apps SDK sends even when nothing renders, so it logged a phantom verdict. And useViewState flaked: the runner mirrors results into the same host store the test writes to, and the two writers overwrote each other.

The new shape: one plain Playwright script (notte/conformance.py) does everything the Notte cloud function and the TS runner used to, for both hosts. Pick --host chatgpt|claude and --mode local|notte (a local Chrome profile, or a Notte cloud browser over CDP for CI). It clicks the widget's real buttons through the cross-origin iframes and only falls back to postMessage when it can't reach one, so gesture-gated hooks actually fire. Per-host adapters accept the native permission dialogs, the run retries in a fresh conversation when a test stalls or the widget fails to boot, and every verdict is gated against a per-host baseline. CI is now one matrix job per host; the old function and TS harness are gone.

A few app fixes rode along to keep the run honest: the runner stays mounted so a stuck modal can't brick it, restart resets in place, modal_open is only claimed when a modal truly renders, resize runs before fullscreen, the CSP redirect allowlist is dropped so ChatGPT shows its real "External site" modal (which is what lets us verify openExternal), and the viewState probe retries past the mirror clobber.

Claude support: Claude had no automation before; now it runs the same driver with its own baseline. Its adapter types into the ProseMirror composer, accepts the "Open link" and "Download" permission dialogs, and reads the in-widget modal.

Validation: a bunch Notte runs, all matching.


e2e/conformance/src/views/conformance.tsx
```diff
@@ -152,9 +152,16 @@ function Runner({ onRestart }: { onRestart: () => void }) {
...
+  // updater, resurrecting the old run on the next mount).
   const restart = () => {
+    recordedRef.current.clear();
+    setStep(0);
+    setRows(TESTS.map(() => null));
+    setArmed(false);
+    setConfirming(null);
     setViewState((prev) => ({ ...prev, run: undefined }));
-    onRestart();
   };
```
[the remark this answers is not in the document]
paulleseute: Leaving this one. The remote-drive protocol has no restart action (it's run/skip/yes/no/close-modal/restore-inline), so the driver never restarts mid-run; on a stall it starts a fresh conversation instead. Restart is a human-only button, and the worst case there is a cosmetic stale row that the next run overwrites. Making the button itself airtight would mean stamping each run with an epoch and having recordAt drop results from an older epoch, but since it's unreachable from the conformance runs it's not worth the extra state here.

examples/conformance/notte/conformance.py
```diff
@@ -0,0 +1,1046 @@
...
+# (its ChatGPT new tab never surfaced over Notte's CDP), but both hosts now pop
+# an observable "Open link" dialog that accept_native_dialog accepts, so it's
+# verifiable everywhere. Kept as the seam for any hook that becomes unobservable.
+UNVERIFIABLE_BY_MODE: dict[str, frozenset[str]] = {}
+
+
+# ── CLI ───────────────────────────────────────────────────────────────
+
+
+def load_dotenv(path: Path) -> None:
```
qchuchu: Couldn't you use dotenv package ?

examples/conformance/notte/conformance.py
```diff
@@ -0,0 +1,1046 @@
+# /// script
```
qchuchu: Could you separate into 4 files : 

- chatgpt
- claude
- utils (all the utils functions that the runner use)
- conformance.py : run_conformance + the main function

It will be easier to maintain and understand the dependencies. 

But I don't know if it matches notte setup

qchuchu (approved):
OK for me with a few changes.

Might be worth moving it to a separate folder (e2e ?) but maybe in a separate PR for ease of read for this one.

PR #984, opened 2026-07-20
feat(docs): add webmcp devtools setup

paulleseute:
and fixes the skills over mcp one to match house style: icon, unequivocal title, format (em dashes), tone of voice, not discussing internals, go further cards.

### DIFF
--- a/docs/guides/skills.mdx
+++ b/docs/guides/skills.mdx
@@ -1,17 +1,17 @@
 ---
 title: "Serve Skills over MCP"
 description: "Ship Agent Skills alongside your server"
-sidebarTitle: "Skills"
-icon: "book-sparkles"
+sidebarTitle: "Skills Over MCP"
+icon: "book-open-text"
 ---
 
 <Info>
-Skills over MCP tracks [SEP-2640](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2640), which is still under review. This feature is **experimental** and may change with the spec. Most hosts do not yet support it, so skills you serve today may go unused until they adopt the standard.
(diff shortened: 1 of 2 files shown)


docs/guides/skills.mdx
```diff
@@ -43,23 +43,29 @@ description: Process customer refund requests per company policy.
...
+Skills are markdown files only: `SKILL.md` and any supporting `.md` files are served as `text/markdown`, other formats and symlinks are ignored.
+</Info>
 
-Skills are **pull-based**: the server never pushes them. A host reads `skill://index.json`, surfaces each skill's `name` and `description`, and lets the model read the full `SKILL.md` only when it's relevant ([progressive disclosure](https://agentskills.io/specification#progressive-disclosure)). To point the model at a skill, reference its URI from a tool description or a tool result, e.g. *"see `skill://refunds/SKILL.md`"*.
+## How Hosts Use Skills
 
-<Warning>
-Skill content is untrusted model input, not a trusted instruction. Hosts treat it as a prompt-injection surface and never grant it higher authority or execute code from it.
-</Warning>
+Hosts don't inject skills on their own. To put one in front of the model, reference its URI from a tool description or a tool result, for example *"see `skill://refunds/SKILL.md`"*. The host pulls in the full instructions only when they're relevant.
```
fredericbarthelet, line 54: As stated in the SEP, it's still required that host have tools enabling them from downloading a resource. Otherwise resources are application-driven, meaning the model can't choose to invoke and get the full content autonomously.

I don't think nudging the model to get the skill in a tool description or a tool response will work unless the model has access to host-defined tools to do so (like we implemented in the playground)

fredericbarthelet (approved):
Good to me for the devtools part
As for skill over MCP, referencing skill in tool desc/response has no effect on hosts that don't support skills over mcp SEP in its current form

docs/guides/skills.mdx
paulleseute: yes this assumes the host implement the actual standard, otherwise the whole doc is void. we have a callout of the beginning of the pages stating it's far from being widely supported

PR #988, opened 2026-07-21
fix(cli): report a port fallback when PORT is invalid and 3000 is taken

eeshsaxena:
`resolvePort` hard-codes `fallback: false` on the invalid-`PORT` branch, even though that branch calls `detectAvailablePort` and can land on a different port.

## Problem

```ts
if (rawEnv) {
  const parsed = Number(rawEnv)
  if (Number.isInteger(parsed) && parsed > 0) { ... }
  return {
    port: await detectAvailablePort(DEFAULT_PORT),
    fallback: false,   // <- even if 3000 was taken
    envWarning: `Invalid PORT="${rawEnv}", ...`,
  }
}
```

The CLI uses `fallback` to choose the message it prints:

- `commands/dev.tsx` — `"3000 in use, running on "` vs `"Running on "`
- `cli/run-plain.ts` — same distinction

So when `PORT` is invalid **and** 3000 is already taken, the server correctly moves to 3001 but the user is told plain `"Running on ..."` and never learns the default port was busy. The branch below it already computes this correctly (`fallback: port !== DEFAULT_PORT`).

## Fix

Compute the flag the same way the default branch does.

## Tests

`detect-port.ts` had no tests; added `detect-port.test.ts` covering the `--port` flag, a valid `PORT`, a taken default, an invalid `PORT`, and the regression (invalid `PORT` + taken default). I verified the regression test fails against the previous code and passes with the fix. `node:net` is mocked so port availability is deterministic.

Biome is clean. Note: the 3 failures in `build-steps.test.ts` on my checkout are pre-existing and reproduce on a clean tree (an artifact of installing with `--ignore-scripts`), unrelated to this change.

### DIFF
--- a/packages/core/src/cli/detect-port.test.ts
+++ b/packages/core/src/cli/detect-port.test.ts
@@ -1,0 +1,95 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const { busyPorts } = vi.hoisted(() => ({ busyPorts: new Set<number>() }));
+
+vi.mock("node:net", () => {
+  const createServer = () => {
+    const handlers: Record<string, () => void> = {};
+    const server = {
+      once(event: string, cb: () => void) {
+        handlers[event] = cb;
+        return server;
+      },
+      listen(port: number) {
+        queueMicrotask(() => {
+          if (busyPorts.has(port)) {
+            handlers.error?.();
+          } else {
+            handlers.listening?.();
+          }
+        });
+        return server;
+      },
+      close(cb?: () => void) {
+        cb?.();
+        return server;
+      },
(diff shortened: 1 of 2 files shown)


harijoe:
Hi @eeshsaxena , thanks for your contribution!

Since this is your first time contributing to the Skybridge repository, would you mind introducing yourself on [our Discord](https://discord.gg/awV4gu74wK)? We'd love to get to know our contributors.

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #995, opened 2026-07-22
Add OpenUI exemple in documentation

fredericbarthelet:


docs/docs.json
```diff
@@ -214,9 +214,15 @@
             ]
           },
           {
-            "group": "UI and Component libraries",
+            "group": "Generative UI",
```
harijoe: Isn't JSON renderer the adopted term instead of generative UI?

PR #1005, opened 2026-07-27
fix(build): allow extending esbuild externals via skybridge({ serverExternal })

harijoe:
[Slack discussion](https://alpicworkspace.slack.com/archives/C0B2D771DQV/p1785149256035349)

## Summary

- `skybridge build` bundles the server with esbuild, so every reachable dependency must resolve, even ones the server never runs. The external list was hardcoded to `["vite", "@skybridge/devtools"]` with no way to extend it, and the Vercel step is unconditional, so it can't be skipped either.
- Adds a `serverExternal` option to the `skybridge` Vite plugin, exposed via `api` and read by `skybridge build`, same wiring as the existing `viewsDir`.
- Sets `loader: { ".node": "file" }` on the esbuild call. Native addons can't be inlined, so today any native dependency in the tree fails the build and no `external` entry helps.
- `resolveViewsDir` becomes `resolveSkybridgeConfig`, returning the whole plugin api instead of a single field.

```ts
// vite.config.ts
skybridge({ serverExternal: ["dtrace-provider"] })
```

Reported by a user migrating to 1.3.0: bunyan does `require('dtrace-provider' + '')` in a `try/catch`, esbuild constant-folds the concatenation and fails on the native binding. Present since at least v1.1.0.

## Architecture Notes

`vite.config.ts` is the escape hatch's home rather than an env var or a new `package.json` field, because it is already the config file the build loads. Auto-externalizing anything esbuild can't resolve was rejected: it would turn a genuinely missing dependency into a runtime crash instead of a build error.

Docs updated in `api-reference/cli.mdx` under `build`.

### DIFF
--- a/docs/api-reference/cli.mdx
+++ b/docs/api-reference/cli.mdx
@@ -89,6 +89,26 @@ deno run -A npm:skybridge/skybridge build
 
 The output lands in `dist/`, ready for the `deploy` script or `skybridge start`.
 
+### Excluding packages from the server bundle
+
+The build's last step bundles your server into a single deployable function with esbuild. Bundling means every reachable dependency has to be resolved, including ones your code never runs.
+
+Some packages break that step. A common case is a logger with an optional native dependency, like [bunyan](https://github.com/trentm/node-bunyan), which does `require('dtrace-provider')` inside a `try/catch`. You never use DTrace, but esbuild still tries to resolve it and fails on the native binding.
+
+List those packages in the Vite plugin's `serverExternal` to leave them out of the bundle:
+
+```ts vite.config.ts
+import { defineConfig } from "vite";
+import react from "@vitejs/plugin-react";
+import { skybridge } from "skybridge/vite";
+
+export default defineConfig({
+  plugins: [react(), skybridge({ serverExternal: ["dtrace-provider"] })],
+});
+```
+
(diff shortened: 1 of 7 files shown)


docs/api-reference/cli.mdx
```diff
@@ -89,6 +89,26 @@ deno run -A npm:skybridge/skybridge build
 
 The output lands in `dist/`, ready for the `deploy` script or `skybridge start`.
 
+### Excluding packages from the server bundle
+
+The build's last step bundles your server into a single deployable function with esbuild. Bundling means every reachable dependency has to be resolved, including ones your code never runs.
+
+Some packages break that step. A common case is a logger with an optional native dependency, like [bunyan](https://github.com/trentm/node-bunyan), which does `require('dtrace-provider')` inside a `try/catch`. You never use DTrace, but esbuild still tries to resolve it and fails on the native binding.
```
paulleseute, line 96: i think this should be reference in troubleshooting as well. also this seems oddly specific, is it a common case or the only case we had ?

paulleseute (approved):

docs/api-reference/cli.mdx
harijoe: Only case we had so far, it's quite specific indeed. Also now that the API allows it, an LLM is likely to be able to solve it by itself I think.

Issue #1003, opened 2026-07-27
State update error when comibing useRegisterViewTool and useViewState


eric-burel:
Hi,

Using Skybridge 1.3.0 from a blank template, I am trying to craft a view tool that triggers a state update. My goal is to have a persistent component whose state can be updated by the LLM, for instance to track progress visually or similar use cases.

Minimal repro here : https://github.com/eric-burel/skybridge-repro-state-update-from-tool

Using a conversation, ask the model to show the title, and then to update it.
The state update seems to lead to a bad setState with the following warning:

```js
[console.error] Cannot update a 
    component (`Title`) while rendering a different component 
    (`Title`). To locate the bad setState() call inside `Title`, follow the stack trace as described in 
    https://react.dev/link/setstate-in-render
```

In this minimal repro the title is actually updated, but in my previous attempt in a non-trivial component the title wouldn't update at all. I don't have the warning all the time. So it feels like a race condition somewhere maybe.

Update: I've added a non-minimal repro from the Onboarding component of the starter, visible here: https://github.com/eric-burel/repro-skybridge-non-minimal

I have the same warning but also the component display doesn't update.

Other minor nits:
- Can we test useRegisterViewTool lifecycle from playground without using the LLM?
- I struggled to get proper typings, not sure what triggers typing in generateHelpers

harijoe:
Thanks @eric-burel for raising this issue. It should have been fixed by #1006, it will ship in v1.4, feel free to reopen after that if you still face the issue.

PR #1006, opened 2026-07-27
fix(web): don't call adaptor.setViewState inside the React state updater

harijoe:
## Summary

- `useViewState`'s setter called `adaptor.setViewState` from inside the `_setViewState` updater. React runs updaters during render, so the host write happened mid-render.
- Symptoms: `Cannot update a component (X) while rendering a different component (X)` on every `setViewState`, and two `setWidgetState` calls per update under StrictMode (which `mountView` always applies).
- Updater is now pure; the host write happens in the caller. A ref carries the latest state so function updaters still see fresh values.
- Added one test: `setWidgetState` is called once per update under StrictMode. Fails on `main` (2 calls), passes here.

Reported: https://discord.com/channels/1370323135861358713/1530232765058912377

`createStore` has a similar shape but is unaffected: its subscriber runs on `set()`, outside render.

### DIFF
--- a/packages/core/src/web/hooks/use-view-state.ts
+++ b/packages/core/src/web/hooks/use-view-state.ts
@@ -52,8 +52,15 @@ export function useViewState<T extends UnknownObject>(
       : (defaultState ?? null);
   });
 
-  const viewStateRef = useRef(viewState);
-  viewStateRef.current = viewState;
+  const mountedStateRef = useRef<T | null | undefined>(viewState);
+
+  useEffect(() => {
+    mountedStateRef.current = viewState;
+
+    return () => {
+      mountedStateRef.current = undefined;
+    };
+  }, [viewState]);
 
   useEffect(() => {
     if (viewStateFromBridge !== null) {
@@ -63,12 +70,17 @@ export function useViewState<T extends UnknownObject>(
 
   const setViewState = useCallback(
     (state: SetStateAction<T | null>) => {
-      const newState =
-        typeof state === "function" ? state(viewStateRef.current) : state;
+      const prevState = mountedStateRef.current;
+
+      if (prevState === undefined) {
+        return;
+      }
+
+      const newState = typeof state === "function" ? state(prevState) : state;
       const stateToSet = injectViewContext(newState);
       const filteredState = filterViewContext(stateToSet);
 
-      viewStateRef.current = filteredState;
+      mountedStateRef.current = filteredState;
       _setViewState(filteredState);
 
       if (stateToSet !== null) {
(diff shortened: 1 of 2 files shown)


packages/core/src/web/hooks/use-view-state.ts
```diff
@@ -54,16 +72,21 @@ export function useViewState<T extends UnknownObject>(
...
+        return;
+      }
+
+      const newState =
+        typeof state === "function" ? state(viewStateRef.current) : state;
+      const stateToSet = injectViewContext(newState);
+      const filteredState = filterViewContext(stateToSet);
 
-        if (stateToSet !== null) {
-          adaptor.setViewState(stateToSet);
```
fredericbarthelet, line 62: wouldn't it be easier instead to add an effect on `viewState` in order to call `adaptor.setViewState` on change (instead of doing it in line in the exposed `setViewState` API)?

Using `ref` means wanting to opt out of React lifecycle altogether. It feels counterproductive to use both `useState` and `ref` to create stable reference across render

harijoe: I made the change, but we still need to retain the last value somehow and using ref for that makes sense I think?

packages/core/src/web/hooks/use-view-state.ts
```diff
@@ -46,27 +47,31 @@ export function useViewState<T extends UnknownObject>(
...
+      return;
+    }
+
+    const stateToSet = injectViewContext(viewState);
 
-        if (stateToSet !== null) {
-          adaptor.setViewState(stateToSet);
-        }
+    if (stateToSet === null) {
+      return;
```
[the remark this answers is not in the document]
fredericbarthelet: WDYT @harijoe ?

fredericbarthelet (approved):

packages/core/src/web/hooks/use-view-state.ts
```diff
@@ -54,16 +70,22 @@ export function useViewState<T extends UnknownObject>(
...
-        const stateToSet = injectViewContext(newState);
+      const prevState = mountedStateRef.current;
+
+      if (prevState === undefined) {
+        return;
+      }
+
+      const newState = typeof state === "function" ? state(prevState) : state;
+      const stateToSet = injectViewContext(newState);
+      const filteredState = filterViewContext(stateToSet);
```
fredericbarthelet: filter and inject are both side of the bijection `dequal(newState, filterViewContext(injectViewContext(newState))) === true` no?

harijoe:
Closes #1003

PR #1008, opened 2026-07-28
feat!: migrate to @modelcontextprotocol/sdk v2 (MCP 2026-07-28)

harijoe:
> [!NOTE]
> Docs, skills and the migration guide are intentionally out of scope for this PR. They land in a follow-up before the 2.0 release; until then, docs code samples do not compile against this branch.

### Summary

- Migrates skybridge from `@modelcontextprotocol/sdk` v1 to the v2 packages, stable `2.0.0`.
- One `/mcp` endpoint serves both 2025-era hosts and the 2026-07-28 revision.
- Every HTTP request is served by a fresh server instance, rebuilt from a registration ledger.
- Devtools client moved to v2; browser OAuth hardened (`state` validation, discovery persistence, RFC 9207 `iss`).
- Lands the breaking changes decided in the V2 PRD, including the hooks regroup.
- Examples and templates depend only on `skybridge`; their direct sdk deps are gone.
- Pushes to `v2` publish `2.0.0-alpha.<sha>` on the npm `alpha` tag.

### Breaking changes

Server:

- Tool handler `extra` is v2's `ServerContext`:
  - `extra.http?.authInfo` (was `extra.authInfo`)
  - `extra.mcpReq._meta` (was `extra._meta`, carries the typed ChatGPT client hints)
  - `extra.mcpReq.{signal,id,notify,send}` (were top-level)
- `InvalidTokenError` removed: throw `OAuthError("invalid_token", ...)`.
- oauth `verify` option removed: `verifier: createJwksVerifier({ ... })` is the only path.
- Legacy `registerTool(name, config, handler)` overload removed.
- `KnownToolMeta.securitySchemes` removed: declare via top-level config or `auth`.
- `ViewConfig.hosts` removed; `ViewHostType` narrowed to `"mcp-app"` (the `"apps-sdk"` value was a no-op label) and no longer exported.
- `registerResource` / `registerPrompt` signatures now owned by skybridge (typed wrappers can land in a minor).
- `connectStatelessTransport(transport)` renamed `createStatelessServerInstance()`.
- sdk v1 is no longer a peerDependency: import MCP types from skybridge re-exports (`RawInputShape`, `InferSchemaOutput`, `McpExtra`, ...).

Web:

- `useLayout` removed: `theme` moved into `useUser`, `maxHeight`/`safeArea` into the new `useViewport`.
- `useHostInfo` renamed `useHost`, same shape.
- `useDownload` returns the function bare.
- `useToolInfo` idle state removed.
- View tool `inputSchema` is typed zod-only (`ViewToolInputShape`): the bridge validates with `z.object()` at runtime.

### Architecture Notes

- **Registration ledger**: protocol era is instance state in v2, so sharing the singleton's handler maps pinned all traffic to the 2025 codec and leaked negotiated state across concurrent clients. Registrations are recorded once and replayed onto a fresh `McpServerBase` per request (measured ~35 µs per tool).
- **Ledger replays registrations, not mutations**: `registered.disable()` / raw `setRequestHandler` after startup do not reach per-request instances.
- **sdk v1 stays as a direct dependency of core**: `ext-apps` peer-requires it and its `App` runs on it. Exits when ext-apps ships on v2; only four web-bridge files still import it.
- **Legacy responses are SSE-framed**: v2 does not plumb `enableJsonResponse` through `createMcpHandler`'s legacy leg. Verified harmless on real workerd via `wrangler dev`; upstream issue drafted.
- **`WWW-Authenticate` sanitizer kept over the SDK's escaping**: RFC 6750 forbids the quotes the SDK escapes into the challenge.

### Alternatives Considered

- **Keep sharing the singleton's handler maps with per-request shells**: the SDK's tool wrappers read the negotiated era from the instance they were registered on. The singleton never negotiates, so `codecForVersion(undefined)` serialized 2026-era results with the 2025 codec (`structuredContent: {"result": 42}` instead of `42`), and `initialize` from one client overwrote another's negotiated state. Replaying registrations creates the wrappers on the instance the SDK actually stamps.
- **Register inside `createMcpHandler`'s factory (the SDK idiom)**: removes the ledger entirely, but also removes the module-scope singleton that `typeof server` type inference and the `server.run()` deployment contract hang on, changing the shape of every app file. The ledger is ~40 framework-internal lines to keep that contract intact.
- **Rebuild v1's `enableJsonResponse` workaround for the legacy leg**: the option existed to dodge a workerd bug where post-await chunked writes were dropped. Tested on real workerd (async tools, 300 KB post-await payloads, 2026-era negotiation): it no longer reproduces, so the workaround was retired instead of reimplemented against the SDK's API.

Tracked in SKY-514.


packages/core/package.json
fredericbarthelet, line 1: can't use peerDeps model anymore for dependency to the SDK ?

packages/core/package.json
```diff
@@ -59,7 +58,12 @@
   },
   "dependencies": {
     "@babel/core": "^7.29.7",
+    "@modelcontextprotocol/client": "2.0.0-beta.5",
```
fredericbarthelet: can be moved to dev deps, only used in `.test.ts` files

packages/core/package.json
harijoe: We can, but we now need to import 4 deps instead of just the sdk. It's a discussion to have but it makes sense on a DX perspective to not use peerDeps anymore.

fredericbarthelet:
> Note
> 
> Docs, skills and the migration guide are intentionally out of scope for this PR. They land in a follow-up before the 2.0 release; until then, docs code samples do not compile against this branch.
> 
> ### Summary
> ...
> * Pushes to `v2` publish `2.0.0-alpha.<sha>` on the npm `alpha` tag.

Please move from alpha to beta after this merge for subsequent docs, skills and migration publish before GA release.

packages/core/src/server/auth/setup.ts
```diff
@@ -1,8 +1,8 @@
-import type { BearerAuthMiddlewareOptions } from "@modelcontextprotocol/sdk/server/auth/middleware/bearerAuth.js";
 import {
+  type BearerAuthMiddlewareOptions,
   getOAuthProtectedResourceMetadataUrl,
```
fredericbarthelet, line 3: are you sure this is not in `@modelcontextprotocol/server` instead? I feel like some of the types/API are re-exported. How did you map from old sdk v1 to new split of v2 (any migration guide available on the SDK side ?)

packages/devtools/package.json
```diff
@@ -40,7 +40,7 @@
     "@alpic-ai/ui": "^1.155.0",
     "@base-ui/react": "^1.6.0",
     "@fontsource-variable/jetbrains-mono": "^5.3.0",
-    "@modelcontextprotocol/sdk": "^1.29.0",
+    "@modelcontextprotocol/client": "^2.0.0",
```
fredericbarthelet, line 43: are you sure this shouldn't be packaged as dependency ?
We were relying on the SDK installed in the project (since it was only a peer dep in v1), but now the `@modelcontextprotocol/client` only is required to run the devtools
It's cool because version of client and server don't need to be aligned anymore (and we therefore don't have a problem anymore with needing to go through a peer dep to avoid locking sdk minor/patch version in skybridge) and client can be packaged on its own by the devtools (there is no reason any project would list client as a dependency.


It could also be listed as a dev dependency if this package was requiring it as a peer dependency (that's another viable option)

packages/core/package.json
```diff
@@ -50,16 +50,20 @@
   "author": "Frédéric Barthelet",
   "license": "ISC",
   "peerDependencies": {
-    "@modelcontextprotocol/sdk": ">=1.27.0",
-    "@skybridge/devtools": "^1.1.0",
+    "@skybridge/devtools": "^1.1.0 || ^2.0.0-alpha.0",
```
fredericbarthelet:
```suggestion
    "@skybridge/devtools": "^2.0.0-alpha.0",
```
you still require backward compat with v1?

packages/core/src/server/auth/setup.ts
harijoe: `mcpAuthMetadataRouter` exists only in `@modelcontextprotocol/express` as it returns an Express Router, so it's coupled to that package. `getOAuthProtectedResourceMetadataUrl` and `requireBearerAuth` are exported from both, so importing all three from express keeps one import and is correct

I couldn't find any v1→v2 migration guide from the SDK, the mapping was done from the v2 packages' actual exports and by guessing + trying.

packages/devtools/package.json
harijoe: The thing is, the peer+devDep alternative would force every consumer to install a package they never import. Not ideal I think

packages/core/package.json
harijoe: You're right, applying the suggestion

packages/core/package.json
```diff
@@ -50,16 +50,20 @@
...
     "react-dom": ">=18.0.0",
     "vite": ">=7.3.1"
   },
   "dependencies": {
     "@babel/core": "^7.29.7",
+    "@modelcontextprotocol/core": "^2.0.0",
     "@modelcontextprotocol/ext-apps": "^1.7.4",
+    "@modelcontextprotocol/express": "^2.0.0",
+    "@modelcontextprotocol/node": "^2.0.0",
+    "@modelcontextprotocol/sdk": "^1.29.0",
```
fredericbarthelet, line 65:
```suggestion
```
should be completly removed, or are there still import statement from old sdk v1 ?

packages/core/package.json
```diff
@@ -50,16 +50,20 @@
...
+    "@modelcontextprotocol/server": "^2.0.0",
```
fredericbarthelet, line 66: i see an alternative here: keep server as a peer dep and let people bring their own version (with matching zod version), core only packages express (or node, but probably not both) to act as the framework (as if we were "just" another middleware just like hono, fastify and express)
WDYT?

packages/core/package.json
```diff
@@ -75,10 +79,11 @@
     "posthog-node": "^5.45.2",
     "superjson": "^2.2.6",
     "yaml": "^2.9.0",
+    "zod": "^4.4.3",
```
fredericbarthelet: why did zod move ?

packages/core/package.json
```diff
@@ -75,10 +79,11 @@
...
     "zustand": "^5.0.14"
   },
   "devDependencies": {
-    "@modelcontextprotocol/sdk": "^1.29.0",
+    "@modelcontextprotocol/client": "^2.0.0",
```
fredericbarthelet, line 84: this should remain the scope of devtools I think

packages/core/src/web/bridges/types.ts
```diff
@@ -1,15 +1,13 @@
-import type {
-  SchemaOutput,
-  ZodRawShapeCompat,
-} from "@modelcontextprotocol/sdk/server/zod-compat.js";
```
fredericbarthelet, line 4: no equivalent?

packages/core/src/web/hooks/use-call-tool.test.ts
```diff
@@ -24,7 +24,7 @@ describe("useCallTool - onSuccess callback", () => {
     McpAppBridge.resetInstance();
     AppsSdkBridge.resetInstance();
     vi.stubGlobal("parent", { postMessage: vi.fn() });
-    vi.stubGlobal("skybridge", { hostType: "apps-sdk" });
+    vi.stubGlobal("skybridge", { hostType: "mcp-app" });
```
fredericbarthelet: didn't we kill hostType altogether with a single entry bridge now ?

packages/core/src/server/auth/setup.ts
fredericbarthelet: found this: https://github.com/modelcontextprotocol/typescript-sdk/blob/main/docs/migration/upgrade-to-v2.md

packages/core/src/server/express.ts
```diff
@@ -1,6 +1,7 @@
 import type http from "node:http";
 import path from "node:path";
-import { StreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/streamableHttp.js";
+import { toNodeHandler } from "@modelcontextprotocol/node";
```
fredericbarthelet, line 3: I don't understand why you don't use the express middleware `@modelcontextprotocol/express` instead

packages/core/package.json
harijoe: ext-apps still depends on it unfortunately, they haven't shipped a v2 yet

packages/core/package.json
harijoe: It's actually a fix, we use already zod in skills.ts for example. It worked because zod was installed through @modelcontextprotocol/sdk

packages/core/package.json
harijoe: To discuss, I'm not sure. What putting server as a peer-deps means is that we don't control the version of the server that skybridge actually relies on, it feels a bit dangerous.

packages/core/src/web/bridges/types.ts
harijoe: We can bring it back. To discuss.

packages/core/package.json
harijoe: It's only a devDependency, only used by core's own integration tests, replacing the sdk v1 entry that did the same job

packages/core/src/server/express.ts
harijoe: As far as I understand it, `@modelcontextprotocol/express` isn't providing a middleware but just a way to create an express server with some guards already bundled. I'd say it's better if we keep relying on the express package itself for flexibility

harijoe:
@greptile-apps review

valentinbeggi (commented):
+ It's weird that serverFactory takes a server as input and supercharge it, I'd expect it to build a server instead. Maybe it's just a naming concern?

+ What are the SkybridgeServer type params exactly, why is it needed, can't it be inferred somehow? : e.g `SkybridgeServer<{ email?: string; subject?: string }>`

+ `import { type SkybridgeServer, auth0Provider, Skybridge } from "skybridge/server";` -> importing Skybridge from skybridge/server feels weird. I'd expect Skybridge name to be independent of backend/frontend (server/client). Maybe a better name would be `SkybridgeApp` ? Not 100% sure about this because express for example is Express not ExpressApp but it's only a server, there's no client 

+ It would maybe be worth to split app/server in 2 files in the examples ? Or have skybridge app in index.ts ? Or just rename server.ts to app.ts

harijoe:
@greptileai

fredericbarthelet (approved):
LGTM, what an awesome work 🤯 !
Great job !

PR #1010, opened 2026-07-28
Fix/devtools preview host surface

### LINKED ISSUE #841: devtools: preview doesn't render the host surface background behind the iframe
The devtools preview doesn't reproduce the background a real host (ChatGPT, Claude…) paints behind the view's iframe, so the rendered result doesn't match what ships. Two symptoms:

**Fullscreen** — the background shows as white regardless of the view's own background. E.g. a view with a lavender (light) background renders white in fullscreen.

**Inline / PiP** — the background is always effectively transparent. Switching the simulated theme to *Dark* updates the view's CSS, but the area behind stays transparent, so white text/buttons sit on the devtools' grey panel and become unreadable.

**Impact:** hard to iterate on UI and judge how a view will look inside the final host.

**Likely area:** the preview container in `packages/devtools/src/components/layout/tool-panel/view/index.tsx` — fullscreen forces `bg-background` (devtools chrome) behind the iframe, while inline/PiP applies no surface color at all. The container should paint the simulated host surface for the current theme behind the iframe in all display modes.

Boj4ckk:
## Why
Closes #841

The dark mode toggle in devtools had no visual effect because
index.css was not using the dark-theme CSS variables.

## What
- Applied dark-theme CSS variables for background and grid colors
  in the preview component
- Background color is now driven by `--preview-surface` which
  resolves to the correct `--color-background` value based on
  the active theme

The dark-theme colors were chosen without existing specs — happy
to adjust if needed.

## Testing
- Verified dark mode works across all screen modes (full, pip, inline)
- Added test case for grid color correctness in dark mode

## Screenshots
Before → After

chess_before

chess_after

### DIFF
--- a/packages/devtools/e2e/tests/preview-surface.spec.ts
+++ b/packages/devtools/e2e/tests/preview-surface.spec.ts
@@ -1,0 +1,92 @@
+import { expect, type Locator, test } from "@playwright/test";
+
+// Issue #841: the preview surface (the area behind the rendered view) must
+// darken with the dark theme in every display mode, and the grid must stay
+// visible. Colours come from the centralized @theme tokens in index.css:
+//   light: --color-background #ffffff → rgb(255, 255, 255)
+//   dark:  --color-background #071718 → rgb(7, 23, 24)
+const LIGHT_SURFACE = "rgb(255, 255, 255)";
+const DARK_SURFACE = "rgb(7, 23, 24)";
+const LIGHT_GRID = "#e3eaea";
+const DARK_GRID = "#1f4449";
+
+test.describe("preview surface (issue #841)", () => {
+  // The preview container: the div wrapping the view iframe, whose background
+  // is `var(--preview-surface)`. Selected via its iframe child so it stays
+  // stable regardless of layout classes.
+  let container: Locator;
+  // The scrollable region behind the container that carries `data-theme` and
+  // the grid CSS variables.
+  let previewRegion: Locator;
+  // The view-options toolbar form holding the display-mode radios and the
+  // theme checkbox (all sr-only inputs wrapped in visible labels).
+  let toolbar: Locator;
+
+  test.beforeEach(async ({ page }) => {
+    await page.goto("/");
+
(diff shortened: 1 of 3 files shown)


harijoe:
Hi @Boj4ckk, thanks for your contribution!

As the contributor guidelines state, we require that either Greptile gives your PR a 5/5, or that you respond to the issues it raised explaining why you disagree with its review.

Also, since this is your first time contributing to the Skybridge repository, would you mind introducing yourself on [our Discord](https://discord.gg/awV4gu74wK)? We'd love to get to know our contributors.

Boj4ckk:
> Hi @Boj4ckk, thanks for your contribution!
> 
> As the contributor guidelines state, we require that either Greptile gives your PR a 5/5, or that you respond to the issues it raised explaining why you disagree with its review.
> 
> Also, since this is your first time contributing to the Skybridge repository, would you mind introducing yourself on [our Discord](https://discord.gg/awV4gu74wK)? We'd love to get to know our contributors.

Hey! Yes, I'm on it, thanks. I introduced myself in the #introductions Discord channel, btw

Boj4ckk:
While manually testing the fix on the @examples/chess app, I noticed a behavior I'm unsure about from a product perspective. Wanted to flag it here.

Observed behavior
- Toggling dark mode from the devtools preview toolbar correctly updates both the preview surface and the chess board theme.
- However, the chess board has its own internal dark/light toggle.
- Toggling the chess board's internal toggle updates the board locally, but it does not sync back to the devtools preview toggle.

image


Question
Is this the intended behavior? My understanding is that in a real host (ChatGPT/Claude) the host owns the theme and pushes it down to the view the view should not override the host's theme toggle. But I wanted to confirm this is the expected product behavior for devtools as well.

For context, my change in this PR only touches the preview region's CSS attribute (data-theme → data-preview-theme) to isolate the preview surface from the devtools UI chrome. I did not modify useSyncOpenaiTheme or the host→view theme propagation. So this behavior likely predates my change, but I wanted to make sure it's acknowledged.

PR #1011, opened 2026-07-29
feat(web): add useHostInfo hook to identify the rendering host

### LINKED ISSUE #909: View API to get current host it's being rendered in
## Summary

Provide a simple API or new property in existing API to best guess the current host the view is being rendered in.

## Sourcing strategy

There should be some priority in order not to get false positive (`window.openai` exists in more than just ChatGPT host). Most of the hosts can be identified through the MCP official `userAgent`

qchuchu:
Closes #909.

## What

Adds a `useHostInfo` hook so a view can tell which host it's rendering in (Claude, Cursor, Goose, …) and adapt copy, shortcuts, or layout accordingly.

```tsx
import { useHostInfo } from "skybridge/web";

const { name, version } = useHostInfo();
if (name === "claude") return <ClaudeLayout />;
```

The value is sourced from the MCP Apps `ui/initialize` handshake (`hostInfo`), which ext-apps already receives but Skybridge was dropping. `name` is normalized to a `Host` slug when recognized, and passes through the raw string otherwise.

## How

- `bridges/mcp-app/types.ts` — add `hostInfo: Implementation | null` to the bridge context
- `bridges/mcp-app/bridge.ts` — populate it in `connect()` via `app.getHostVersion()`
- `hooks/use-host-info.ts` — the hook, the `Host` slug union, and the raw-name→slug map; exported from `skybridge/web`
- docs: new `api-reference/use-host-info` page, nav entry, and overview-matrix row

## Design notes (vs the proposal in #909)

The issue sketched a `useHost` returning `{ name: "chatgpt" | "claude" | ... | "other" }`. This PR keeps that capability and the lowercase-slug shape, with a few deliberate refinements:

1. **`useHostInfo` returning `{ name, version }`** instead of `useHost` returning only `name`. It mirrors the MCP `Implementation` type the data actually comes from, and `version` is free once we're reading `hostInfo` — useful for host-version gating.
2. **Sourced from `hostInfo` (the `ui/initialize` handshake), not `userAgent`/`window.openai` sniffing.** The issue noted `window.openai` isn't ChatGPT-exclusive; `hostInfo.name` is the protocol's own, unambiguous host identity, so there's no false-positive heuristic to maintain.
3. **`Host` is a string-literal union of slugs** (`"chatgpt" | "claude" | "cursor" | "goose" | "mistral-vibe" | "alpic"`), and we translate the raw reported names (`"Claude"`, `"MCP-UI Host"`, `"Le Chat"`, …) to those slugs in one lookup map. `name === "claude"` narrows with full autocomplete and no import. Unknown hosts keep their raw string rather than collapsing to `"other"`, so callers can still read them — the type is `Host | (string & {}) | undefined`, which surfaces the known slugs in autocomplete while accepting any string.

Because the name arrives asynchronously from the handshake, this has to be a hook (not a `Platform`-style constant): `name`/`version` are `undefined` on first render and populate on the re-render once the host responds.

Happy to adjust naming/slugs — flagging that #909 has volunteers, so shout if there's a parallel attempt to reconcile with.

### DIFF
--- a/docs/api-reference/overview.mdx
+++ b/docs/api-reference/overview.mdx
@@ -26,6 +26,7 @@ Every Skybridge export. Server APIs run in your MCP server and work with any hos
 | [`useDisplayMode`](/api-reference/use-display-mode) | Read and request inline, pip, or fullscreen. | <Compat compact alpic chatgpt claude goose /> |
 | [`useDownload`](/api-reference/use-download) | Save files to the user's device. | <Compat compact claude /> |
 | [`useFiles`](/api-reference/use-files) | Upload and pick host-managed files. | <Compat compact chatgpt /> |
+| [`useHostInfo`](/api-reference/use-host-info) | Identify which host is rendering the view. | <Compat compact alpic claude cursor goose /> |
 | [`useLayout`](/api-reference/use-layout) | Read theme, max height, and safe-area insets. | <Compat compact alpic chatgpt claude cursor goose /> |
 | [`useOpenExternal`](/api-reference/use-open-external) | Open a URL outside the view iframe. | <Compat compact alpic chatgpt claude cursor goose /> |
 | [`useRegisterViewTool`](/api-reference/use-register-view-tool) | Expose a tool that runs inside the view. | <Compat compact alpic /> |
--- a/docs/api-reference/use-host-info.mdx
+++ b/docs/api-reference/use-host-info.mdx
@@ -1,0 +1,81 @@
+---
+title: useHostInfo
+description: "Identify which host is rendering the view"
+---
+
+import { Compat } from "/components/compat.jsx";
+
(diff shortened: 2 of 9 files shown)


harijoe (approved):
I'm just a bit afraid of the exact match mapping. Would it be worth relaxing it a bit to at least avoid casing issues?

qchuchu:
@harijoe We can indeed make a simple hook - but I feel here is the value we bring -> by knowing what are the exact host values, and it's will easy to modify in the future. Also, it the host change its value, people can fix it by changing the value in the code (it's string compatible).

qchuchu:
I verified the values while running the mcp apps conformance test (https://alpic-ai.github.io/mcp-app-conformance/ - it's the names & version at the top)

Issue #14, opened 2025-11-22
Support MCP Apps


rodincave:
MCP Apps has been announced as an official extension to MCP: 

Related [SEP-1865](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1865/commits/290629c72f3cc4436e265ebeb390b13c024cd3df) in modelcontextprotocol github

Draft spec to implement: [ext-apps](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/draft/apps.mdx)

ochafik:
Btw, [@modelcontextprotocol/ext-apps](https://www.npmjs.com/package/@modelcontextprotocol/ext-apps) has now a 0.0.1 package on npmjs :-)

PR #1012, opened 2026-07-29
feat(web): apply host styles via skybridge({ hostStyles }) opt-in

qchuchu:
## What

On the **MCP Apps** runtime, hosts (Claude, etc.) can hand a view a set of design tokens — theme, CSS variables, and fonts — via `hostContext.styles`. Skybridge received them but dropped them on the floor. This wires them onto the view, behind an opt-in flag.

Enable it on the Vite plugin:

```ts
skybridge({ hostStyles: true })
```

With it on, Skybridge applies (and keeps in sync on theme toggle):
- **CSS variables** on `:root` — the standardized set (`--color-*`, `--font-*`, `--border-radius-*`, `--shadow-*`), referenced with `var(…)`
- **`color-scheme`** on the document, so the host's `light-dark()` values resolve and native controls match
- **Host fonts**, injected so `var(--font-sans)` renders in the host typeface

Reuses the ext-apps helpers (`applyDocumentTheme` / `applyHostStyleVariables` / `applyHostFonts`) — no reimplementation.

## Why opt-in / default off

Turning it on lets the host set variables on `:root` and flip `color-scheme`, which restyles the view. Since the variables are applied inline on the document root, an app's own `:root { --color-… }` gets overridden. Default off means no existing app is silently restyled; enabling it is a deliberate one-liner.

No-op under the ChatGPT Apps SDK runtime (it sends no tokens).

## How it's wired

`skybridge({ hostStyles })` → generated `mountView(component, { hostStyles })` → `McpAppBridge.getInstance({ applyHostStyles })` → gated apply inside the bridge's `updateContext` (covers the initial connect snapshot and live `host-context-changed` updates).

## Changes

- **plugin**: `SkybridgePluginOptions.hostStyles`, forwarded into the generated entry
- **mountView**: `MountViewOptions.hostStyles`, seeds the bridge before render
- **McpAppBridge**: `applyStylesEnabled` gate; idempotent apply
- **tests**: applies when enabled / untouched by default even when host sends styles / untouched when enabled but host sends nothing
- **docs**: "Match the Host Styles" section in `build/view`, `useLayout` theme tip
- **skills**: host-styles note in the shared `ui-guidelines` reference

## Tracking

Addresses the `styles` (80 CSS variables) context-property gap tracked in #14.

### DIFF
--- a/docs/api-reference/use-layout.mdx
+++ b/docs/api-reference/use-layout.mdx
@@ -53,6 +53,10 @@ theme: Theme;
 
 The host's active color theme, `"light"` or `"dark"`. Mirror it in the view's styling for a native feel. It updates when the user toggles the host theme.
 
+<Tip>
+To match the host automatically, without branching on `theme` yourself, enable [host styles](/build/view#match-the-host-styles): Skybridge then applies the host's theme, palette, and fonts to your view as CSS variables.
+</Tip>
+
 ### `maxHeight`
 
 ```tsx
--- a/docs/build/view.mdx
+++ b/docs/build/view.mdx
@@ -68,7 +68,7 @@ server
 ```
 </CodeGroup>
 
-The sections below walk through it: the component itself, reading the tool, calling tools back, the typed helpers binding views to the server, and opening the sandbox to external domains.
+The sections below walk through it: the component itself, reading the tool, calling tools back, the typed helpers binding views to the server, matching the host's styles, and opening the sandbox to external domains.
 
 ## The Component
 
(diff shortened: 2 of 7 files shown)


harijoe:
Maybe templates should have this option enabled by default? And we would enable it by default definitively in v2?

qchuchu:
Well I don't know how it's going right now - maybe in the future host will require to have a minimum of styling that matches there website but for now, I don't see that much.

But I thought it was a good idea to make some marketing (and we could actually add in the devtools the current "styles" that are applied on each host that we know, so that users can check how their app would render if they use the variables).

Issue #1018, opened 2026-07-31
Build fails for any view whose CSS contains a `url()`


thomaswelton:
## Summary

`renderBuiltUrl` in [`packages/core/src/web/plugin/plugin.ts#L134`](https://github.com/alpic-ai/skybridge/blob/main/packages/core/src/web/plugin/plugin.ts#L134) rewrites **every** built asset URL into a runtime JavaScript expression. Vite has nowhere to evaluate a JS expression inside a `.css` file, so `vite:css-post` throws and the build fails outright for any view importing a stylesheet that contains a `url()` — a `@font-face`, a `background-image`, anything.

## Reproduction

```bash
npm create skybridge@latest repro -- --yes --pm npm
cd repro

cat >> src/index.css <<'EOF'

.repro-background {
  background-image: url("./views/images/mascot/beret.png");
}
EOF

npm run build
```

`npm run build` on the untouched scaffold succeeds. Appending those three lines is the only change, and it fails:

```
error during build:
Build failed with 1 error:

[plugin vite:css-post]
Error: { runtime: "window.skybridge.serverUrl + "/assets/assets/beret-Cdr47jta.png"" } is not supported for assets in css files: assets/beret-Cdr47jta.png
    at toOutputFilePathWithoutRuntime (…/node_modules/vite/dist/node/chunks/node.js:34022:30)
    at …/node_modules/vite/dist/node/chunks/node.js:22675:28
    at String.replace (<anonymous>)
    at resolveAssetUrlsInCss (…/node_modules/vite/dist/node/chunks/node.js:22672:26)
    at PluginContextImpl.renderChunk (…/node_modules/vite/dist/node/chunks/node.js:22769:18)
```

**Environment:** `skybridge@1.3.2`, `vite@8.1.5` (as installed by the scaffolder), Node 24.15.0, macOS.

## Cause

Introduced by #658 ("fix: resolve asset URLs at runtime to support tunnels"):

```ts
experimental: {
  renderBuiltUrl: (filename) => {
    return {
      runtime: `window.skybridge.serverUrl + "/assets/${filename}"`,
    };
  },
},
```

The `{ runtime }` form is applied unconditionally, but Vite only supports it where there's a JS scope to evaluate it in.

## Suggested fix

Narrow the rewrite to `hostType === "js"`, the pattern from [Vite's own `renderBuiltUrl` example](https://vite.dev/guide/build#advanced-base-options):

```ts
experimental: {
[...]

PR #1021, opened 2026-07-31
fix(1018): only rewrite asset URLs at runtime for JS, not CSS

thomaswelton:
Fix for #1018

`renderBuiltUrl` rewrote every built asset URL into a runtime JS expression. Vite has nowhere to evaluate that inside a stylesheet, so `vite:css-post` throws and the build fails outright for any view whose CSS contains a `url()` — a `@font-face`, a background image, anything a component library ships:

    [plugin vite:css-post]
    Error: { runtime: "window.skybridge.serverUrl + "/assets/assets/demo-0IQFxQqs.woff2"" }
    is not supported for assets in css files: assets/demo-0IQFxQqs.woff2

Narrow the rewrite to `hostType === "js"`, per Vite's documented `renderBuiltUrl` example, and fall back to `{ relative: true }`.

The runtime rewrite (#658) is still required for JS: views render in a host sandbox iframe, so `import.meta.url` points at the sandbox domain rather than the Skybridge server and can't follow a tunnel. That doesn't apply to CSS — a `url()` resolves against the stylesheet's own URL, and the stylesheet is already served from `${serverUrl}/assets/…`, so the tunnel origin is inherited automatically. This also brings the build path in line with the dev-server transform, which has excluded CSS since #697.

### DIFF
--- a/packages/core/src/web/plugin/plugin-build.test.ts
+++ b/packages/core/src/web/plugin/plugin-build.test.ts
@@ -1,0 +1,67 @@
+// @vitest-environment node
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { build } from "vite";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { skybridge } from "./plugin.js";
+
+/**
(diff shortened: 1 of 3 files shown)


harijoe (approved):
Thanks for your contribution @thomaswelton

PR #1023, opened 2026-07-31
feat(auth): type-safe AuthInfo.extra, owned by the verifier

### LINKED ISSUE #1016: Type-safe AuthInfo.extra (generic over Record<string, unknown>)
## Feedback

`AuthInfo["extra"]` is currently typed as `Record<string, unknown>`. It would be useful to type the `extra` fields more precisely so auth code does not need type inference / casting everywhere. This would fit Skybridge's type-safe approach.

## Context

Skybridge re-exports `AuthInfo` from `@modelcontextprotocol/sdk`:

```ts
export type { AuthInfo } from "@modelcontextprotocol/sdk/server/auth/types.js";
```

Built-in JWT verification already populates `extra` with claims (`subject` + remaining JWT payload) in `createJwksVerifier`, but consumers still see an untyped bag:

```ts
extra: { subject: sub, ...rest },
```

Today app code typically casts or narrows manually when reading `extra.authInfo` / `auth.extra` in tools and middleware.

## Suggested direction

Allow apps (and providers) to parameterize `AuthInfo` / the auth surface with a concrete `extra` shape, e.g.:

```ts
type MyExtra = { subject: string; orgId: string; roles: string[] };

// Ideal: AuthInfo<MyExtra> or a Skybridge-level generic that flows into
// tool handlers / mcpMiddleware / requireBearerAuth without casts.
```

Possible approaches to explore:

1. Generic `AuthInfo<TExtra extends Record<string, unknown> = Record<string, unknown>>` (SDK upstream and/or Skybridge wrapper).
2. A Skybridge-level declaration / server generic that threads `TExtra` into `extra.authInfo` for tools and MCP middleware.
3. Provider helpers that return a typed verifier (`verifyAccessToken(): Promise<AuthInfo<MyExtra>>`).

## Why it matters

- Removes `as AuthInfo` / claim casting scattered through auth examples and production apps.
- Aligns with Skybridge's type-safe DX for tools, views, and middleware.
- Makes IdP-specific claims (org, roles, tenant, …) first-class at the type level.

## Source

User feedback.

harijoe:
Closes #1016

## Summary

- `AuthInfo<TExtra>` types the `extra` claim bag; bare `AuthInfo` is unchanged.
- The claim type is owned by the verifier and flows outward: `createJwksVerifier` → `OAuthConfig<TExtra>` → `SkybridgeServerOptions<TExtra>` → inferred at `new McpServer(...)` → tool handlers and `mcpMiddleware`.
- Each branded provider ships the claims its docs promise, so `workosProvider()` alone types `extra.authInfo?.extra?.org_id`. Every field is optional, since claims are tenant-configurable.
- Override per provider to add your own: `workosProvider<{ tenant: string }>({ ... })` intersects with the provider's base, so IdP claims are never redeclared.
- Casts dropped from all 7 auth examples.
- `OAuthConfig` gains `verifier` (the built verifier, carrying the claim type) alongside the now-deprecated `verify` parameters. Exactly one is required, enforced by the type. A hand-built config keeps working on `verify`, with claims left as `Record<string, unknown>`.

Additive: `TAuthExtra` is a second generic with a default, so positional `McpServer<MyTools>` and plain `McpServer` annotations keep working.

## Alternatives Considered

**A server-level `withAuthExtra<Claims>()` carrier**, chained before the tools, in the style of Fastify's `withTypeProvider` and tRPC's `.context`. It covers every path, including a hand-written verifier mounted through `requireBearerAuth` with no `oauth` option at all. Rejected on Fred's review: a no-op call that exists only to carry a type is a lot of API surface, and it puts the declaration somewhere the verifier's actual return type cannot be checked against, so the two can silently diverge. Attaching the type to the verifier gives one source instead of two.

**Generic reordering (`new McpServer<Claims>(...)`).** Reads well and is what Hono and Kysely do. Rejected because it reorders `McpServer`'s type params, breaking positional `McpServer<MyTools>` users for a type-only feature.

**A zod claims schema on `oauth.verify`, inferred into the generic.** Would validate claims at the JWKS boundary rather than asserting them. Rejected because `extra` is not always JWT claims: a hand-written `verifyAccessToken` doing introspection or a DB lookup fills it too, and those apps have no `oauth.verify` to hang a schema on.

**Declaration merging on an `AuthExtra` interface**, as Passport does with `Express.User`. Smallest diff, but one global shape per app and no way to differ between two auth sources.

## Compatibility

One breaking change, and it only bites code reading the resolved JWKS parameters off a provider's result:

- A provider now returns `verifier`, not `verify`, so `(await workosProvider({ … })).verify.issuer` no longer compiles, and reads as `undefined` at runtime in JS. The issuer and JWKS URL are still what discovery resolved; nothing else about them changed. Hand-built configs are unaffected, since `verify` stays accepted as an input.

Two further type-level effects, both runtime-neutral:

- Reading a claim a provider does not declare now errors where it used to be `unknown` — for example `extra.authInfo?.extra?.my_claim` on a WorkOS-typed server. Name it on the provider (`workosProvider<{ my_claim: string }>({ ... })`), or keep casting through bare `AuthInfo`, which still compiles.
- Passing `TTools` explicitly (`new McpServer<MyTools>(...)`) falls back to the untyped claim bag: TypeScript has no partial type-argument inference, so naming the first generic drops inference of the second. Leave `TTools` inferred.

Everything else is unchanged, verified against `main`: no export removed, bare `AuthInfo` interchangeable with the SDK's in both directions, bare `OAuthConfig`, bare `McpServer`, positional `McpServer<MyTools>`, `InferTools`/`$types`, a no-`oauth` server's untyped claim bag, and the `extra.authInfo as AuthInfo` pattern all still compile. Runtime verification and `extra` contents are identical.

## Notes

- Still an assertion, not a runtime check: nothing validates that a token carries the declared claims. What changes is that the type has a single owner.
- Provider defaults are drawn from each provider's own docs and linked in TSDoc. No provider ships `email` in an access token; where a JWT template or claims action can add it, it is typed optional. Auth0 is excluded because non-namespaced custom claims on an API-audience access token are silently dropped there.
- `RegisteredClaims` (`iss`, `aud`, `iat`, `nbf`, `jti`) is intersected into every JWKS-verified shape, since those claims survive the mapping into `extra` and code reading them would otherwise stop compiling.
- Writing the verifier by hand still works: type it `TokenVerifier<Claims>` and put it in the config, and the shape flows the same way. Only the middleware-only path (`requireBearerAuth` with no `oauth` option) gets nothing, since there is no config for the type to travel through.
- Spreading a provider config to swap its verifier (`{ ...config, verifier }`) no longer type-checks, since the union forbids carrying both fields. Build the config explicitly instead.

### DIFF
--- a/packages/core/src/server/auth.ts
+++ b/packages/core/src/server/auth.ts
@@ -15,6 +15,9 @@ export {
   type AuthMetadataOptions,
   mcpAuthMetadataRouter,
 } from "@modelcontextprotocol/sdk/server/auth/router.js";
+/** Claims a verifier puts in `AuthInfo["extra"]`: any JSON-ish bag of them. */
+export type ExtraClaims = Record<string, unknown>;
+
 /**
  * A validated access token, as resolved by a
  * [verifier](https://docs.skybridge.tech/api-reference/verifier) and handed to
@@ -35,9 +38,10 @@ export {
  * }
  * ```
  */
-export type AuthInfo<
-  TExtra extends Record<string, unknown> = Record<string, unknown>,
-> = Omit<SdkAuthInfo, "extra"> & { extra?: TExtra };
+export type AuthInfo<TExtra extends ExtraClaims = ExtraClaims> = Omit<
+  SdkAuthInfo,
+  "extra"
+> & { extra?: TExtra };
 
 /**
  * Checks a bearer token and resolves the claims it carries. The type parameter
@@ -49,9 +53,7 @@ export type AuthInfo<
  *
  * @typeParam TExtra - Claims this verifier populates in `AuthInfo["extra"]`.
  */
-export type TokenVerifier<
-  TExtra extends Record<string, unknown> = Record<string, unknown>,
-> = {
+export type TokenVerifier<TExtra extends ExtraClaims = ExtraClaims> = {
   verifyAccessToken(token: string): Promise<AuthInfo<TExtra>>;
 };
 
--- a/packages/core/src/server/auth.ts
+++ b/packages/core/src/server/auth.ts
@@ -2,6 +2,7 @@ import {
   type BearerAuthMiddlewareOptions,
   requireBearerAuth,
 } from "@modelcontextprotocol/sdk/server/auth/middleware/bearerAuth.js";
+import type { AuthInfo as SdkAuthInfo } from "@modelcontextprotocol/sdk/server/auth/types.js";
 
 import type { RequestHandler } from "express";
 
@@ -14,7 +15,29 @@ export {
   type AuthMetadataOptions,
   mcpAuthMetadataRouter,
 } from "@modelcontextprotocol/sdk/server/auth/router.js";
-export type { AuthInfo } from "@modelcontextprotocol/sdk/server/auth/types.js";
+/**
+ * A validated access token, as resolved by a
+ * [verifier](https://docs.skybridge.tech/api-reference/verifier) and handed to
+ * handlers on `extra.authInfo`.
+ *
+ * Unparameterized, `extra` stays the SDK's `Record<string, unknown>` bag. Pass
+ * the claims your verifier produces to type it, and declare the same shape on
+ * the server with {@link McpServer.withAuthExtra} so handlers see it too.
+ *
+ * @typeParam TExtra - Claims populated in `extra`. Makes `extra` required.
+ *
+ * @example
+ * ```ts
+ * type Claims = { sub: string; email?: string };
+ *
+ * async function verifyAccessToken(token: string): Promise<AuthInfo<Claims>> {
+ *   // ...
+ * }
+ * ```
+ */
+export type AuthInfo<
+  TExtra extends Record<string, unknown> = Record<string, unknown>,
+> = Omit<SdkAuthInfo, "extra"> & { extra?: TExtra };
 
 /**
  * Like `requireBearerAuth`, but lets requests through when no
--- a/packages/core/src/server/auth.ts
+++ b/packages/core/src/server/auth.ts
@@ -20,15 +20,15 @@ export {
  * [verifier](https://docs.skybridge.tech/api-reference/verifier) and handed to
  * handlers on `extra.authInfo`.
  *
- * Unparameterized, `extra` stays the SDK's `Record<string, unknown>` bag. Pass
- * the claims your verifier produces to type it, and declare the same shape on
- * the server with {@link McpServer.withAuthExtra} so handlers see it too.
+ * Unparameterized, `extra` stays the SDK's `Record<string, unknown>` bag. The
+ * claims are owned by whoever verifies the token: parameterize the verifier and
+ * the shape flows through {@link OAuthConfig} to every handler.
  *
- * @typeParam TExtra - Claims populated in `extra`. Makes `extra` required.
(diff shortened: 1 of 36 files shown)


fredericbarthelet (commented):
Thanks for putting together this first implementation @harijoe 

using a no-op extra chained function call to pass a type parameter feels a bit too much. You also introduce risks of diverging from actual return type from the verifier.

Just for the sake of sharing, I would have start this work with updating the return type of the auth Skybridge implementation to use a type parameter instead of default AuthInfo type from SDK.

https://github.com/alpic-ai/skybridge/blob/6b1b79ab378c0fba6eed4c4f0c1bdcc4d78faa1b/packages/core/src/server/auth/verify.ts#L26

would become 

```ts
async verifyAccessToken<TAuthInfo extends AuthInfo>(token: string): Promise<TAuthInfo> {
```

This would need to then be bubbled up, probably up to the customProvider method expected by the `oauth` property. Each provider implementation could then override this param using provider-known claims provided in token, or allowing user to provide their own oauth provider implementation with their own verifier function.

Did you try this approach?

packages/core/src/server/auth.ts
```diff
@@ -14,7 +15,45 @@ export {
...
+ * type Claims = { subject?: string; email?: string };
+ *
+ * async function verifyAccessToken(token: string): Promise<AuthInfo<Claims>> {
+ *   // ...
+ * }
+ * ```
+ */
+export type AuthInfo<
+  TExtra extends Record<string, unknown> = Record<string, unknown>,
+> = Omit<SdkAuthInfo, "extra"> & { extra?: TExtra };
```
fredericbarthelet:
```suggestion
> = Omit<SdkAuthInfo, "extra"> & { extra: {subject?: string;} & TExtra };
```

packages/core/src/server/auth.ts
```diff
@@ -14,7 +15,45 @@ export {
...
+ * ```ts
+ * type Claims = { subject?: string; email?: string };
+ *
+ * async function verifyAccessToken(token: string): Promise<AuthInfo<Claims>> {
+ *   // ...
+ * }
+ * ```
+ */
+export type AuthInfo<
+  TExtra extends Record<string, unknown> = Record<string, unknown>,
```
fredericbarthelet: just maybe worth putting a type `type ExtraClaims = Record<string, unknown>` to be used in place of default and extended of type to avoid repetition

packages/core/src/server/auth/index.ts
```diff
@@ -11,9 +23,28 @@ export type OAuthConfig = {
...
+      /**
+       * JWKS verification parameters, used to build the verifier.
+       *
+       * @deprecated Pass `verifier: createJwksVerifier({ … })` instead, which
+       * also types the claims handlers receive. Claims stay
+       * `Record<string, unknown>` on this path.
+       */
+      verify: JwksVerifyConfig;
+      verifier?: never;
+    }
```
fredericbarthelet, line 47: I don't understand the need to change the interface contract for oauth config. Can you explain why revert initial `verify` to `verifier`?

fredericbarthelet (approved):
LGTM ! Thanks @harijoe for taking this. Only remaining small details are config to be removed from return type of `createJwksVerifier` and `subject` to be used as base claim

PR #1024, opened 2026-07-31
fix(devtools): scope forwarded headers to the MCP server origin

muralx:
DevTools sends `x-forwarded-host` and `x-forwarded-proto` so the MCP server can recover its public origin in local development, where no proxy sets them. They are supplied through `requestInit`, which the transport applies to every request it makes — including the cross-origin discovery, registration and token requests to the authorization server.

Custom headers make a cross-origin request non-simple, so the browser sends a preflight first, and the request proceeds only if the authorization server names both headers in `Access-Control-Allow-Headers`. An authorization server has no reason to allow headers it does not read, so the OAuth flow stops at discovery.

This scopes the headers to the MCP server's own origin. Requests to the configured server URL carry them exactly as before; cross-origin requests go out untouched.

## Contents

| | |
|---|---|
| `packages/devtools/src/lib/mcp/client.ts` | scope the headers to the server origin |
| `packages/devtools/src/lib/mcp/client.test.ts` | 7 tests |

## Behaviour

| request | before | after |
|---|---|---|
| MCP server | forwarded headers sent | unchanged |
| authorization server — discovery, registration, token | forwarded headers sent | no forwarded headers |

The headers are applied through the transport's `fetch` option rather than `requestInit`, since that is the only hook with visibility of a request's destination. Headers the transport has already set — `Authorization`, `mcp-session-id`, `content-type` — are preserved, and the remainder of the request init passes through unchanged.

## Verification

`@skybridge/devtools`: 13 unit tests pass, types pass, `biome ci` clean across 81 files.

Three of the seven new tests fail against the current implementation, each on the header reaching the authorization server rather than on the change of mechanism.

End to end against an authorization server with dynamic client registration: discovery, registration, authorization, consent, token issuance, `tools/list` and `tools/call` all succeed, and no forwarded headers reach the authorization server.

### DIFF
--- a/packages/devtools/src/lib/mcp/client.test.ts
+++ b/packages/devtools/src/lib/mcp/client.test.ts
@@ -1,0 +1,160 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { McpClient } from "./client.js";
+
+const SERVER_URL = "http://localhost:3000/mcp";
+const AUTH_SERVER_URL =
+  "http://localhost:9000/.well-known/openid-configuration";
+
+type TransportOptions = {
+  fetch?: typeof fetch;
+  requestInit?: RequestInit;
+};
+
+const mock = vi.hoisted(() => ({
+  transportOptions: undefined as TransportOptions | undefined,
+}));
+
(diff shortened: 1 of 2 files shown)


harijoe (approved):
Thanks for your contribution @muralx

PR #760, opened 2026-05-12
fix: allow esbuild builds for pnpm 11 installs

asim48-ctrl:
## Summary
- allow pnpm to run esbuild's required build step from the workspace config
- fixes fresh pnpm 11 installs failing with ERR_PNPM_IGNORED_BUILDS for esbuild

Fixes #758.

## Verification
- git diff --check
- corepack pnpm@11.1.1 --pm-on-fail=ignore install --lockfile-only --ignore-scripts

Note: I intentionally left scripts disabled during local validation. The pnpm 11 workspace config parsed cleanly; my local Node is 24.12.0, so pnpm also emitted the expected engine warning because this repo requires >=24.14.x.

<h3>Greptile Summary</h3>

This PR fixes fresh pnpm 11 installs failing with `ERR_PNPM_IGNORED_BUILDS` by adding an `allowBuilds` section to `pnpm-workspace.yaml` that explicitly permits esbuild's postinstall build step, which pnpm 11 blocks by default unless whitelisted.

- Adds `allowBuilds: esbuild: true` — the correct pnpm 11 syntax to allow esbuild's required build script to run during install.
- The existing `ignoredBuiltDependencies` entries (`sharp`, `unrs-resolver`) are left in place; in pnpm 11 this field is deprecated and silently ignored, but since pnpm 11 blocks all builds by default, those packages remain effectively blocked without any explicit `allowBuilds: false` entry.

<h3>Confidence Score: 5/5</h3>

Safe to merge — a single-field addition to the workspace config that unblocks esbuild installs on pnpm 11 with no risk to existing functionality.

The change is minimal and correct: `allowBuilds: esbuild: true` is exactly the pnpm 11 replacement for allowing esbuild's postinstall step. The deprecated `ignoredBuiltDependencies` entries are inert in pnpm 11 (all builds are blocked by default unless explicitly allowed), so sharp and unrs-resolver remain protected. No logic, API surface, or runtime behavior is affected.

No files require special attention.

<details><summary>Prompt To Fix All With AI</summary>

`````markdown
Fix the following 1 code review issue. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 1
pnpm-workspace.yaml:9-14
[...]

PR #1030, opened 2026-08-03
fix(create-skybridge): allow esbuild builds for pnpm scaffolds

### LINKED ISSUE #1029: bug: fresh create-skybridge scaffolds fail to install with pnpm ≥10 (ERR_PNPM_IGNORED_BUILDS: esbuild)
## Steps to reproduce

```bash
pnpm create skybridge@latest my-app --yes --pm pnpm
```

## Expected

Scaffold completes and dependencies install.

## Actual

The install step fails and the CLI aborts:

```
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: esbuild@0.28.1
Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
■  Try manually: cd my-app && pnpm install
└  Aborted
```

Running `pnpm dev` in an affected project fails the same way, because `skybridge dev` re-runs the install check.

## Root cause

pnpm ≥10 skips dependency build scripts unless allow-listed, and pnpm 11 turns the skipped esbuild postinstall into a hard install error. alpic-ai/skybridge#758 reported this for the skybridge monorepo itself and [#760](<https://github.com/alpic-ai/skybridge/issues/760>) fixed it by adding `allowBuilds` to the repo's own `pnpm-workspace.yaml` — but the `create-skybridge` templates ship no such file, so every freshly scaffolded project still hits the error.

## Environment

* create-skybridge / Skybridge v1.3.2
* pnpm 11.15.1 (hard error) — pnpm 10 warns and skips the build instead
* Node 22.23.1, macOS (darwin-arm64)

## Suggested fix

Have the create CLI write a `pnpm-workspace.yaml` into the project when the selected package manager is pnpm, before running the install:

```yaml
onlyBuiltDependencies:
  - esbuild
allowBuilds:
  esbuild: true
```

`onlyBuiltDependencies` is read by pnpm 10, `allowBuilds` by pnpm 11+, and each version ignores the other's key (verified locally with pnpm 10.18.3 and 11.15.1). Writing it conditionally in the CLI keeps npm/yarn/bun/deno scaffolds untouched. PR incoming.

qchuchu:
## Summary

- Fresh `create-skybridge` scaffolds abort at the install step on pnpm ≥10: pnpm skips esbuild's postinstall unless allow-listed, and pnpm 11 turns that into a hard `ERR_PNPM_IGNORED_BUILDS` error. Same failure repeats on `pnpm dev`, which re-runs the install check.
- #760 fixed this for the monorepo's own `pnpm-workspace.yaml` but the scaffolded templates were left uncovered — this closes that gap.
- The create CLI now writes a `pnpm-workspace.yaml` into the project right before installing, only when the selected package manager is pnpm, so npm/yarn/bun/deno scaffolds are untouched:

```yaml
onlyBuiltDependencies:
  - esbuild
allowBuilds:
  esbuild: true
```

`onlyBuiltDependencies` is read by pnpm 10, `allowBuilds` by pnpm 11+; each version ignores the other's key.

Fixes #1029

## Notes

- The file also makes the scaffold its own pnpm workspace root. For a standalone project that's a no-op; it's arguably desirable isolation if someone scaffolds inside an existing monorepo, but flagging it for review.
- Existing projects scaffolded before this fix need the yaml added manually — the error message pnpm prints (`pnpm approve-builds`) also resolves it interactively.

## Verification

- `pnpm run test:unit` in `packages/create-skybridge` — 5 passed (2 new: pnpm scaffolds get the file with both keys, `--pm npm` scaffolds don't).
- `pnpm run test:format` (biome ci) — clean.
- Dual-key file verified against real installs of an esbuild-depending project: pnpm 10.18.3 and pnpm 11.15.1 both install cleanly and esbuild's binary is present.
- Repro of the original failure: `pnpm create skybridge@latest my-app --yes --pm pnpm` with pnpm 11.15.1 aborts as described in #1029.

### DIFF
--- a/packages/create-skybridge/src/index.test.ts
+++ b/packages/create-skybridge/src/index.test.ts
@@ -70,7 +70,7 @@ describe("create-skybridge", () => {
       path.join(process.cwd(), tempDirName, "project", "pnpm-workspace.yaml"),
       "utf-8",
     );
-    // onlyBuiltDependencies covers pnpm 10, allowBuilds covers pnpm 11+.
+    expect(workspaceRaw).toContain('packages:\n  - "."');
     expect(workspaceRaw).toContain("onlyBuiltDependencies:\n  - esbuild");
     expect(workspaceRaw).toContain("allowBuilds:\n  esbuild: true");
   });
--- a/packages/create-skybridge/src/index.test.ts
+++ b/packages/create-skybridge/src/index.test.ts
@@ -62,6 +62,30 @@ describe("create-skybridge", () => {
     ).rejects.toThrow();
   });
 
+  it("writes pnpm-workspace.yaml allowing esbuild builds when using pnpm", async () => {
+    const name = `${tempDirName}/project`;
+    await init([name, "--yes", "--skip-skills", "--pm", "pnpm"]);
+
+    const workspaceRaw = await fs.readFile(
+      path.join(process.cwd(), tempDirName, "project", "pnpm-workspace.yaml"),
+      "utf-8",
+    );
+    // onlyBuiltDependencies covers pnpm 10, allowBuilds covers pnpm 11+.
+    expect(workspaceRaw).toContain("onlyBuiltDependencies:\n  - esbuild");
+    expect(workspaceRaw).toContain("allowBuilds:\n  esbuild: true");
+  });
+
+  it("does not write pnpm-workspace.yaml for other package managers", async () => {
+    const name = `${tempDirName}/project`;
+    await init([name, "--yes", "--skip-skills", "--pm", "npm"]);
+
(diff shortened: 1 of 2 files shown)


qchuchu:
Follow-up: the initial version broke pnpm 9, which hard-errors on a `pnpm-workspace.yaml` without a `packages` field (`ERROR packages field missing or empty`). The written file now includes `packages: ["."]`. Verified real installs of an esbuild-depending project on pnpm 9.15.9, 10.18.3 and 11.15.1 — all exit 0 with no warnings and esbuild's binary present.

harijoe (approved):

PR #1036, opened 2026-08-03
feat(core): implement skills/list and skills/get per current SEP-2640

### LINKED ISSUE #1035: Skills over MCP tracks a superseded SEP-2640 draft — missing skills/list and skills/get (breaks ChatGPT skill import)
## Problem

`packages/core/src/server/skills.ts` implements [SEP-2640](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2640) as it stood in early July: a well-known `skill://index.json` resource with `{url, digest, frontmatter}` entries (single SKILL.md digest per skill) plus the optional `directoryRead` capability and `resources/directory/read` method.

The SEP was revised mid-July and those shapes no longer exist in it:

- `cff984c3` (Jul 13) — Replace single skill digest with per-file `resources` manifest
- `9b9d299a` (Jul 13) — Replace `skill://index.json` with a `skills/list` method
- `d7490ecd` (Jul 15) — Add `skills/get` for single-skill entry retrieval

The current revision states that declaring the `io.modelcontextprotocol/skills` extension commits the server to implementing `skills/list` and `skills/get`. Skybridge declares the extension but returns `-32601 Method not found` for both.

[OpenAI's skill importer](https://developers.openai.com/plugins/build/mcp-server#import-skills-from-the-mcp-server) implements a bounded, static subset of the current revision and uses `skills/list` as its discovery entry point, so importing skills from a Skybridge server into a ChatGPT plugin draft fails at the first step. Verified against two production servers on skybridge 1.3.2 (alpix.alpic.ai and a fresh template deploy).

## Expected (current SEP-2640)

- `skills/list` — paginated (optional `cursor` param, optional `nextCursor` in result), returning entries `{uri, frontmatter, resources: [{uri, digest}]}` where `frontmatter` is the verbatim parsed SKILL.md frontmatter and `resources` is the complete file set — SKILL.md and every supporting file — each digested as `sha256:<64 lowercase hex>` over the file's raw bytes
- `skills/get` — accepts `{uri}` of a listed SKILL.md, returns `{skill: <same entry shape>}`

## Notes

- The change is additive and needs no SDK upgrade: neither v1 (1.30.0) nor v2 (2.0.0-beta.5) of the MCP SDK ships skills support, so both methods can be registered via `server.server.setRequestHandler`, exactly like `resources/directory/read` today.
- Everything needed is already in memory in `registerSkills` (each skill's `files` map); per-file digests just need computing.
- `skill://index.json` is no longer part of the SEP — decide whether to keep serving it for backward compatibility or drop it.

qchuchu:
Closes #1035

## Why

Skybridge's skills-over-MCP implementation tracks the pre-July-13 draft of [SEP-2640](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2640): a `skill://index.json` resource with a single SKILL.md digest per skill, plus `resources/directory/read`. The SEP has since replaced the index with two mandatory methods — `9b9d299a` (index.json → `skills/list`), `d7490ecd` (add `skills/get`), `cff984c3` (per-file `resources` digests) — and the current revision states that declaring the `io.modelcontextprotocol/skills` extension commits the server to both. Skybridge declares the extension but returned `-32601` for both methods, which breaks [ChatGPT's skill importer](https://developers.openai.com/plugins/build/mcp-server#import-skills-from-the-mcp-server) at its discovery step (it consumes a static subset of the current SEP revision, entering through `skills/list`).

## What

- `skills/list`: returns all entries in one page as `{uri, frontmatter, resources: [{uri, digest}]}` — verbatim frontmatter, complete file set (SKILL.md first, then supporting files sorted), each file digested as `sha256:<hex>` over its raw bytes. Accepts the optional `cursor` param; never emits `nextCursor` since the whole in-memory catalog fits one page.
- `skills/get`: accepts `{uri}` of a SKILL.md, returns `{skill: <same entry shape>}`; `InvalidParams` for unknown skills or URIs not pointing at a SKILL.md.
- Both registered via `setRequestHandler` like the existing `resources/directory/read` — no SDK change needed (neither SDK v1 nor v2-beta ships skills support).
- Entries are computed once at registration from the manifest already in memory.
- `skill://index.json` is removed: it is no longer part of the SEP, and no host sources skills from it — ChatGPT's importer enters through `skills/list`, and the Alpic playground is migrating to it in alpic-ai/alpic#2308.
- Docs: one paragraph in `guides/skills.mdx` on host discovery via the new methods.

## Verification

- Unit tests for both handlers (entry shape, digest correctness, ordering, error paths) and an integration round-trip over `connectStatelessTransport` with a real MCP client asserting `skills/list` and `skills/get` return identical, digest-carrying entries.
- `pnpm test && pnpm build` green from root (398 unit tests).

### DIFF
--- a/packages/core/src/server/skills.ts
+++ b/packages/core/src/server/skills.ts
@@ -181,6 +181,31 @@ const DirectoryReadRequestSchema = z.object({
   params: z.object({ uri: z.string(), cursor: z.string().optional() }),
 });
 
+const SkillsListRequestSchema = z.object({
+  method: z.literal("skills/list"),
+  params: z.object({ cursor: z.string().optional() }).optional(),
+});
+
+const SkillsGetRequestSchema = z.object({
+  method: z.literal("skills/get"),
+  params: z.object({ uri: z.string() }),
+});
+
+const toSkillEntry = (skill: Skill) => ({
+  uri: `skill://${skill.name}/SKILL.md`,
+  frontmatter: skill.frontmatter,
+  resources: Object.entries(skill.files)
+    .sort(
+      ([a], [b]) =>
+        Number(b === "SKILL.md") - Number(a === "SKILL.md") ||
+        a.localeCompare(b),
+    )
+    .map(([path, content]) => ({
+      uri: `skill://${skill.name}/${path}`,
+      digest: sha256(content),
+    })),
+});
+
 export function registerSkills(
   server: SkillRegistrar,
   manifest: SkillsManifest,
@@ -240,6 +265,26 @@ export function registerSkills(
       ],
     }),
   );
+
+  const entryByName = new Map(
+    manifest.map((skill) => [skill.name, toSkillEntry(skill)]),
+  );
+
+  server.server.setRequestHandler(SkillsListRequestSchema, () => ({
+    skills: [...entryByName.values()],
+  }));
+
+  server.server.setRequestHandler(SkillsGetRequestSchema, ({ params }) => {
+    const { name, relPath } = skillUriToRelPath(params.uri);
+    const entry = relPath === "SKILL.md" ? entryByName.get(name) : undefined;
+    if (!entry) {
+      throw new McpError(
+        ErrorCode.InvalidParams,
+        `Unknown skill: ${params.uri}`,
+      );
+    }
+    return { skill: entry };
+  });
 
   server.server.setRequestHandler(DirectoryReadRequestSchema, ({ params }) => {
     const { name, relPath } = skillUriToRelPath(params.uri);
--- a/packages/core/src/server/skills.ts
+++ b/packages/core/src/server/skills.ts
@@ -37,8 +37,8 @@ export interface SkillResource {
 }
 
 export interface Skill {
-  name: string;
-  frontmatter: Record<string, unknown>;
+  uri: string;
+  frontmatter: { name: string; description: string } & Record<string, unknown>;
   resources: SkillResource[];
 }
 
@@ -125,7 +125,7 @@ export function discoverSkills(dir: string): SkillsManifest {
       }));
 
     skills.push({
-      name: entry.name,
+      uri: `skill://${entry.name}/SKILL.md`,
       frontmatter: parsed.data,
       resources,
     });
@@ -154,7 +154,7 @@ function listDir(
   relPath: string,
 ): { name: string; mimeType: string }[] | null {
   const prefix = relPath === "" ? "" : `${relPath}/`;
-  const base = `skill://${skill.name}/`;
+  const base = skill.uri.slice(0, -"SKILL.md".length);
   const children = new Map<string, string>();
   let matched = relPath === "";
   for (const filePath of skill.resources.map((resource) =>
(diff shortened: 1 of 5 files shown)


packages/core/src/server/skills.ts
```diff
@@ -181,6 +180,31 @@ const DirectoryReadRequestSchema = z.object({
...
+  method: z.literal("skills/list"),
+  params: z.object({ cursor: z.string().optional() }).optional(),
+});
+
+const SkillsGetRequestSchema = z.object({
+  method: z.literal("skills/get"),
+  params: z.object({ uri: z.string() }),
+});
+
+const toSkillEntry = (skill: Skill) => ({
```
fredericbarthelet: your change is very additif. `files` is not in-use anymore on the spec. It has been replaced with `resources`. Instead of creating a new transformer from the "old" skill format to the new one, might be easier to just replace `files` on Skill type, with `resources`. WDYT ?

packages/core/src/server/skills.ts
```diff
@@ -241,6 +266,26 @@ export function registerSkills(
     }),
   );
 
+  const entryByName = new Map(
+    manifest.map((skill) => [skill.name, toSkillEntry(skill)]),
+  );
```
fredericbarthelet: if you change the skill format, you can reuse the existing Map named `byName` instead of creating a new one

packages/core/src/server/skills.ts
```diff
@@ -181,13 +199,33 @@ const DirectoryReadRequestSchema = z.object({
...
+  uri: `skill://${skill.name}/SKILL.md`,
+  frontmatter: skill.frontmatter,
+  resources: skill.resources.map(({ uri, digest }) => ({ uri, digest })),
+});
+
 export function registerSkills(
   server: SkillRegistrar,
   manifest: SkillsManifest,
 ): void {
   const byName = new Map(manifest.map((s) => [s.name, s]));
```
fredericbarthelet: manifest should use `byUri` instead of byName since name isn't considered as required to be unique anymore

> Within a server's listing, names SHOULD be unique, but they are not guaranteed to be: two skills at different paths may share a final segment (acme/billing/refunds and acme/support/refunds are both named refunds), and a nested skill may share its name with a top-level one. Hosts MUST NOT assume name uniqueness.

packages/core/src/server/skills.ts
```diff
@@ -31,11 +30,16 @@ const SkillFrontmatterSchema = z
...
 
+export interface SkillResource {
+  uri: string;
+  digest: string;
+  content: string;
+}
+
 export interface Skill {
   name: string;
   frontmatter: Record<string, unknown>;
```
fredericbarthelet:
```suggestion
  uri: string;
  frontmatter: { name: string; description: string;} & Record<string, unknown>;
```
name isn't surfaced outside of the frontmatter anymore as per skills/list and skills/get method. URI is on the other hand

fredericbarthelet (approved):
2 final small comments, but otherwise good to go :) Thanks @qchuchu

PR #1037, opened 2026-08-03
feat: add ecom guide

paulleseute:
and order guide so user-facing order is lexicographical

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -72,11 +72,12 @@
             "root": "guides/index",
             "pages": [
               "guides/csp",
-              "guides/ux",
+              "guides/ecommerce",
               "guides/files",
-              "guides/skills",
               "guides/auth-providers",
-              "guides/migrate"
+              "guides/migrate",
+              "guides/skills",
+              "guides/ux"
             ],
             "expanded": true
           },
--- a/docs/guides/ecommerce.mdx
+++ b/docs/guides/ecommerce.mdx
@@ -1,0 +1,125 @@
+---
+title: "Build an Ecommerce App"
+description: "Scaffold a catalog search and carousel UI, then fill it with your data and brand"
+sidebarTitle: "Ecommerce"
+icon: "shopping-bag"
+---
+
+Ecommerce inside a conversation is a hard UX problem. A storefront's job has to fit the host-controlled frame, and merchandising moves to a model.
+
+The `ecom` template ships that architecture as a skeleton, with opinionated patterns refined by the [Alpic build team](https://alpic.ai/solutions/build/studio): an iterative search loop, an inline carousel opening into a fullscreen product detail, sparse variant selection, and a token-based design system. The wiring is in place, the data is not: a coding agent skill fills in your catalog and brand, pausing for your sign-off at each design decision.
+
+## Scaffold the Template
+
+Scaffold a project with the `--ecom` flag:
+
+```bash
+npx skybridge create my-shop --ecom
+```
+
+The template works best with the `chatgpt-app-builder` skill, which drives the fill workflow below. The scaffolder installs it by default; to add it to an existing project:
+
+```bash
+npx skills add alpic-ai/skybridge --skill chatgpt-app-builder
+```
+
+## Install the Tooling
+
+A few MCP servers let the agent do work it would otherwise hand back to you:
+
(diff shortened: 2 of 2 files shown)


docs/guides/ecommerce.mdx
```diff
@@ -0,0 +1,125 @@
+---
+title: "Build an Ecommerce App"
+description: "Scaffold a catalog search and carousel UI, then fill it with your data and brand"
+sidebarTitle: "Ecommerce"
+icon: "shopping-bag"
+---
+
+Ecommerce inside a conversation is a hard UX problem. A storefront's job has to fit the host-controlled frame, and merchandising moves to a model.
+
+The `ecom` template ships that architecture as a skeleton, with opinionated patterns refined by the [Alpic build team](https://alpic.ai/solutions/build/studio): an iterative search loop, an inline carousel opening into a fullscreen product detail, sparse variant selection, and a token-based design system. The wiring is in place, the data is not: a coding agent skill fills in your catalog and brand, pausing for your sign-off at each design decision.
```
fredericbarthelet, line 10: Maybe worth adding a small "prerequistes" or a "who is this template for" section specifically for this guide. It might be good mentioning somewhere that this template is made for anyone with an existing headless ecomerce engine (or something that resemble it like a CSV). Worth mentioning a few of them (like Shopify, Prestashop, Medusa,... to help on SEO too)

docs/guides/ecommerce.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+```bash
+npx skills add alpic-ai/skybridge --skill chatgpt-app-builder
+```
+
+## Install the Tooling
+
+A few MCP servers let the agent do work it would otherwise hand back to you:
+
+- **[Chrome DevTools MCP](https://github.com/ChromeDevTools/chrome-devtools-mcp)**: a browser for the agent: it [drives the local DevTools](/test/devtools#drive-devtools-from-a-coding-agent) to verify its own work, previews UI components, and inspects your live site's styles for token extraction.
```
fredericbarthelet, line 32: isn't this not related at all specifically to this guide ?

docs/guides/ecommerce.mdx
```diff
@@ -0,0 +1,125 @@
...
+- **[Playwright MCP](https://github.com/microsoft/playwright-mcp)**: the same, as a fallback when Chrome DevTools is unavailable.
+- **[Figma MCP](https://www.figma.com/mcp-catalog/):** if your brand lives in Figma, the agent extracts color ramps, type scale, and spacing itself.
+
+All are optional: without them, the agent asks you instead.
+
+## What the Template Ships
+
+Four patterns carry the app. The fill workflow customizes their values, never their shape.
+
+### Search, Then Render
```
fredericbarthelet, line 42: in addition to "what", a little bit of "why" here would be welcome. It's not obvious why it's recommended to split search an render.
I'd put forward a big benefit (and backreference OpenAI documentation on that point to give credibility to the implementation too: https://developers.openai.com/plugins/build/chatgpt-ui#separate-data-processing-from-ui-rendering): it allows the model to apply its intelligence to data it fetched before rendering UI to the user. Doing so it can perform multiple searches to best reflect user intent that can't be expressed through faceted search properties. It can also use previous conversation user knowledge to filter out specific item.

On a side note, it also allows hosts that don't support the app extension to still use the MCP server search feature

docs/docs.json
```diff
@@ -72,11 +72,12 @@
...
-              "guides/ux",
+              "guides/ecommerce",
               "guides/files",
-              "guides/skills",
               "guides/auth-providers",
-              "guides/migrate"
+              "guides/migrate",
+              "guides/skills",
+              "guides/ux"
             ],
```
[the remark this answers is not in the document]
fredericbarthelet: Do we want to put guides in alphabetical order? I'd argue some guides are more important than others. Might want to stick with the existing order.

fredericbarthelet (approved):
A few comments to improve, but a good base already :)

docs/docs.json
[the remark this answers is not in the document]
paulleseute: i'd agree but the current order feels very right to me and also importance is a tad subjective, so since we have a few items i figure lexicographical was the neutral choice. no strong opinion tho
Screenshot 2026-08-04 at 09 06 58

docs/guides/ecommerce.mdx
paulleseute: yes, but i think this feature is exotic enough to deserve a oneliner explanation. there's also a reference to the actual doc with details and setup

PR #1040, opened 2026-08-04
DevTools: warn on large tool output and view state

thleonard:
## Summary

- surface warning badges and scoped inline alerts for large Tool output and View state
- warn at 5,000 estimated tool-output tokens and 20,000 estimated view-state tokens
- count model-visible tool content and structured content, plus persisted view state and data-llm context
- align browser and server console warnings while keeping the full payload inspectable

## Threshold rationale

We use a 200k-token Haiku context window as the baseline.

These thresholds are advisory, not hard limits. They are meant to prompt developers to review how their app uses model context; exceeding them may be perfectly reasonable depending on the app. DevTools does not block or truncate anything.

- **Tool output — 5k tokens (2.5%)**: one tool call at this size may be fine, but tool responses remain in the conversation and accumulate when an app makes multiple tool calls. The lower threshold warns before repeated calls rapidly consume the context window.
- **View state — 20k tokens (10%)**: view state gets a larger allowance because it does not accumulate across new rounds of inference. The latest state is sent to the model, so 20k represents a meaningful but bounded share of the context window.

Images taken with a 1 token threshold to trigger them:
image
image
image

### DIFF
--- a/packages/core/src/context-warnings.test.ts
+++ b/packages/core/src/context-warnings.test.ts
@@ -1,0 +1,46 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+  estimateContextTokens,
+  TOOL_OUTPUT_WARNING_TOKENS,
+  warnOnLargeToolOutput,
+} from "./context-warnings.js";
+
+describe("context warnings", () => {
+  it("estimates UTF-8 bytes and returns zero for no content", () => {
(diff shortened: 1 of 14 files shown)


harijoe (approved):
The duplication introduced doesn't feel good but I don't think we can get rid of it without other drawbacks, so okay.

PR #1058, opened 2026-08-20
feat(test): evals

harijoe:
### Summary

- Adds `@skybridge/test`: eval runner answering whether a prompt reaches the right tool with the right arguments.
- Enabled via `skybridge({ evals })`; scenarios use vitest's own `it`/`expect` plus `start<AppType>()`.
- The model is a `provider/model-id` string, and only the provider named gets imported.
- Anthropic, OpenAI, Mistral and Google are optional peers; each reads its own API key variable.
- For anything those prefixes cannot reach, `defineEvalModel()` registers a model from a setup file.
- Extracts core's Vite plugin to `@skybridge/vite-plugin`.
- **Breaking**: `skybridge/vite` is removed; import `skybridge` from `@skybridge/vite-plugin` instead.
- `server.run()` now reports its bound port over IPC, letting the runner use an ephemeral port with no bind race.
- Repeated runs with an acceptance threshold are deferred to SKY-551.

Implements SKY-546. API validated on the POC (#1052) and the design review thread.


harijoe:
@greptile-apps

packages/vite-plugin/src/plugin.ts
```diff
@@ -61,14 +90,46 @@ function getViewEntryPattern(viewsDir: string): RegExp {
...
  */
-export function skybridge(options?: SkybridgePluginOptions): Plugin {
+export function skybridge(
+  options?: SkybridgePluginOptions,
+): [Plugin, ...PluginOption[]] {
+  const plugins: [Plugin, ...PluginOption[]] = [viewsPlugin(options)];
+  const evals = options?.evals;
+  if (evals !== undefined) {
+    plugins.push(loadEvalsPlugin(evals));
+  }
```
fredericbarthelet: I don't think it's a good idea to have polymorphic plugins and using skybridge/vite-plugin should "always" package both views folder for build and prepare server for unit test.
Happy to discuss split of concerns between skybridge/test and this package if you want

packages/test/src/plugin.ts
```diff
@@ -0,0 +1,29 @@
...
+  return fileURLToPath(new URL(file, import.meta.url));
+}
+
+/**
+ * Wires the eval runner into vitest. Usually reached through the `evals`
+ * option of the `skybridge()` Vite plugin rather than registered directly.
+ * Everything the runner needs is configured here, so a scenario file imports
+ * nothing but `start` and its own `AppType`.
+ */
+export function evals(options: EvalsOptions): Plugin {
```
fredericbarthelet: why not put this directly in `@skybridge/vite-plugin` ?

packages/vite-plugin/src/evals/global-setup.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (options.server !== undefined) {
+    project.provide("skybridgeEvalsUrl", options.server);
+    return;
+  }
+  if (options.project === undefined) {
+    throw new Error("The evals plugin needs either `server` or `project`");
+  }
+
+  const server = await startServer(options.project);
```
fredericbarthelet, line 19: this should be the responsability of the `start` method instead to avoid booting the server on test suits that don't need it

packages/test/src/chat.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  static async open<App>(
+    serverUrl: string,
+    host: HostConfig,
+  ): Promise<Chat<App>> {
+    const client = new Client({ name: "skybridge-eval", version: "0" });
+    const transport = new StreamableHTTPClientTransport(new URL(serverUrl));
+    await client.connect(transport);
+    const chat = new Chat<App>(client, transport, host);
```
fredericbarthelet, line 58: wild thoughts: use STDIO transport to communicate with the server in test mode: faster to start, controlled via the command to run it itself. Don't know the rest of the implication, but just wondering if it might help

packages/test/src/chat.ts
```diff
@@ -0,0 +1,169 @@
...
+              this.failures.set(
+                toolCallId,
+                `the call never reached the server: ${error instanceof Error ? error.message : String(error)}`,
+              );
+              throw error;
+            }
+            if (result.isError === true) {
+              this.failures.set(
+                toolCallId,
+                `the tool returned an error: ${JSON.stringify(result.content)}`,
```
fredericbarthelet, line 146: you might want to have a helper to assert a tool returned an error

packages/vite-plugin/src/plugin.ts
```diff
@@ -23,6 +43,15 @@ export interface SkybridgePluginOptions {
...
   serverExternal?: string[];
+  /**
+   * Wires the eval runner into vitest: contributes `setupFiles`,
+   * `globalSetup`, `provide` and `testTimeout` from `@skybridge/test`, loaded
+   * lazily so production builds never touch it. Set it only in the config
+   * vitest uses for evals; when omitted, no test config is contributed and
+   * ordinary unit tests are unaffected. Requires `@skybridge/test` as a dev
+   * dependency.
+   */
+  evals?: EvalsOptions;
```
fredericbarthelet, line 50: might want to stick with "test" name in order to match the package name (and the expectation of configuring vitest), WDYT ?

packages/vite-plugin/src/plugin.ts
```diff
@@ -23,6 +43,15 @@ export interface SkybridgePluginOptions {
    * typically optional native modules pulled in by a transitive dependency.
    */
   serverExternal?: string[];
+  /**
+   * Wires the eval runner into vitest: contributes `setupFiles`,
+   * `globalSetup`, `provide` and `testTimeout` from `@skybridge/test`, loaded
+   * lazily so production builds never touch it. Set it only in the config
+   * vitest uses for evals; when omitted, no test config is contributed and
+   * ordinary unit tests are unaffected. Requires `@skybridge/test` as a dev
+   * dependency.
```
fredericbarthelet: this feels more like a claude comment that something to help end user understand what this config option will allow him to configure

packages/vite-plugin/README.md
```diff
@@ -0,0 +1,18 @@
+# @skybridge/vite-plugin
+
+The Vite plugin behind [Skybridge](https://skybridge.tech): it discovers view
+modules, serves them as virtual entries, applies the `data-llm` transform, and
+tells `skybridge build` which packages to leave unbundled.
```
fredericbarthelet, line 5: it also extend vitest assertions with chat-oriented APIs to assert various scenario occurring chances

fredericbarthelet (commented):
a few comments, I think you can greatly reduce complexity by avoiding plugin definition polymorphism based on config options and by starting the server in the start chat method itself

packages/vite-plugin/src/evals/global-setup.ts
harijoe: The thing is, moving the server booting the `start` method means spawning a new server per worker. I agree the server shouldn't start for test suites that don't need it though — probably via a config somewhere?

packages/vite-plugin/src/plugin.ts
harijoe: I'm not sure about this one. "test" name is here because today it contains evals but is planned to contain other test-related stuff such as frontend testing tools. 

The evals key here is exposing options only relevant for evals (model, temperature…), I feel it would be confusing to rename it to test.

Also when other testing tools land, we could add a new key specific to their configuration next to evals? It doesn't feel bad I think.

packages/test/src/chat.ts
harijoe: It feels a bit weird to choose not to be iso with what the test is trying to test. To discuss!

.github/workflows/publish.yml
```diff
@@ -53,16 +53,19 @@ jobs:
             fi
 
             echo "version=$VERSION" >> "$GITHUB_OUTPUT"
+            echo "beta_version=$VERSION-beta" >> "$GITHUB_OUTPUT"
```
fredericbarthelet: don't forget to remove those lines before merging :)

.github/workflows/publish.yml
```diff
@@ -79,25 +86,39 @@ jobs:
...
+        run: |
+          pnpm pkg set 'dependencies.skybridge=^${{ steps.version.outputs.version }}'
+          pnpm pkg set 'devDependencies.@skybridge/vite-plugin=^${{ steps.version.outputs.version }}'
 
       - name: Pin ecom template skybridge dep for publish
         working-directory: ./packages/create-skybridge/templates/ecom
-        run: pnpm pkg set 'dependencies.skybridge=^${{ steps.version.outputs.version }}'
+        run: |
+          pnpm pkg set 'dependencies.skybridge=^${{ steps.version.outputs.version }}'
+          pnpm pkg set 'devDependencies.@skybridge/vite-plugin=^${{ steps.version.outputs.version }}'
```
fredericbarthelet, line 118: I think we should also include @skybridge/test by default on starters (or only some of them, like ecom and demo, blank might still want to be as minimal as possible)

packages/core/src/server/server.ts
```diff
@@ -1051,6 +1055,9 @@ export class McpServer<
       });
     });
 
+    const { port: boundPort } = httpServer.address() as AddressInfo;
```
fredericbarthelet: how could `port` (that was already used to serve) be different from `boundPort` ?
If `httpServer.listen(port)` is best effort, then maybe instead rename original `port` to `intendedPort` ?

packages/core/package.json
```diff
@@ -27,9 +27,9 @@
       "types": "./dist/web/index.d.ts",
       "default": "./dist/web/index.js"
     },
-    "./vite": {
-      "types": "./dist/web/plugin/plugin.d.ts",
-      "default": "./dist/web/plugin/plugin.js"
+    "./views": {
```
fredericbarthelet, line 30: do you really need to have view index accessible from outside this package now that plugin is in a separate package ?

fredericbarthelet: I mean, I understand why you did it, it's in use in the plugin, but I'm wondering if this is a smell of something fishy that these utils are both used directly by build and dev commands as well as the vite plugin itself

packages/test/src/model.ts
fredericbarthelet, line 1: I think you can remove this file and simply expect a `LanguageModel ` from the AI SDK in the `model` prop from your chat method. What's the added value of this provider processor?

packages/test/src/chat.ts
```diff
@@ -0,0 +1,151 @@
...
+      chat.tools = tools;
+    } catch (error) {
+      await transport.close();
+      throw error;
+    }
+    return chat;
+  }
+
+  /** The tool definitions the model was looking at, for failure reports. */
+  get toolDefinitions(): ToolDefinition[] {
```
fredericbarthelet: why is this part of the public API, I feel like assertions on the tool definition is like asserting that an exact line of code is written 😅

packages/test/src/chat.ts
```diff
@@ -0,0 +1,151 @@
...
+    return this.tools;
+  }
+
+  /** Takes a turn and returns the model's answer. */
+  async send(prompt: string): Promise<string> {
```
fredericbarthelet, line 70: returning only the last turn text response might be a bit restrictive : it's probably better to have assertion helpers that can dig in all properties of the assistant messages to test for the presence of a response that providing it in the public API and letting people run their own assertion on the string content, WDYT?

packages/test/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "url": "https://github.com/alpic-ai/skybridge.git"
+  },
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./vitest-config": {
+      "types": "./dist/vitest-config.d.ts",
+      "default": "./dist/vitest-config.js"
```
fredericbarthelet: this is not used anymore

packages/test/package.json
```diff
@@ -0,0 +1,75 @@
...
+    },
+    "@ai-sdk/openai": {
+      "optional": true
+    }
+  },
+  "devDependencies": {
+    "@ai-sdk/anthropic": "^2.0.0",
+    "@ai-sdk/google": "^2.0.0",
+    "@ai-sdk/mistral": "^2.0.0",
+    "@ai-sdk/openai": "^2.0.0",
```
fredericbarthelet: you need all 4?

packages/test/package.json
```diff
@@ -0,0 +1,75 @@
...
+    },
+    "@ai-sdk/google": {
+      "optional": true
+    },
+    "@ai-sdk/mistral": {
+      "optional": true
+    },
+    "@ai-sdk/openai": {
+      "optional": true
+    }
```
fredericbarthelet: if you switch to having a model being a `LanguageModel `,  you only need a peer dep to the core ai sdk package

fredericbarthelet: The ecosystem convention is peer + dev on ai, and no @ai-sdk/{openai,anthropic,…} at all.
[Evalite](https://v1.evalite.dev/api/ai-sdk) uses LanguageModel interface from ai as input for their instrumentation

fredericbarthelet: [Eve](https://eve.dev/docs/agent-config) does that too

packages/test/README.md
```diff
@@ -0,0 +1,90 @@
...
+| --- | --- | --- |
+| `anthropic` | `@ai-sdk/anthropic` | `ANTHROPIC_API_KEY` |
+| `openai` | `@ai-sdk/openai` | `OPENAI_API_KEY` |
+| `mistral` | `@ai-sdk/mistral` | `MISTRAL_API_KEY` |
+| `google` | `@ai-sdk/google` | `GOOGLE_GENERATIVE_AI_API_KEY` |
+
+```ts
+skybridge({
+  evals: {
+    model: "anthropic/claude-sonnet-4-5",
```
fredericbarthelet: this notation is usually reserved for people using the vercel ai gateway, it can be also supported in the project if you don't provide a `LanguageModel`

packages/test/README.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+`command` has to start the server process directly, since wrappers such as
+`pnpm run` do not forward the channel the runner uses to learn the port.
+
+Any other prefix is passed to the AI SDK untouched, so a string such as
+`openai/gpt-5` also resolves through the Vercel AI Gateway if you have
+`AI_GATEWAY_API_KEY` set and no provider package installed.
+
+## Models the prefixes cannot reach
```
fredericbarthelet: this use case is supported by ai sdk custom provider

packages/vite-plugin/src/evals/global-setup.ts
fredericbarthelet: Discussed together, we keep server startup in the setup script of the plugin for now, and see how it affects vitest run performances that are not using chat assertions

fredericbarthelet (commented):
A few comments, but it starts to take shape :)
Also, could you throw a first version of documentation for this new package in this PR?

.github/workflows/publish.yml
harijoe: I agree, but in a follow-up PR maybe?

packages/test/src/model.ts
harijoe: Just the sugar as we intended to make it as easy as possible to use models

packages/test/src/chat.ts
fredericbarthelet: you agreed and resolved, but didn't change the API 😅

fredericbarthelet (approved):
LGTM :) Awesome work @harijoe !

Issue #1062, opened 2026-08-25
Devtools installs the Apps SDK mock even in the Claude preview


harijoe:
`packages/devtools/src/components/layout/tool-panel/view/index.tsx` calls both `createAndInjectOpenAi` and `createMcpHostMock` on every view mount, with no dependency on `previewClient`. A view previewed as "Claude" therefore still sees a defined `window.openai` and takes the Apps SDK code path, not the MCP Apps one a real Claude host would give it.

Consequences we already hit:

- `HostAdaptor` picks the Apps SDK overlay stores over its MCP Apps polyfills whenever `window.openai` is defined. `closeModal()` deliberately no-ops in that case ("host owns modal lifecycle"), so the modal polyfill's backdrop-click and Escape handlers in `modal-provider.tsx` are dead in the preview even though they work against a real Claude host. This is why SKY-557's "cannot close the modal" reproduced on *both* previews.
- More generally, any hook that branches on `this.openai` is exercised through the wrong bridge in the Claude preview, so the preview can pass while the real MCP Apps path is broken (or vice versa).

Proposed fix: install the openai mock only when the previewed host is an Apps SDK one, so the Claude preview runs the ext-apps path end to end.

Worth its own QA pass: it changes which bridge every hook takes in that preview, so the blast radius is wider than the modal.

Found while working on SKY-557 (QA feedbacks on realistic UI preview).

dawNotPoi:
I am working on this. I will keep the host-specific mock selection focused and add regression coverage for the Claude preview taking the MCP Apps path.

harijoe:
Thanks @dawNotPoi. Could you introduce yourself on [our Discord](https://discord.com/invite/awV4gu74wK) as well? We'd love to know more about our contributors.

MatLBS:
Hi @harijoe, is this issue still relevant ? I would like to work on it if possible

PR #1069, opened 2026-08-26
feat(devtools): mimic more host chrome in the conversation preview

harijoe:
Stacked on #1063. Covers the QA requests from the [realistic UI session](https://app.notion.com/p/3c31e5ce535b80788943ef8cb1832efd) that #1063 deliberately left out, all of them host-chrome fidelity rather than breakage. Tracked in SKY-557.

## What it adds

- **Host logos in the switch button.** The preview toolbar showed a generic swap arrow; it now shows the logo of the host you are switching to. ChatGPT reuses the existing `BlossomIcon`; Claude needed a compact logomark, which did not exist in the repo (only the wordmark), so `ClaudeLogomark` is a new geometric burst. **Swap it for the official asset if it reads off-brand.**
- **Composer in fullscreen.** `chatgpt-shell` hid the composer entirely in fullscreen, and `claude-shell` only skipped it on mobile. Both now float it over the bottom of the widget, which is the shape Claude desktop already used here and what the real apps do.
- **Claude's scroll-to-bottom arrow.** Appears once the thread is scrolled more than 24px from the bottom, centered above the composer, and it can overlap the widget exactly as QA observed. Clicking it smooth-scrolls to the bottom. It tears down in fullscreen so it cannot fight the pin-to-bottom effect from #1063.

## How to test

Run any tool in the everything example, then click **preview**.

- Switch clients and check both logos.
- Go fullscreen in each shell, on desktop and mobile, and confirm the composer floats at the bottom. It covers the bottom of the widget, which matches the real hosts.
- Scroll the Claude thread up: the arrow appears above the composer, over the widget. Click it to return to the bottom.

Verified in both shells, light and dark, desktop and mobile, across inline, pip and fullscreen. No tests: this is preview chrome, and an e2e test asserting on shell markup would cost more than it protects.

## Not included

An earlier revision of this branch also rendered ChatGPT's open-in-app pill above the widget, driven by the `openInAppUrl` the store already tracks. Removed on review; that data is still stored and still displayed nowhere.

ChatGPT's dedicated open-external confirmation modal, the last request in the QA batch, is deliberately left for its own PR: it changes the behaviour of a call views depend on (`useOpenExternal`), rather than adding static chrome, and it should be built after #1062 / SKY-558 so it targets the bridge each preview should actually be running.

ChatGPT has a scroll arrow too; QA asked only for Claude's, so only Claude's is here.

### DIFF
--- a/packages/devtools/src/components/layout/preview/chatgpt-shell.tsx
+++ b/packages/devtools/src/components/layout/preview/chatgpt-shell.tsx
@@ -408,9 +408,11 @@ export function ChatgptShell({ children }: { children: ReactNode }) {
         <div
           aria-hidden
           className={cn(
-            "w-full shrink-0 cursor-not-allowed px-4",
+            "w-full cursor-not-allowed px-4",
             isMobile ? "pb-4" : "pb-6",
-            isFullscreen && "hidden",
+            isFullscreen
+              ? "pointer-events-none absolute inset-x-0 bottom-0 z-30"
+              : "shrink-0",
           )}
         >
           <div className="mx-auto w-full max-w-[768px]">
--- a/packages/devtools/src/components/layout/preview/claude-icons.tsx
+++ b/packages/devtools/src/components/layout/preview/claude-icons.tsx
@@ -54,6 +54,41 @@ export function ChevronDownIcon(props: IconProps) {
   );
 }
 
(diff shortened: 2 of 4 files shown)


harijoe:
Folded into #1063 as a single commit instead of a stacked PR.

PR #1070, opened 2026-08-27
feat!: replace the registrar with a user-defined factory via the Skybridge app class

harijoe:
> [!NOTE]
> Proposal stacked on the SKY-546 evals branch (the PR base) so the eval layer is part of the story; when that branch merges into `v2`, this retargets with one command. Docs and skills stay out of scope.

### Summary

- Replaces the registration ledger with a user-written factory.
- Apps are now defined as `new Skybridge(config, setup)` in a pure, import-safe `src/server.ts`.
- The setup callback receives a fresh `McpServer` per request and returns the chained registrations, `mcpMiddleware` included.
- `src/index.ts` becomes the hardcoded entry point (`export default await app.run()`).
- `app.fetchHandler` exposes the in-process fetch handler, so apps are testable without Express or sockets.
- Evals run in-process only: `start({ app, model })` — no child process, no port, no server config.
- Auth-sensitive evals inject a claimed identity via `authInfo`; real scope enforcement and in-band challenges run.
- Working example: `examples/capitals/evals/explore.eval.ts` (`pnpm evals`).
- CLI, templates, all 22 examples, core test suite and devtools e2e fixture migrated.
- Fixes two latent `expect.chat` bugs (augmentation lost in `.d.ts` emit; `App` inference) that bite the evals branch on its own.

### Architecture Notes

- **`McpServer` is registration sugar again** — it extends the SDK class directly; run/express/oauth moved to `Skybridge` (new `app.ts`), since a per-request instance cannot own app-level state. `mcpMiddleware` chains on the server inside the callback, so no registration lock exists.
- **Setup runs once at construction plus once per request** — the construction pass fail-fasts registration errors and feeds OAuth security-scheme introspection; user factories must be side-effect-free.
- **Manifest priming is no longer consume-once** — every per-request instance needs the build/skills manifests; the Vite manifest disk read is cached per instance.
- **Tool typing is unchanged for views** — `Skybridge` carries the same `$types` marker, so `generateHelpers<AppType>` works with `AppType = typeof app` and no web-side changes were needed.
- **In-process evals skip only token verification** — per-tool enforcement reads `ctx.http.authInfo` in the MCP layer, so injected identities exercise the real authorization path; `@skybridge/test` stays structurally decoupled from `skybridge`.
- **`skybridge:listening` IPC stays** — the eval harness that consumed it is gone; kept because it is generic supervisor plumbing from the base branch. Open question for SKY-546 whether to keep or drop it.

### Alternatives Considered

- **Keep the registrar (status quo on `v2`)** — it fakes the SDK's per-request factory contract by recording registrations as thunks and replaying them onto fresh instances. That machinery has a documented blind spot (post-registration mutations like `registered.disable()` silently don't propagate) and adds a replay layer users can't see. A real user factory is the SDK's own recommended shape and deletes the machinery outright.
- **Declarative config bag (`app(config, { tools: [...] })`) instead of a callback** — separating each tool's config from its handler breaks the per-tool type inference that `AppType` and the typed web hooks rely on, and views/resources/prompts/skills would sprawl the bag. The callback keeps the SDK's own factory shape and full chaining, at the cost of one lambda.
- **Framework-generated entry instead of `src/index.ts`** — the CLI could synthesize the entry and have userland export only the app. That hides the lifecycle completely and breaks the Vercel path, where the deployed module's default export must be the Express app; a visible two-line `index.ts` keeps that contract explicit.

Tracked in SKY-514 (v2).


packages/core/src/server/app.ts
```diff
@@ -0,0 +1,481 @@
...
+    if (typeof pathOrHandler === "string") {
+      this.errorMiddleware.push({ path: pathOrHandler, handlers });
+    } else {
+      this.errorMiddleware.push({ handlers: [pathOrHandler, ...handlers] });
+    }
+    return this;
+  }
+
+  /** Register MCP protocol-level middleware (catch-all). */
+  mcpMiddleware(handler: McpMiddlewareFn<TAuthExtra>): this;
```
fredericbarthelet: shouldn't this be on the McpServer class instead? Chances are the definition of tool and of MCP method handlers should be done (from an API point of view) in the same place in userland code

packages/core/src/server/app.ts
```diff
@@ -0,0 +1,481 @@
...
+   * serverless function entry can call it as a `(req, res)` handler. On
+   * Node, returns `undefined` once listening. When the process was spawned
+   * with an IPC channel, the bound port is reported to the parent as
+   * `{ type: "skybridge:listening", port }`, the readiness signal test runners
+   * wait on instead of polling.
+   */
+  async run(): Promise<
+    { fetch: (...args: unknown[]) => unknown } | Express | undefined
+  > {
+    this.middlewareLocked = true;
```
fredericbarthelet: you should be able to remove this lock if mcp middleware are moved to the McpServer class

packages/core/src/server/app.ts
```diff
@@ -0,0 +1,481 @@
...
+    if (process.env.VERCEL === "1") {
+      // createApp only reads httpServer inside its dev-only branch
+      // (viewsDevServer); under VERCEL=1 + NODE_ENV=production it's a
+      // bare object passed to satisfy the required parameter.
+      const httpServer = http.createServer();
+      await createApp({
```
fredericbarthelet, line 263: does it still make sense to split express and app ?

packages/test/README.md
```diff
@@ -23,6 +23,53 @@ it("finds products by category", async () => {
...
+  });
+});
+```
+
+Prefer it whenever you can: there is no child process and no plugin config, so
+it runs anywhere vitest does. The caveat is that it bypasses the Express layer,
+so nothing verifies a bearer token; pass `authInfo` to claim an identity, or use
+the process mode above when the evals must go through real tokens.
+
+### Authenticated evals
```
fredericbarthelet: nice

packages/test/src/session-registry.ts
```diff
@@ -3,35 +3,119 @@ import { inject, onTestFinished } from "vitest";
...
     model: options.model,
     temperature: options.temperature ?? config?.temperature,
     systemPrompt: options.systemPrompt ?? config?.systemPrompt,
     maxSteps: options.maxSteps ?? config?.maxSteps,
-  });
+  };
+
+  const chat = await (options.app === undefined
+    ? Chat.open<App>(urlFromPlugin(), host)
+    : inProcessChat<App>(options.app, host, options.authInfo));
```
fredericbarthelet: you'd still offer both in-memory and http path? What would be the difference between the 2 making you want to switch from one to the other?

examples/auth-auth0/src/server.ts
```diff
@@ -37,91 +35,91 @@ const server = new McpServer(
...
-      },
-      view: {
-        component: "search-coffee-paris",
-        description: "Search for coffee shops in Paris",
-        csp: {
-          resourceDomains: ["https://images.unsplash.com"],
+  (server) =>
+    server.registerTool(
+      {
+        name: "search-coffee-paris",
```
thleonard: I find it weird to have a function here. I liked the `app().registerTool()` shape better. I believe it's also the same shape as the typescript SDK which is a + to me.

If I understand correctly it's now: `app({...}, (server) => server.registerTool())`? The function shape is weird to me. If you want to move the tools definition inside the Server constructor then I'd maybe have something like that instead: `app({...}, { tools: [{...}, {...}], skills: [ {...} ]})`

I also liked that we could chain the middlewares with it: `app().registerTool().mcpMiddleware()`

thleonard: Is there a benefit that I don't understand right now?

harijoe: Well, the v2 of the official SDK requires a factory function instead of a singleton. The reason is that now, your server is being reinstanciated with every requests, this is by design. I agree that from a user perspective it's worse than before but we don't really have the choice here.

packages/test/src/session-registry.ts
harijoe: That's an artifact, I think we should only provide in-memory as long as nothing contradicts it.

packages/core/src/server/app.ts
harijoe: You mean you're challenging the existence of express.ts itself? Maybe not, no strong conviction on this

packages/core/src/server/app.ts
harijoe: image
Haha, okay let's do it

PR #1076, opened 2026-09-03
fix(cli): report a crashed dev server child instead of showing it as healthy

### LINKED ISSUE #1075: skybridge dev does not report a crashed server child
Reported by an external user (Amey Pawar) building an MCP App on Skybridge 1.4.1.

## Problem

`skybridge dev` restarts the server child on file change, but it has no notion of the child having crashed. When the server dies at boot, the parent process and the tsc watcher stay alive and the dev UI keeps looking healthy, while nothing is listening on the port.

In the reporter's case the crash came from the documented `oauth: await ...Provider()` shape, which throws before `run()` when the IdP is unreachable. `pgrep -f 'skybridge dev'` reported a running dev server, `:3000` was dead, and no error was visible in the UI. It took an hour and left three orphaned parent processes behind. Touching any `.ts` under `src/` revives it, which they only found by accident.

## Cause

nodemon emits a `crash` event when the child exits nonzero, then sits idle waiting for a file change. `packages/core/src/cli/use-nodemon.ts` only wires `readable` and `restart`, so `crash` is dropped and neither the Ink dev UI nor `--plain` ever learns the server is down. A boot-time throw can also happen before nodemon's stdout/stderr streams become readable, so even the stack trace can be lost.

## Proposed fix

- Add `onCrash` to `NodemonHandlers`, wire `nodemon.on("crash", ...)`, and declare the event in `packages/core/src/cli/nodemon.d.ts`.
- Ink dev UI: surface a crash message and hold a "down" state so the status line stops implying the server is up.
- `--plain` (`packages/core/src/cli/run-plain.ts`): same message on stderr.

Deliberately not auto-restarting on crash: nodemon does not hot-loop a process that dies at boot, and a restart loop would bury the actual stack trace. Visibility is the fix.

## Open questions

- Should the dev UI show a persistent "down" indicator, or is a log line enough?
- Should `skybridge dev` exit nonzero when the very first boot crashes, instead of waiting for a file change? That first-boot case is what cost the reporter the hour.

Reporter's repo: https://github.com/ameyypawar/pr-radar

harijoe:
Closes #1075

## Summary

- `skybridge dev` now knows when the server child has died. nodemon emits `crash` on a nonzero exit and then sits idle waiting for a file change; we only wired `readable` and `restart`, so the dev UI kept advertising a URL nothing was listening on.
- `NodemonHandlers` gains `onCrash`, wired to nodemon's `crash` event and declared in `nodemon.d.ts`.
- `useNodemon` returns a `crashed` flag, cleared on the next restart. While crashed, the Ink UI drops the `🏠 Running on …` and DevTools lines and shows `💥 Server crashed. Fix the error, then save a file under src/ to restart it.` in red.
- `--plain` prints the same sentence on stderr, from one shared `CRASH_MESSAGE`.

## Notes

The issue's claim that a boot-time stack trace can be lost does not hold: nodemon assigns its stdout/stderr and emits `readable` from its own `start` handler, before the child can write anything. No fix needed there, and the stack trace shows up in the logs as expected.

Two gaps left deliberately, both worth a follow-up only if they show up in practice:

- A child that exits with code 0 without listening leaves the server equally down, but nodemon emits `exit` rather than `crash`, so the UI does not flag it. Telling that apart from the `exit` nodemon emits while killing the child for a restart depends on an undocumented signal argument, and a code-0 exit is a much rarer shape than a throw.
- `crashed` clears on `restart`, so between saving a file and the new child booting the UI shows the URL again. That is the same behaviour as a healthy restart, where the URL never goes away.

## Verification

Manual, against `examples/capitals` with a top-level `throw` in `src/index.ts`: the banner replaces the URLs, the stack trace lands in the logs, and saving a file clears the banner and brings the URLs back. Same sentence on stderr with `--plain`.

No test added: the change is event wiring plus a boolean, and a test would be entirely a nodemon mock.

### DIFF
--- a/packages/core/src/cli/nodemon.d.ts
+++ b/packages/core/src/cli/nodemon.d.ts
@@ -7,4 +7,5 @@ export type ExtendedNodemon = import("nodemon").default & {
     event: "restart",
     listener: (files: string[]) => void,
   ): import("nodemon").Nodemon;
+  on(event: "crash", listener: () => void): import("nodemon").Nodemon;
 };
--- a/packages/core/src/cli/run-plain.ts
+++ b/packages/core/src/cli/run-plain.ts
@@ -1,5 +1,5 @@
 import type { TunnelManager, TunnelState } from "./tunnel.js";
-import { startNodemon } from "./use-nodemon.js";
+import { CRASH_MESSAGE, startNodemon } from "./use-nodemon.js";
 import { startTypeScriptCheck, type TsError } from "./use-typescript-check.js";
 
 export interface RunPlainOptions {
@@ -67,6 +67,7 @@ export function runPlain(options: RunPlainOptions): () => void {
     onStderr: (message) => info(message),
     onRestart: (files) =>
       info(`✓  Server restarted due to file changes: ${files.join(", ")}`),
+    onCrash: () => info(CRASH_MESSAGE),
   });
 
   let lastTsErrorKey = "";
--- a/packages/core/src/cli/use-nodemon.ts
+++ b/packages/core/src/cli/use-nodemon.ts
@@ -1,7 +1,7 @@
 import { existsSync } from "node:fs";
 import { createRequire } from "node:module";
 import { resolve } from "node:path";
-import { useEffect } from "react";
+import { useEffect, useState } from "react";
 import type { ExtendedNodemon } from "./nodemon.d.ts";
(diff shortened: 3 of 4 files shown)


harijoe:
@greptileai

qchuchu (approved):

PR #1079, opened 2026-09-03
chore: v2 release prep fixes for docs, skills, examples and templates

harijoe:
Repo-side fixes found while auditing the v2.0.0 draft release against the code.

- Examples: delete the 8 `examples/auth-*/nodemon.json`, which still ran `tsx src/server.ts`; on v2 that file only exports the app, so `skybridge dev` started nothing. The CLI default already runs `src/index.ts`.
- `@skybridge/test` is published on the `beta` dist-tag only, so docs and the skill say `@skybridge/test@beta`, and the two examples depending on it pin the `beta` dist-tag instead of a range. A `^2.0.0-beta.0` range resolves to the lexically highest prerelease sha, not the newest build.
- `skybridge/server` re-exports the type `StandardSchemaWithJSON`, which the docs and the release notes reference but users had no import path for.
- `docs/api-reference/register-tool.mdx`: the `extra` table now reflects the SDK v2 shape (`http?.authInfo`, `http?.req`, `mcpReq.signal`, `mcpReq._meta`).
- Delete `docs/fundamentals/` (two empty files) and point `AGENTS.md` at existing docs directories; align the Node floor in `AGENTS.md` and `CONTRIBUTING.md` with the templates (`>= 24.18.0`).
- `skills/.../migrate-to-v1.md`: the `@skybridge/vite-plugin` note applies guide-wide, the entry-file section describes the 2.x `src/server.ts` + `src/index.ts` split, and the version strategy lands on `2.0.0` then continues with the v2 release notes.
- Landing hero snippet uses `new Skybridge({ name, version, handler })`.
- `extra.authInfo` → `extra.http?.authInfo` in the authplane README and 4 TSDoc lines.
- Blank template gains `zod`, since `skybridge` peer-depends on it and a fresh project otherwise reports an unmet peer.

Second pass, from a fresh audit:

- Examples move `skybridge`, `@skybridge/vite-plugin` and `@skybridge/devtools` to `^2.0.0`, and the four peer ranges in `core`, `vite-plugin` and `test` do the same, so the 2.0.0 tarballs and the `v2.0.0` tag snapshot (which `create-skybridge --example` downloads) carry stable ranges. `@skybridge/devtools` joins the root `workspace:*` overrides so the monorepo links the local package like it already does for the other three; the examples were resolving a stale devtools beta from npm. Until 2.0.0 is published, external `npm install` of an example fails on `^2.0.0`, which is accepted since the release follows this merge.
- `scripts/bump.js` also rewrites `@skybridge/vite-plugin`; `publish.yml` pins the demo template's `@skybridge/test` to the exact version it just published instead of the floating `beta` tag.
- `skybridge/server` re-exports `inputRequired`, which the release notes point users at for the 2026-07-28 elicitation path.
- Dead docs links fixed in the root README (shipped as the npm README), the landing site, and the seven auth example READMEs; `AGENTS.md` names `@modelcontextprotocol/server`.

Root `prepare` now builds the devtools server entry (`tsc -p tsconfig.server.json`, no dependency on `skybridge` or the vite plugin) before `skybridge`, since `skybridge`'s build type-checks the workspace devtools once the override is in place and a fresh checkout has no `dist/` yet.

The CI steps were run locally as well: `pnpm install --frozen-lockfile` from a wiped `packages/*/dist`, so the whole `prepare` chain ran from scratch, then `pnpm test`, `pnpm build`, the landing build and the capitals example build, all green.


docs/test/evals.mdx
```diff
@@ -8,20 +8,20 @@ A tool that works is not the same as a tool the model decides to call. [DevTools
 
 ## Set Up
 
-Add the test runtime, vitest, the AI SDK, and the provider you want to drive the conversation:
+Add the test runtime (published on the `beta` dist-tag while the API settles), vitest, the AI SDK, and the provider you want to drive the conversation:
```
fredericbarthelet, line 11: but you removed the beta suffix on the publish.yml

examples/auth-descope/package.json
```diff
@@ -18,15 +18,15 @@
...
     "vite": "^8.1.5",
     "zod": "^4.4.3"
   },
   "devDependencies": {
     "@ai-sdk/anthropic": "^2.0.90",
-    "@skybridge/devtools": "^2.0.0-beta.0",
-    "@skybridge/test": "^2.0.0-beta.0",
-    "@skybridge/vite-plugin": "^2.0.0-beta.0",
+    "@skybridge/devtools": "^2.0.0",
+    "@skybridge/test": "beta",
```
fredericbarthelet, line 28: shouldn't this be 2.0.0 too?

skills/chatgpt-app-builder/references/migrate-to-v1.md
```diff
@@ -129,9 +129,9 @@ The old `include` and `@/*` path pointed at the split layout. Repoint both at `s
 
 If `include` still names the deleted `server/src` / `web/src`, `tsc` reports `No inputs were found` and silently typechecks nothing — a passing build that verified zero files.
 
-### 2.8 The server entry file must be `src/server.ts`
+### 2.8 The server entry file name matters
```
fredericbarthelet: not the case anymore, no? index.ts entry point is common enough that people don't need to be explained this I think

package.json
```diff
@@ -19,7 +19,7 @@
     "ai",
     "skybridge"
   ],
-  "author": "Frédéric Barthelet",
+  "author": "Fr\u00e9d\u00e9ric Barthelet",
```
fredericbarthelet: encoding issue 😂 ?
You can put your name here too @harijoe now :)

fredericbarthelet (approved):
few small comments but good to go otherwise :)

docs/test/evals.mdx
harijoe: It's a mistake, beta tag is still set in publish.yml and we will point to it in the different package.json

examples/auth-descope/package.json
harijoe: No we keep beta for now for skybridge/test

skills/chatgpt-app-builder/references/migrate-to-v1.md
harijoe: True but it was the case to migrate to v1. I'll delete this paragraph anyway, not useful anymore.

PR #916, opened 2026-07-06
feat(examples): add chatgpt-files

paulleseute:
a simple zip example to showcase how it works. object storage (r2) to actually store and serve zipped asset.

based on #877 by @raphaelboffa

showcase-chatgpt-files

PR #1082, opened 2026-09-05
fix(core): restore per-view Apps SDK opt-in for ChatGPT file APIs

### LINKED ISSUE #1074: Since 1.2.0 views always render as MCP Apps; ChatGPT file APIs are unreachable
Since #833 (v1.2.0) `registerViewResources` only emits `ui.resourceUri` and no longer registers the `text/html+skybridge` resource or `openai/outputTemplate`. The per-view `hosts` option went with it (`ViewHostType` is still exported but unused).

A view rendered through the MCP Apps path gets a `window.openai` compat shim without `uploadFile`, `getFileDownloadUrl` and `selectFiles`. That matches the ext-apps migration guide, which lists the file APIs as having no MCP Apps equivalent yet. So on 1.2.0+ `useFiles().getDownloadUrl` can no longer work on ChatGPT, and tools that use `openai/fileParams` lose access to attachments from the view.

Observed with 1.4.1 on the ChatGPT desktop app (rendered from `ui://views/ext-apps/...`, `window.skybridge.hostType === "mcp-app"`):

```
window.openai keys: appPackageName, callTool, displayMode, isSidebarOpen, locale, maxHeight, maxWidth,
  notifyIntrinsicWidth, openExternal, requestDisplayMode, safeArea, sendFollowUpMessage, setWidgetState,
  subjectId, surfaceBackgroundColor, theme, toolInput, toolOutput, toolResponseMetadata, userAgent, view,
  widget, widgetState
typeof window.openai.getFileDownloadUrl === "undefined"
```

`useFiles().getDownloadUrl` then throws `TypeError: this.openai.getFileDownloadUrl is not a function` (the adaptor only checks `this.openai`, unlike `selectFiles`, which checks the method too).

Two asks:

1. Bring back a per-view opt-in for the Apps SDK path (`hosts: ["apps-sdk"]` or equivalent) for tools that need the ChatGPT-only file APIs, until MCP Apps has a file API.
2. Have `getFileDownloadUrl` and `uploadFile` throw `NotSupportedError` when the method is missing on the host object, like `selectFiles` does.

We pinned to 1.1.2 in the meantime.

atirna:
Fixes #1074

## Summary

- Since #833, `registerViewResources` only emits the ext-apps resource and `ui.resourceUri`, so on ChatGPT every view renders as an MCP App. The `window.openai` compat shim on that path has no `uploadFile`, `getFileDownloadUrl` or `selectFiles`, leaving `useFiles` and `openai/fileParams` attachments unreachable there.
- `view.appsSdk: true` additionally registers the legacy `ui://views/apps-sdk/<component>.html` resource (`text/html+skybridge` with the `openai/*` content meta) and sets `openai/outputTemplate`, so ChatGPT renders the Apps SDK registration where the file APIs exist. `ui.resourceUri` stays canonical for MCP Apps hosts.
- The default is unchanged: with `appsSdk` omitted or `false`, there is one ext-apps resource and no `openai/outputTemplate`.
- `uploadFile` and `getFileDownloadUrl` now throw `NotSupportedError` naming the method when the host object lacks it, like `selectFiles` already did, instead of a `TypeError`. The vendored `window.openai` type marks both optional to match the shim.

The locked ext-apps spec has no upload or select request to proxy over, so the per-view Apps SDK opt-in is the available path for file APIs.

## Verification

On current `main`, before the patch:

- `view: { appsSdk: true }` is a type error.
- With the ChatGPT MCP Apps shim, `useFiles().getDownloadUrl({ fileId: "x" })` rejects with `TypeError: this.openai.getFileDownloadUrl is not a function`.

After:

- `pnpm --filter skybridge exec vitest run src/test/view.test.ts src/web/bridges/adaptor.test.ts` passes 36 tests, including the explicit `appsSdk: true` and `appsSdk: false` registration cases and missing-method guards.
- `pnpm --filter skybridge build` and `pnpm --filter skybridge test:format` pass.

### DIFF
--- a/docs/api-reference/register-tool.mdx
+++ b/docs/api-reference/register-tool.mdx
@@ -122,7 +122,7 @@ type ViewConfig = {
 | --- | --- |
 | `component` | The view's file name, type-checked against your views. |
 | `description` | Label the host may show during view discovery. |
-| `hosts` | Restrict where the view renders; defaults to all. |
+| `hosts` | Include `"apps-sdk"` to also register the legacy Apps SDK resource on ChatGPT — currently required for the [file APIs](/api-reference/use-files) and `openai/fileParams` attachments, which MCP Apps does not expose yet. Defaults to the MCP Apps registration alone. |
 | `prefersBorder` | Apps SDK only: request a visible border around the view. |
 | `domain` | Apps SDK only: override the served domain (advanced). |
 | [`csp`](/api-reference/register-tool#csp) | Per-view CSP overrides; see below. |
--- a/docs/api-reference/register-tool.mdx
+++ b/docs/api-reference/register-tool.mdx
@@ -110,7 +110,7 @@ Bind the tool to a React view to render its result instead of plain text. Each v
 type ViewConfig = {
   component: ViewName;
   description?: string;
-  hosts?: Array<"apps-sdk" | "mcp-app">;
+  appsSdk?: boolean;
   prefersBorder?: boolean;
   domain?: string;
   csp?: ViewCsp;
@@ -122,7 +122,7 @@ type ViewConfig = {
 | --- | --- |
 | `component` | The view's file name, type-checked against your views. |
 | `description` | Label the host may show during view discovery. |
-| `hosts` | Include `"apps-sdk"` to also register the legacy Apps SDK resource on ChatGPT — currently required for the [file APIs](/api-reference/use-files) and `openai/fileParams` attachments, which MCP Apps does not expose yet. Defaults to the MCP Apps registration alone. |
+| `appsSdk` | Also register the legacy Apps SDK resource on ChatGPT — currently required for the [file APIs](/api-reference/use-files) and `openai/fileParams` attachments, which MCP Apps does not expose yet. Defaults to the MCP Apps registration alone. |
 | `prefersBorder` | Apps SDK only: request a visible border around the view. |
 | `domain` | Apps SDK only: override the served domain (advanced). |
 | [`csp`](/api-reference/register-tool#csp) | Per-view CSP overrides; see below. |
--- a/docs/api-reference/use-files.mdx
+++ b/docs/api-reference/use-files.mdx
@@ -93,7 +93,7 @@ server.registerTool(
   {
     name: "scan-receipt",
     description: "Scan a receipt",
-    view: { component: "scan-receipt", hosts: ["apps-sdk"] },
+    view: { component: "scan-receipt", appsSdk: true },
   },
   handler,
 );
--- a/docs/api-reference/use-files.mdx
+++ b/docs/api-reference/use-files.mdx
@@ -84,6 +84,23 @@ function ReceiptAttach() {
 
 Both paths resolve a download URL and build a [`FileRef`](/api-reference/file-ref) by hand: the hook returns camelCase fields, while `FileRef` wants snake_case with a required `download_url`.
 
+## Host support
+
+The file APIs run on ChatGPT's Apps SDK runtime. Since Skybridge 1.2.0 views render through MCP Apps by default, where ChatGPT does not expose them — a view that needs files must opt back in per tool:
+
+```ts
+server.registerTool(
+  {
+    name: "scan-receipt",
+    description: "Scan a receipt",
+    view: { component: "scan-receipt", hosts: ["apps-sdk"] },
+  },
+  handler,
+);
+```
+
+On every other host (and without the opt-in) the methods throw `NotSupportedError`, so feature-detect with `instanceof` when the view runs cross-host.
+
 ## Returns
 
 `useFiles` returns three functions.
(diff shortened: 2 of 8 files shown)


atirna:
linking the report: #1074 (pinned to 1.1.2 there because of this)

the ext-apps shim angle, for whoever reviews: the compat `window.openai` object ChatGPT installs on the MCP Apps path has none of the three file methods, and `@modelcontextprotocol/ext-apps` 1.7.4 has no upload/select requests in the spec, so the adaptor literally has nothing to route to until the spec grows them. that's why this brings back the per-view `hosts` opt-in rather than teaching the bridge the file APIs.

also happy to drop the `hosts` name if it collides with the v2 cleanup plans (`outputTemplate` was kept out on purpose per the comment in `registerViewResources`) - something like `view.appsSdk: true` would work too if that reads better

atirna:
switched the public API to `view.appsSdk: true`. it names the additive Apps SDK registration, while the MCP Apps resource stays canonical, so it no longer promises host filtering.

catlan:
Any update on this? would love start using it :)

atirna:
the `appsSdk` rename is in. this is waiting for review.

harijoe:
Hey, thanks for the detailed report, and to @atirna for #1082.

I looked into this and I don't think the premise holds: ChatGPT does expose the file APIs on the MCP Apps path. Two data points:

  - Our chatgpt-files example (examples/chatgpt-files, added in #916 after #833) registers only the ext-apps resource, no openai/outputTemplate, and uses upload, selectFiles and getDownloadUrl from useFiles(). It runs in ChatGPT on 1.3.0+.
  - OpenAI's guidance in MCP Apps in ChatGPT is to build on MCP Apps and layer window.openai extensions on top. It lists uploadFile, selectFiles and getFileDownloadUrl as those extensions, available in MCP Apps mode, and asks apps to
    feature-detect them.

  So Skybridge never provided those methods, and #833 didn't remove anything on that side. What you observed (no file methods on window.openai, hostType === "mcp-app") looks like a difference in the ChatGPT client. You mention the desktop app.
  Could you check the same view in ChatGPT web, and share the desktop app version? If web has the methods and desktop doesn't, that's a report for OpenAI rather than something Skybridge can fix.

Agreed on 2) though. uploadFile and getFileDownloadUrl should throw NotSupportedError when the method is missing, like selectFiles does. Can be put in another PR if you want it merged quickier or wait for this discussion to settle first.

catlan:
You're right, thanks for pushing back. I re-tested with a throwaway Skybridge 2.0.0 app (ext-apps resource only, no `openai/outputTemplate`) whose view dumps `typeof window.openai.{uploadFile,selectFiles,getFileDownloadUrl}` and calls `getFileDownloadUrl` with the attached file's id. Same server, same URL, every surface I have:

| surface | view rendered | `window.openai` keys | `uploadFile` / `selectFiles` / `getFileDownloadUrl` | `getFileDownloadUrl(attached file)` |
| --- | --- | --- | --- | --- |
| ChatGPT web, Chat (Safari 27, macOS) | yes | 35 | all functions | ok, signed URL |
| ChatGPT web, Work | yes | 35 | all functions | rejects: `Error: Invalid file id` |
| ChatGPT desktop 26.915.31945, Work | yes | 25 | all `undefined` | `TypeError: this.openai.getFileDownloadUrl is not a function` |
| ChatGPT desktop, Codex | yes (mounted with a delay) | 25 | all `undefined` | same `TypeError` |
| ChatGPT desktop, Chat | no view (host never reads the resource) | – | – | – |

So #1074's premise was wrong for ChatGPT web; the gap is the desktop client. The app I reported from is the current macOS ChatGPT app, which is the Codex-based build (bundle id `com.openai.codex`, MCP `clientInfo` `codex-mcp-client` 0.155.0-alpha.9.2; the view's `ui/initialize` host is `chatgpt 26.915.31945`, `userAgent.device.platform: "native"`). Its `window.openai` has the core bridge only, none of the documented extensions (`uploadFile`, `selectFiles`, `getFileDownloadUrl`, `requestModal`, `requestCheckout`, `setOpenInAppUrl`, `requestClose`), while its handshake advertises `capabilities.experimental["openai/files"]`, so file handling there may be headed for the ext-apps bridge instead. The web Work surface has the methods but hands the tool a `file_…` id that `getFileDownloadUrl` rejects; only Chat's `sediment://file_…` ids resolve. Both go to OpenAI.

atirna:
catlan's retest settles it: web has all three methods, the codex-based desktop build has none, so the opt-in half was papering over an OpenAI gap. dropped it and rescoped this PR to just the guards: uploadFile and getFileDownloadUrl now throw NotSupportedError when window.openai lacks the method, and selectFiles throws the same error instead of a plain Error, so all three file methods behave identically. the apps-sdk types mark them optional and use-files has a host-support section with the instanceof pattern

PR #1083, opened 2026-09-07
fix(devtools): reset display mode on requestClose so modals can be dismissed

### LINKED ISSUE #1055: DevTools: no way out of modal mode, requestClose is a no-op
## Summary

In DevTools, a view that enters modal mode can never leave it. `requestClose()` is a no-op in the host mock, and nothing ever resets `openai.view`, so `useRequestModal().isOpen` stays `true` for the rest of the session.

## Repro

1. In a view, call `open()` from `useRequestModal()`.
2. Render a close button wired to `useRequestClose()` (or to `setDisplayMode("inline")` from `useDisplayMode()`).
3. Click it. The log drawer shows `requestClose`, but the view stays in modal mode and `isOpen` remains `true`.

## Cause

`packages/devtools/src/components/layout/tool-panel/view/create-openai-mock.ts`:

- `requestModal` sets `openai.view = { mode: "modal", params }` (line 76).
- `requestClose` only logs, it never touches `openai.view` or `displayMode` (line 49).
- `requestDisplayMode` updates `openai.displayMode` but leaves `openai.view` alone.

`openai.view` is the source for the `display` host context in core (`packages/core/src/web/bridges/apps-sdk/adaptor.ts`, `getHostContextStore("display")`), and `useRequestModal` derives `isOpen` from `display.mode === "modal"`. Since `openai.view` is written in exactly one place and never cleared, there is no path back to a non-modal mode.

## Expected

`requestClose` should reset `openai.view` (and `displayMode`) to the pre-modal mode, so `isOpen` flips back to `false`. Real hosts dismiss the modal on `requestClose`, so DevTools currently diverges from them, and modal flows cannot be tested locally at all.

Worth considering as well: a manual escape hatch in the DevTools UI (a close affordance on the modal frame, or letting the display-mode selector move out of modal), for apps that do not call `requestClose` themselves.

## Versions

`skybridge@1.4.0`, `@skybridge/devtools@1.4.0`. Also present on `main` (dfb012a).

MatLBS:
## Summary
- `requestClose` in the DevTools `openai` mock only logged the call and never reset any state, so a view that entered modal mode via `useRequestModal().open()` could never leave it — `isOpen` stayed `true` for the rest of the session.
- Fix resets the persisted `displayMode` preference from `"modal"` back to `"inline"` when `requestClose` is called while a modal is open. This reuses the existing `useInspectorPreferencesStore.subscribe(...)` callback already in this file, which syncs `openai.view` away from modal whenever `displayMode` changes — the same mechanism `useDisplayMode()`'s `setDisplayMode("inline")` path relies on, so there's no duplicated reset logic.

Fixes #1055


packages/devtools/vitest.config.ts
```diff
@@ -1,6 +1,12 @@
+import path from "node:path";
 import { defineConfig } from "vitest/config";
 
 export default defineConfig({
+  resolve: {
```
harijoe: This change doesn't belong to the PR

harijoe (changes requested):
Thanks @MatLBS

Just a bit of cleanup to keep this PR focused on the problem it's aiming to solve.

packages/devtools/vitest.config.ts
MatLBS: The regression test I added (`create-openai-mock.test.ts`) imports the preferences store via the `@/` alias (`@/lib/inspector-preferences-store.js`), matching the convention already used across the app (`tsconfig.json` and `vite.config.ts` both define `@` → `./src`). `vitest.config.ts` didn't have that alias yet, so without this change the test fails to resolve the import at runtime.

Happy to switch the test to a relative import instead and drop this config change entirely if you'd rather keep the PR strictly to the `requestClose` fix — let me know which you prefer.

harijoe: Yes, please use a relative link. Moving to absolute imports could be a good idea, but that should be done in a dedicated PR refactoring all the imports, not just this one.

MatLBS: Quick heads up: the module under test (`create-openai-mock.ts`) itself already uses the `@/` alias internally (`@/lib/context-warnings.js`, `@/lib/inspector-preferences-store.js`) — pre-existing, not something I added. So switching just the test's import to a relative path isn't enough; without the alias in `vitest.config.ts`, vitest still fails to resolve those imports when it loads the module under test.

Given that, do you want me to:
1. Drop the regression test entirely for this PR, or
2. Keep the `vitest.config.ts` alias (scoped only to making the existing `@/` imports resolvable in tests, not a refactor)?

I'll hold off on the import refactor either way, per your note.

MatLBS:
@harijoe I just removed the test and reset vitest.config 👌

harijoe (approved):
LGTM, thanks @MatLBS

PR #1085, opened 2026-09-09
feat(evals): add tools-in-order and no-tools matchers with a dedicated evals example

harijoe:
Closes SKY-553.

## Matchers

Two additions to the `expect.chat` namespace:

- `toHaveCalledToolsInOrder(...names)` asserts the named tools succeeded in that relative order. It matches a subsequence, so unrelated calls in between do not break it, and a repeated name requires that many distinct calls. Failure names the tool that never came and the one it was supposed to follow.
- `toHaveCalledNoTools()` asserts nothing was attempted, failed calls included in the count.

Both go through the existing report helper, so failure output keeps the same shape as the other five matchers.

Ticket candidates deliberately skipped: `first-tool-was` is a one-element ordering assertion that plain vitest already expresses readably, and `response-contains` is already covered by `toHaveSaid`. The ticket's note about `JestExtendError` needed no action, since nothing in the repo classifies assertion errors today.

## Prior art

The subsequence semantics match where the field has converged. Promptfoo and Mastra both ship an ordered-subsequence tool matcher and both made it the default over strict sequence matching. LangChain's trajectory evaluator has no subsequence mode at all, its subset and superset modes are unordered containment.

One known gap this leaves: a subsequence matcher permits unlimited extra calls by construction. Mastra, promptfoo and DeepEval all pair ordering with a call-count bound to close that. Worth a follow-up ticket rather than expanding this one.

## The evals example

New code-only example at `examples/evals`, no docs page, no showcase row, no deploy target. A four-tool ski shop with no views, since evals never render one. The tools are shaped so each matcher gets a natural scenario: search and details for a two-step flow, a checkout that needs an identity, and a cart-clearing tool the model should leave alone.

Eight scenarios cover the whole `@skybridge/test` surface rather than just the matchers: every matcher, multi-turn conversations, the authenticated and anonymous paths, `.not`, direct reads of `chat.toolCalls` and `chat.assistantTurns`, the `systemPrompt` and `maxSteps` overrides, and the plugin-level shared defaults.

All eight pass against a live model, verified on several runs.

## Removals

The eval suites are gone from `examples/capitals` and `examples/auth-descope`, along with their tsconfigs, scripts, dev dependencies, README sections and the `evals` plugin option.

One thing this costs: auth-descope was the only place `authInfo` ran against a real identity provider. The new example fakes it by having the checkout tool inspect the auth info directly, which exercises the matcher but no longer covers the Descope path. Evals never ran in CI, so nothing regresses there.

## Notes for review

- `temperature` is not exercised as a per-scenario override. The plugin default and the library default are both zero, so an override would have been indistinguishable from no override, and a non-zero value would only add flakiness.
- Matchers see the whole conversation, not just the last `send`. That was already true of the existing five; the docs now say so.


skills/chatgpt-app-builder/references/evals.md
```diff
@@ -44,8 +44,10 @@ All typed against the app's registry (`name` autocompletes, `args` is checked ag
 | `toNeverHaveCalledTool(name)` | no call was attempted |
 | `toHaveFailedToolCall(name)` | a call was refused (auth) or threw |
 | `toHaveSaid(text \| RegExp)` | an assistant turn contains it (string match is case- and whitespace-insensitive) |
+| `toHaveCalledToolsInOrder(...names)` | the named tools *succeeded* in that relative order (subsequence, gaps allowed) |
```
qchuchu, line 47: how do we do for the failing symlink path in the case of the skybridge skill ? I guess maybe it's not important

qchuchu (approved):
I just added a comment regarding the symlink - I would have prefered to have the removals + the example in separate commits, easier to read !

All good otherwise !

PR #1086, opened 2026-09-09
docs: add a monetization guide

harijoe:
## Summary

- Adds `docs/guides/monetization.mdx`, a guide to the three ways money can reach a Skybridge app: a shopper paying on your own checkout, the agent paying for a tool call, or an advertiser paying for a slot.
- Registers the page in `docs.json` and adds its card to the guides index.
- Adds the four cards the guides index was missing (`ecommerce`, `monetization`, `auth-providers`, `skills`) and orders them to match the sidebar.

Closes SKY-560.

_Screenshot of the full page: to be attached._

## Architecture Notes

The MPP section charges from `mcpMiddleware`, not from inside the tool handler. This is not a style choice: a challenge thrown inside a handler is flattened into an `isError` tool result, so the machine-readable payload never reaches the client and the agent gets a message with nothing to pay against. From middleware the challenge arrives as a JSON-RPC error with the signed challenge intact. Verified against a running server rather than inferred.

The upstream Stripe guide documents a different shape (a tool returns a payment link, a separate HTTP endpoint runs the exchange). That needs a custom Express route, and Alpic Cloud routes only `/mcp`, so it is mentioned as an alternative for self-hosted deployments instead of being the recommended path.

## Verification

Both code samples were run, not just written.

- The MPP flow was exercised against a live Skybridge server: an unpaid `tools/call` returns `-32042` with a signed challenge carrying amount, currency, method and expiry.
- Three constraints came out of that run and are documented: `MPP_SECRET_KEY` must be at least 32 bytes, `realm` should be set to avoid a boot warning, and `@modelcontextprotocol/sdk` (v1) must be installed alongside because `mppx` imports it to build the error.
- `pnpm test` (458 tests) and `pnpm build` pass. `mintlify broken-links` passes.

Not verified: paying a challenge and redeeming the receipt, which needs Shared Payment Token access. The section carries a warning saying so.

## Alternatives Considered

A "Sell with Stripe Checkout" section was written and then cut. Closing the payment loop meant a server-side long poll bounded under the MCP client request timeout (60s by default), which a shopper entering card details will often exceed. A `get-payment-status` tool polled from the view is the better shape, and the section will come back built that way.

A dedicated access-gating section was also dropped. The mechanism is entirely `/build/auth` and `/guides/auth-providers`, with nothing monetization-specific to add.

## Screenshot

image

### DIFF
--- a/docs/guides/monetization.mdx
+++ b/docs/guides/monetization.mdx
@@ -1,0 +1,195 @@
+---
+title: "Monetize Your MCP App"
+description: "Discover the different ways to make money with your MCP App"
+sidebarTitle: "Monetization"
+icon: "circle-dollar-sign"
+---
+
+The two main ways to monetize your MCP App are selling:
+- goods or services to your app users
+- sponsored slots to advertisers
+
+A third payer is possible too, and newer: the agent itself, paying out of a wallet its user funded.
+
+## Hand Off to Your Own Checkout
+
+The simplest option, and what the [`ecom` template](/guides/ecommerce) ships: send the shopper to a page you already run. The navigation has to go through the host with [`useOpenExternal`](/api-reference/use-open-external).
+
+```tsx views/carousel/detail/index.tsx
+import { useOpenExternal } from "skybridge/web";
+
+function BuyButton({ variant }) {
+  const openExternal = useOpenExternal();
+
+  return (
+    <button
+      type="button"
+      onClick={() => openExternal(variant.url)}
+    >
+      View on site
+    </button>
+  );
+}
+```
+
+List your checkout origin in [`redirectDomains`](/guides/csp#redirect-off-app) so the host skips its safe-link confirmation on the way out.
+
+```ts server.ts
+view: {
+  component: "carousel",
+  csp: { redirectDomains: ["https://checkout.myshop.com"] },
+},
+```
+
+What you give up is the return leg. The host opens the page outside the iframe and nothing comes back, so the view keeps rendering the cart and the model still believes the order is pending. This is usually fine for catalogs focused on sending qualified traffic to a storefront. If you want to update the cart after redirect, one strategy is to have the view poll the cart status until it's updated elsewhere.
+
+## Charge the Agent
+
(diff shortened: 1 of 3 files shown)


docs/guides/monetization.mdx
```diff
@@ -0,0 +1,226 @@
...
+});
+```
+
+The middleware runs before the handler, so an unpaid call never reaches your code. Pricing lives in one table keyed by tool name, and a tool absent from it stays free. `MPP_SECRET_KEY` signs the challenges, so it has to be at least 32 bytes (`openssl rand -base64 32`), and setting `realm` avoids a fallback warning at boot.
+
+<Warning>
+This is experimental on every side. It needs [Shared Payment Token](https://docs.stripe.com/agentic-commerce/concepts/shared-payment-tokens) access and a Stripe profile, the protocol is young, and no chat host pays for a tool call on a user's behalf today. Agent harnesses with their own wallet, such as a coding agent, are where it currently applies. We have verified the challenge leg against a live Skybridge server; paying one and redeeming the receipt is untested.
+</Warning>
+
+Note that `mppx` imports `@modelcontextprotocol/sdk` (the v1 SDK) to build the error, so it has to be installed alongside Skybridge's own v2 packages. Without it the charge fails at runtime with a missing-dependency message.
```
fredericbarthelet: Is that necessary? Or doest the package comes bundled in mppx dependencies?

docs/guides/monetization.mdx
```diff
@@ -0,0 +1,226 @@
...
+sidebarTitle: "Monetization"
+icon: "circle-dollar-sign"
+---
+
+The two main ways to monetize your MCP App are selling:
+- goods or services to your app users
+- sponsored slots to advertisers
+
+A third payer is possible too, and newer: the agent itself, paying out of a wallet its user funded.
+
```
fredericbarthelet, line 15: You list 1/2 and present a third "aside" option, and then detail 1/3/2

Maybe consider all options as viable monetization options and detail them in the same order :

- sell goods / services to human
- sell goods / services to agent
- advertise

WDYT?

fredericbarthelet (approved):

docs/guides/monetization.mdx
harijoe: It's a peer dep of mppx. I made the wording a bit clearer

PR #1090, opened 2026-09-14
feat(test): grade a conversation with an LLM judge

harijoe:
## Summary

- `expect.chat(chat).toPassJudgment(criteria, options?)`: an async matcher that hands the conversation to a second model and grades it against criteria written in plain English. `.not` works.
- The judge runs `generateObject` at temperature 0 with a fixed system prompt and a `{ pass, reasoning }` schema, so a verdict never depends on parsing prose. Its reasoning lands in the vitest failure message; the transcript is not printed.
- Provider errors throw `judge unavailable: <error>`, worded apart from a `FAIL` verdict so a flaky provider is never read as an app regression. No retry on top of the AI SDK's own.
- `options`: `model`, another judge, defaulting to the chat's own. The judge always reads the whole conversation.
- `Chat` now records an ordered `transcript` (user turns, assistant turns, tool calls with arguments and failures), which is what the judge reads.
- Docs and the app-builder skill reference updated, per the public-API rule.

Implements [SKY-549](https://linear.app/alpic/issue/SKY-549/evals-judge-helper), from the Evals Judge PRD.

## Architecture Notes

The PRD put a default judge model in the plugin config (`evals.judge.model`). That option is not here: the config reaches the worker through vitest's `provide`/`inject`, which structured-clones its values, and a `LanguageModel` instance does not survive that. The resolution order is the assertion's `options.model`, then the model the chat already runs on, which every scenario sets on `start`.

The matcher lives with the others in `@skybridge/vite-plugin`, which has no dependency on `ai`. It reaches `generateObject` through a dynamic `import("ai")` inside the judge call, and `ai` is declared an optional peer, the same treatment `vitest` already gets in that package.


harijoe:
@greptileai

packages/vite-plugin/src/evals/matchers.ts
```diff
@@ -1,8 +1,15 @@
...
+import type { LanguageModel } from "ai";
 import type { ToolInput, ToolNames } from "skybridge/server";
 import { expect } from "vitest";
 import type { ChatLike, ToolCall } from "./types.js";
 
+/** Options for {@link ChatMatchers.toPassJudgment}. */
+export interface JudgmentOptions {
+  /** Overrides the judge for this assertion. Defaults to the chat's model. */
+  model?: LanguageModel;
+}
```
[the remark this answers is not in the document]
harijoe: That was on purpose

packages/vite-plugin/src/evals/matchers.ts
```diff
@@ -87,6 +95,49 @@ function spoken(chat: ChatLike<unknown>): string {
...
+function transcriptFor(chat: ChatLike<unknown>): string {
+  if (chat.transcript.length === 0) {
+    return "(the conversation is empty)";
+  }
+  return chat.transcript
+    .map((entry) => `${entry.role}: ${fence(entry.text)}`)
+    .join("\n");
+}
+
+async function askJudge(
```
fredericbarthelet: why did you split it in a dedicated function? is it planned to be used by another matcher?

packages/vite-plugin/package.json
```diff
@@ -35,22 +35,27 @@
   "peerDependencies": {
     "skybridge": "^2.0.0",
     "vite": ">=7.3.1",
-    "vitest": ">=4.1.0"
+    "vitest": ">=4.1.0",
+    "ai": "^7.0.0"
```
fredericbarthelet: why a peerdep to ai now?

packages/test/src/chat.ts
```diff
@@ -101,18 +126,25 @@ export class Chat<App = unknown> {
     });
 
     for (const step of result.steps) {
+      if (step.text !== "") {
+        this.assistantTurns.push(step.text);
+        this.transcript.push({ role: "assistant", text: step.text });
```
fredericbarthelet: it feels redundant to store in both properties. Can't consumers of assistantTurns now use transcript for their needs too?

fredericbarthelet (approved):
A few open questions and optimization comments, but nothing blocking. LGTM :)

